PR #1, opened 2017-03-13
iptables-based portmapping plugin

squeed:
This is a fairly standard iptables-based portmapping plugin. It supports ipv4 and ipv6.

This replaces the proposed plugin in the [cni repository](https://github.com/containernetworking/cni/pull/378).

### DIFF
--- a/plugins/meta/portmap/README.md
+++ b/plugins/meta/portmap/README.md
@@ -1,0 +1,117 @@
+## Port-mapping plugin
+
+This plugin will forward traffic from one or more ports on the host to the
+container. It expects to be run as a chained plugin.
+
+## Usage
+You should use this plugin as part of a network configuration list. It accepts
+the following configuration options:
+
+* `snat` - boolean, default true. If true or omitted, set up the SNAT chains
+* `conditionsV4`, `conditionsV6` - array of strings. A list of arbitrary `iptables` 
+matches to add to the per-container rule. This may be useful if you wish to 
+exclude specific IPs from port-mapping
+
+The plugin expects to receive the actual list of port mappings via the 
+`portMappings` [capability argument](https://github.com/containernetworking/cni/blob/master/CONVENTIONS.md)
+
+So a sample standalone config list (with the file extension .conflist) might
+look like:
+
+```json
+{
+        "cniVersion": "0.3.1",
+        "name": "mynet",
+        "plugins": [
+                {
+                        "type": "ptp",
+                        "ipMasq": true,
+                        "ipam": {
+                                "type": "host-local",
+                                "subnet": "172.16.30.0/24",
+                                "routes": [
+                                        {
+                                                "dst": "0.0.0.0/0"
+                                        }
+                                ]
+                        }
+                },
+                {
+                        "type": "portmap",
+                        "capabilities": {"portMappings": true},
+                        "snat": false,
+                        "conditionsV4": ["!", "-d", "192.0.2.0/24"],
+                        "conditionsV6": ["!", "-d", "fc00::/7"]
+                }
+        ]
+}
+```
+
+
+
+## Rule structure
+The plugin sets up two sequences of chains and rules - one "primary" DNAT
+sequence to rewrite the destination, and one additional SNAT sequence that
+rewrites the source address for packets from localhost. The sequence is somewhat
+complex to minimize the number of rules non-forwarded packets must traverse.
+
+
+### DNAT
+The DNAT rule rewrites the destination port and address of new connections.
+There is a top-level chain, `CNI-HOSTPORT-DNAT` which is always created and
+never deleted. Each plugin execution creates an additional chain for ease
+of cleanup. So, if a single container exists on IP 172.16.30.2 with ports 
+8080 and 8043 on the host forwarded to ports 80 and 443 in the container, the 
+rules look like this:
+
+`PREROUTING`, `OUTPUT` chains:
+- `--dst-type LOCAL -j CNI-HOSTPORT-DNAT`
+
+`CNI-HOSTPORT-DNAT` chain:
+- `${ConditionsV4/6} -j CNI-DN-xxxxxx` (where xxxxxx is a function of the ContainerID and network name)
+
+`CNI-DN-xxxxxx` chain: 
+- `-p tcp --dport 8080 -j DNAT --to-destination 172.16.30.2:80`
+- `-p tcp --dport 8043 -j DNAT --to-destination 172.16.30.2:443`
+
+New connections to the host will have to traverse every rule, so large numbers
+of port forwards may have a performance impact. This won't affect established
+connections, just the first packet.
+
+### SNAT 
+The SNAT rule enables port-forwarding from the localhost IP on the host.
+This rule rewrites (masquerades) the source address for connections from
+localhost. If this rule did not exist, a connection to `localhost:80` would
+still have a source IP of 127.0.0.1 when received by the container, so no 
+packets would respond. Again, it is a sequence of 3 chains. Because SNAT has to
+occur in the `POSTROUTING` chain, the packet has already been through the DNAT
+chain.
+
+`POSTROUTING`:
+- `-s 127.0.0.1 ! -d 127.0.0.1 -j CNI-HOSTPORT-SNAT`
+
+`CNI-HOSTPORT-SNAT`:
+- `-j CNI-SN-xxxxx`
+
+`CNI-SN-xxxxx`:
+- `-p tcp -s 127.0.0.1 -d 172.16.30.2 --dport 80 -j MASQUERADE`
+- `-p tcp -s 127.0.0.1 -d 172.16.30.2 --dport 443 -j MASQUERADE`
+
+Only new connections from the host, where the source address is 127.0.0.1 but
+not the destination will traverse this chain. It is unlikely that any packets
+will reach these rules without being SNATted, so the cost should be minimal.
+
+Because MASQUERADE happens in POSTROUTING, it means that packets with source ip
+127.0.0.1 need to pass a routing boundary. By default, that is not allowed
+in Linux. So, need to enable the sysctl `net.ipv4.conf.IFNAME.route_localnet`,
+where IFNAME is the name of the host-side interface that routes traffic to the
+container.
+
+There is no equivalent to `route_localnet` for ipv6, so SNAT does not work
+for ipv6. If you need port forwarding from localhost, your container must have
+an ipv4 address.
+
+
+## Known issues
+- ipsets could improve efficiency
+- SNAT does not work with ipv6.
--- a/plugins/meta/portmap/chain.go
+++ b/plugins/meta/portmap/chain.go
@@ -1,0 +1,127 @@
+// Copyright 2017 CNI authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/coreos/go-iptables/iptables"
+	shellwords "github.com/mattn/go-shellwords"
+)
+
+type chain struct {
+	table       string
+	name        string
+	entryRule   []string // the rule that enters this chain
+	entryChains []string // the chains to add the entry rule
+}
+
+// setup idempotently creates the chain. It will not error if the chain exists.
+func (c *chain) setup(ipt *iptables.IPTables, rules [][]string) error {
+	// create the chain
+	exists, err := chainExists(ipt, c.table, c.name)
+	if err != nil {
+		return err
+	}
+	if !exists {
+		if err := ipt.NewChain(c.table, c.name); err != nil {
+			return err
+		}
+	}
+
+	// Add the rules to the chain
+	for i := len(rules) - 1; i >= 0; i-- {
+		if err := prependUnique(ipt, c.table, c.name, rules[i]); err != nil {
+			return err
+		}
+	}
+
+	// Add the entry rules
+	entryRule := append(c.entryRule, "-j", c.name)
+	for _, entryChain := range c.entryChains {
+		if err := prependUnique(ipt, c.table, entryChain, entryRule); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+// teardown idempotently deletes a chain. It will not error if the chain doesn't exist.
+// It will first delete all references to this chain in the entryChains.
+func (c *chain) teardown(ipt *iptables.IPTables) error {
+	// flush the chain
+	// This will succeed *and create the chain* if it does not exist.
+	// If the chain doesn't exist, the next checks will fail.
+	if err := ipt.ClearChain(c.table, c.name); err != nil {
+		return err
+	}
+
+	for _, entryChain := range c.entryChains {
+		entryChainRules, err := ipt.List(c.table, entryChain)
+		if err != nil {
+			// Swallow error here - probably the chain doesn't exist.
+			// If we miss something the deletion will fail
+			continue
+		}
+
+		for _, entryChainRule := range entryChainRules[1:] {
+			if strings.HasSuffix(entryChainRule, "-j "+c.name) {
+				chainParts, err := shellwords.Parse(entryChainRule)
+				if err != nil {
+					return fmt.Errorf("error parsing iptables rule: %s: %v", entryChainRule, err)
+				}
+				chainParts = chainParts[2:] // List results always include an -A CHAINNAME
+
+				if err := ipt.Delete(c.table, entryChain, chainParts...); err != nil {
+					return fmt.Errorf("Failed to delete referring rule %s %s: %v", c.table, entryChainRule, err)
+				}
+			}
+		}
+	}
+
+	if err := ipt.DeleteChain(c.table, c.name); err != nil {
+		return err
+	}
+	return nil
+}
+
+// prependUnique will prepend a rule to a chain, if it does not already exist
+func prependUnique(ipt *iptables.IPTables, table, chain string, rule []string) error {
+	exists, err := ipt.Exists(table, chain, rule...)
+	if err != nil {
+		return err
+	}
+	if exists {
+		return nil
+	}
+
+	return ipt.Insert(table, chain, 1, rule...)
+}
+
+func chainExists(ipt *iptables.IPTables, tableName, chainName string) (bool, error) {
+	chains, err := ipt.ListChains(tableName)
+	if err != nil {
+		return false, err
+	}
+
+	for _, ch := range chains {
+		if ch == chainName {
+			return true, nil
+		}
+	}
+	return false, nil
+}
--- a/plugins/meta/portmap/main.go
+++ b/plugins/meta/portmap/main.go
@@ -1,0 +1,159 @@
+// Copyright 2017 CNI authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// This is a post-setup plugin that establishes port forwarding - using iptables,
+// from the host's network interface(s) to a pod's network interface.
+//
+// It is intended to be used as a chained CNI plugin, and determines the container
+// IP from the previous result. If the result includes an IPv6 address, it will
+// also be configured. (IPTables will not forward cross-family).
+//
+// This has one notable limitation: it does not perform any kind of reservation
+// of the actual host port. If there is a service on the host, it will have all
+// its traffic captured by the container. If another container also claims a given
+// port, it will caputure the traffic - it is last-write-wins.
+package main
+
+import (
+	"encoding/json"
+	"fmt"
+
+	"github.com/containernetworking/cni/pkg/skel"
+	"github.com/containernetworking/cni/pkg/types"
+	"github.com/containernetworking/cni/pkg/types/current"
+	"github.com/containernetworking/cni/pkg/version"
+)
+
+// PortMapEntry corresponds to a single entry in the port_mappings argument,
+// see CONVENTIONS.md
+type PortMapEntry struct {
+	HostPort      int    `json:"hostPort"`
+	ContainerPort int    `json:"containerPort"`
+	Protocol      string `json:"protocol"`
+	HostIP        string `json:"hostIP,omitempty"`
+}
+
+type PortMapConf struct {
+	types.NetConf
+	SNAT          *bool     `json:"snat,omitempty"`
+	ConditionsV4  *[]string `json:"conditionsV4"`
+	ConditionsV6  *[]string `json:"conditionsV6"`
+	RuntimeConfig struct {
+		PortMaps []PortMapEntry `json:"portMappings,omitempty"`
+	} `json:"runtimeConfig,omitempty"`
+	RawPrevResult map[string]interface{} `json:"prevResult,omitempty"`
+	PrevResult    *current.Result        `json:"-"`
+	ContainerID   string
+}
+
+func cmdAdd(args *skel.CmdArgs) error {
+	netConf, err := parseConfig(args.StdinData)
+	if err != nil {
+		return fmt.Errorf("failed to parse config: %v", err)
+	}
+
+	if netConf.PrevResult == nil {
+		return fmt.Errorf("must be called as chained plugin")
+	}
+
+	if len(netConf.RuntimeConfig.PortMaps) == 0 {
+		return types.PrintResult(netConf.PrevResult, netConf.CNIVersion)
+	}
+
+	netConf.ContainerID = args.ContainerID
+
+	// Loop through IPs, setting up forwarding to the first container IP
+	// per family
+	hasV4 := false
+	hasV6 := false
+	for _, ip := range netConf.PrevResult.IPs {
+		if ip.Version == "6" && hasV6 {
+			continue
+		} else if ip.Version == "4" && hasV4 {
+			continue
+		}
+
+		// Skip known non-sandbox interfaces
+		intIdx := ip.Interface
+		if intIdx >= 0 && intIdx < len(netConf.PrevResult.Interfaces) && netConf.PrevResult.Interfaces[intIdx].Name != args.IfName {
+			continue
+		}
+
+		if err := forwardPorts(netConf, ip.Address.IP); err != nil {
+			return err
+		}
+
+		if ip.Version == "6" {
+			hasV6 = true
+		} else {
+			hasV4 = true
+		}
+	}
+
+	// Pass through the previous result
+	return types.PrintResult(netConf.PrevResult, netConf.CNIVersion)
+}
+
+func cmdDel(args *skel.CmdArgs) error {
+	netConf, err := parseConfig(args.StdinData)
+	if err != nil {
+		return fmt.Errorf("failed to parse config: %v", err)
+	}
+
+	netConf.ContainerID = args.ContainerID
+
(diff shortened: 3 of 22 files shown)


plugins/portmap/README.md
```diff
@@ -0,0 +1,103 @@
...
+
+### DNAT
+The DNAT rule rewrites the destination port and address of new connections.
+There is a top-level chain, `CNI-HOSTPORT-DNAT` which is always created and
+never deleted. Each plugin execution creates an additional chain for ease
+of cleanup. So, if a single container exists on IP 172.16.30.2 with ports 
+8080 and 8043 on the host forwarded to ports 80 and 443 in the container, the 
+rules look like this:
+
+`PREROUTING`, `OUTPUT` chains:
```
thockin: Something we discovered in Kubernetes - this expression is looser than we really want.  This will match any address in the `local` routing table, whereas I think we only want addresses that are actually bound to interfaces.  The delta here is things like load-balancer VIPs on GCE.

The solutions that is pending is to loop over all interfaces and expand them into a set of exact-IP matches.  It is sort of awful because, at least in theory, this can change, so it needs to be refreshed periodically. I don't know how else to express this condition, but I am open to ideas.

An alternative would be to document this as "must be the last rule in the chain" but that feels very brittle in the face of collaboration users of iptables.

plugins/portmap/README.md
```diff
@@ -0,0 +1,103 @@
...
+### SNAT
+The SNAT rule rewrites (masquerades) the source address for connections from
+localhost. If this rule did not exist, a connection to `localhost:80` would
+still have a source IP of 127.0.0.1 when received by the container, so no 
+packets would respond. Again, it is a sequence of 3 chains. Because SNAT has to
+occur in the `POSTROUTING` chain, the packet has already been through the DNAT
+chain.
+
+`POSTROUTING`:
+- `-s 127.0.0.1 ! -d 127.0.0.1 -j CNI-HOSTPORT-SNAT`
```
thockin: should be 127.0.0.0/8  for both?

plugins/portmap/README.md
```diff
@@ -0,0 +1,103 @@
...
+
+`CNI-HOSTPORT-SNAT`:
+- `-j CNI-SN-xxxxx`
+
+`CNI-SN-xxxxx`:
+- `-p tcp -s 127.0.0.1 -d 172.16.30.2 --dport 80 -j MASQUERADE`
```
thockin: Do we need the -s at all?  It was a pre-condition to end up here in the first place.  Do we even need the destination and the port?  Is there any case where the initial condition (-s localhost -d not-localhost) doesn't need masquerade?

squeed:
> We open and bind() the actual port on the host. 

I thought about doing this, with a helper daemon à la DHCP, but decided against the trouble. If you think it's a worthy idea, it could still be done. It can also be added later.

thockin:
I think it is closing a real gap where users can get confused, but it is a
relatively slim gap.  Obviously, if it is easy, we should do it.  How easy
is easy enough?

Also, yes, it can be additive

plugins/portmap/README.md
squeed: Interesting. I'd always thought the `local` routing table was essentially read-only, but a quick experiment shows it's not (neat!).

Question: are you just concerned about performance, or is it expected that traffic from these interfaces are not forwarded? In a (non-k8s?) case, is it desirable that a GCE vip *would* be forwarded to a container? A quick glance at the documentation makes this at least seem plausible.

A few thoughts on possible solutions:
- Have a daemon that listens to netlink messages and updates a linux `ipset` with the list of local ifs
- Add a configuration option to the plugin to add conditions to the second-level entry rule. Then the administrator could exclude either specific CIDRs, devgroups, or ipsets
- Write the port reservation daemon, then match on `socket` (this is medium sketchy)

Either way, I'm thinking about adding a portset (via `ipset(8)`) to the top-level entry rule for improved performance.

plugins/portmap/README.md
squeed: We definitely don't need the `-s`. I added the `-d` and `--dport` so that the minimal possible set of traffic from 127.0.0.1 is allowed out.  Otherwise this is a nasty potential firewall bypass owing to the fact that `route_localnet` is enabled.

thockin:
On Wed, Mar 15, 2017 at 10:48 AM, Casey Callendrello
<notifications@github.com> wrote:
>
> @squeed commented on this pull request.
>
> ________________________________
>
> In plugins/portmap/README.md:
>
> > +## Rule structure
> +The plugin sets up two sequences of chains and rules - one "primary" DNAT
> +sequence to rewrite the destination, and one additional SNAT sequence that
> +rewrites the source address for packets from localhost. The sequence is somewhat
> +complex to minimize the number of rules non-forwarded packets must traverse.
> +
> +
> +### DNAT
> +The DNAT rule rewrites the destination port and address of new connections.
> +There is a top-level chain, `CNI-HOSTPORT-DNAT` which is always created and
> +never deleted. Each plugin execution creates an additional chain for ease
> +of cleanup. So, if a single container exists on IP 172.16.30.2 with ports
> +8080 and 8043 on the host forwarded to ports 80 and 443 in the container, the
> +rules look like this:
> +
> +`PREROUTING`, `OUTPUT` chains:
>
> Interesting. I'd always thought the local routing table was essentially read-only, but a quick experiment shows it's not (neat!).

Yeah, I learned this through GCE's tricks.

> Question: are you just concerned about performance, or is it expected that traffic from these interfaces are not forwarded? In a (non-k8s?) case, is it desirable that a GCE vip would be forwarded to a container? A quick glance at the documentation makes this at least seem plausible.

GCE by default assigns VIPs as local addresses.  Kubernetes on GCE
instead traps those VIPs and redirects to containers.  Overall this
works really well, but now we have 2 different entities - CNI plugin
and kube-proxy - writing iptables rules that potentially overlap.  We
can either try to ensure ordering of rules from most-specific to least
specific, or we can refine our definition of what we trap so that
ordering is not a factor.  I obviously prefer the latter, but it's not
perfect, either.

> A few thoughts on possible solutions:
>
> Have a daemon that listens to netlink messages and updates a linux ipset with the list of local ifs
> Add a configuration option to the plugin to add conditions to the second-level entry rule. Then the administrator could exclude either specific CIDRs, devgroups, or ipsets
> Write the port reservation daemon, then match on socket (this is medium sketchy)
>
> Either way, I'm thinking about adding a portset (via ipset(8)) to the top-level entry rule for improved performance.

I would not focus on performance yet.  Let's figure out what we think
is the correct semantic.

thockin:
I don't understand the firewall bypass?  If this was (implicitly, before
jumping to this rule) `-s 127.0.0.0/8` and explicitly `! -d 127.0.0.0/8 -J
MASQUERADE`, what am I missing?  One rule to catch it all..

squeed:
I'm not 100% certain about the firewall bypass, but my though is that, since we've enabled route_localnet, if *all* traffic from 127.0.0.1 to the 'net gets MASQ'd, this happens after the `OUTPUT` chain on the filter table. Any policies that depend on source ip or if could be skipped.
Yeah, it's a bit contrived, but not too weird, as far as these things go.

In any case, I'll be on vacation, offline for the coming week. This PR is editable by the general team - all are free to make changes.

squeed:
Yeah, I was partially right about the firewall bypass - it's slightly contrived, but IMHO worth preventing. If you have a FILTER rule that specifies the source IP, then it is bypassed.

e.g.
```
iptables -t filter -A OUTPUT -p tcp -s 192.168.121.184 --dport 5050 -j DROP
```

Then if you have
```
iptables -t nat -A POSTROUTING -s 127.0.0.1 ! -d 127.0.0.1 -j MASQUERADE
```

This command works, though it shouldn't:

```
nc -s 127.0.0.1 10.7.18.117 5050
```

Admittedly it's a bit silly to have source IPs in the `filter.OUTPUT` chain, but it's plausible. Given that we also intend to use this plugin in rkt, which is a bit more general-purpose than k8s, I'm going to keep the more specific MASQUERADE entry conditions.

plugins/portmap/README.md
```diff
@@ -0,0 +1,103 @@
...
+                                "routes": [
+                                        {
+                                                "dst": "0.0.0.0/0"
+                                        }
+                                ]
+                        }
+                },
+                {
+                        "type": "portmap",
+						"capabilities": {"portMappings": true}
```
dcbw: Indent?

feiskyer:
Any updates on this?

squeed:
Sorry, this fell off of my plate. I should be done with this in the next few days.

caseydavenport:
> We open and bind() the actual port on the host.

It does sound like a helper daemon might be desired for this, and the dst type LOCAL concerns as well.

This feels like a minor enough window to me that I'd be happy to pass on it for now and add later if it becomes a problem, or once we add a helper daemon for some other reason.

thockin:
And to be fair, I sort of hate the "specific IPs" model, too - still hoping
for a better answer.

squeed:
Ok, I pushed an update to the plugin. It now takes a `conditions` argument (actually, one per family) so that the administrator can arbitrarily add conditions to the forwarding entry rules. While this doesn't directly solve the GKE / `LOCAL` problem, it does mean that fixing it is not a code change.

At this point in time, I feel it's ready for a release. Adding a port-reservation daemon would be a welcome addition, but that can be an additional feature.

thockin:
I'm not sure I see how the conditions fixes the local problem, but I am OK to get this in and iterate..

feiskyer:
+1 for get this in first.

squeed:
The idea is so the admin can add additional conditions to patch up the
local problem. One could exclude or require specific subnets, if those are
predictable and distinct, for example.

thockin:
excluding subnets isn't going to help the problem I see.  besides that, a
single "conditions" block doesn't allow me to exclude 2 different ranges.
To solve the problem I see (or rather, the way I know how - please teach me
if I am wrong :) is to change 1 "addrtype LOCAL" rule into N "-d" rules.

On Fri, May 5, 2017 at 2:57 AM, Casey Callendrello <notifications@github.com

squeed:
I'm sure both of us don't quite have the whole picture :-).

With a single iptables rule you can *exclude* multiple ranges. You just can't *include* multiple ranges directly. In other words, iptables matches are always AND, never OR. However, you can match on multiple ranges using ipsets, which adds a bit more complexity but isn't too bad.

In any case, any administrator-provided conditions would be in addition to the `--addrtype LOCAL`.

Could you provide a more concrete example of how the GCE vips appear on the host? It's a bit tricky to reason about otherwise.  I was assuming the machine had an eth0 of, say, 10.42.1.2 and an additional local routing table entry of, say, 8.8.8.8.

Does it make sense? My goal is to produce something simple and broadly useful, and this is a very good motivating example. I'm happy to try and get it right.

thockin:
How do you exclude multiple ranges?

```
# iptables -t nat -d 1.2.3.0/24 ! -d 1.2.3.0/25 -j LOG
iptables v1.4.21: multiple -d flags not allowed
Try `iptables -h' or 'iptables --help' for more information.
```


When a packet to a VIP arrives at the node, srcip is the original client,
and dstip is the LB VIP.  To make the kernel receive it in normal VM-mode,
the LB IP is added to the `local` route table.  Because we mux containers
to nodes N:1, we demux LB packets in iptables.  We recognize the dstip and
dstport, and DNAT based on that.

The problem is that those packets match addrtype LOCAL *and* the
VIP-specific rule.

So we could demand ordering, and say that hostport checks should always be
at the end.  That seems delicate.  We could just code the list of allowed
IPs for hostports.  That's tedious, but seems semantically closer.  Very
open to better ideas.

squeed:
Argh, you're totally right about multiple `-d`. I don't know what the heck I was thinking.

plugins/portmap/main.go
```diff
@@ -0,0 +1,159 @@
...
+	// We don't need to parse out whether or not we're using v6 or snat,
+	// deletion is idempotent
+	if err := unforwardPorts(netConf); err != nil {
+		return err
+	}
+	return nil
+}
+
+func main() {
+	skel.PluginMain(cmdAdd, cmdDel, version.PluginSupports("", "0.1.0", "0.2.0", version.Current()))
```
gunjan5: Any reason why the supported versions are specified this way? This will skip `0.3.0` since `version.Current()` returns `0.3.1`, and your sample config in the README.md uses `0.3.0` https://github.com/containernetworking/plugins/pull/1/files#diff-f6ee9b32ca4cfb6c2075e1ee55dadd07R22 
I see other CNI plugins just use `version.All` .

plugins/meta/portmap/chain.go
```diff
@@ -0,0 +1,127 @@
...
+	}
+	if !exists {
+		if err := ipt.NewChain(c.table, c.name); err != nil {
+			return err
+		}
+	}
+
+	// Add the rules to the chain
+	for _, rule := range rules {
+		if err := prependUnique(ipt, c.table, c.name, rule); err != nil {
```
bboreham: should you state in the description that rules are added in reverse order?

plugins/meta/portmap/portmap.go
```diff
@@ -0,0 +1,266 @@
...
+		conditions = config.ConditionsV6
+	} else {
+		ipt, err = iptables.NewWithProtocol(iptables.ProtocolIPv4)
+		conditions = config.ConditionsV4
+	}
+	if err != nil {
+		return fmt.Errorf("failed to open iptables: %v", err)
+	}
+
+	toplevelDnatChan := genToplevelDnatChain()
```
bboreham: s/Chan/Chain/ ?

plugins/meta/portmap/portmap.go
```diff
@@ -0,0 +1,266 @@
...
+	if err := toplevelDnatChan.setup(ipt, nil); err != nil {
+		return fmt.Errorf("failed to create top-level DNAT chain: %v", err)
+	}
+
+	dnatChain := genDnatChain(config.Name, config.ContainerID, conditions)
+	_ = dnatChain.teardown(ipt) // If we somehow collide on this container ID + network, cleanup
+
+	dnatRules := dnatRules(config.RuntimeConfig.PortMaps, containerIP)
+	if err := dnatChain.setup(ipt, dnatRules); err != nil {
```
bboreham: Instead of calling `teardown` then have `setup` call `NewChain` it could just call `ClearChain` which will either create it or flush it.  But that introduces another variant, so maybe not better...

plugins/meta/portmap/portmap.go
```diff
@@ -0,0 +1,266 @@
...
+		toplevelSnatChain := genToplevelSnatChain(isV6)
+		if err := toplevelSnatChain.setup(ipt, nil); err != nil {
+			return fmt.Errorf("failed to create top-level SNAT chain: %v", err)
+		}
+
+		snatChain := genSnatChain(config.Name, config.ContainerID)
+		_ = snatChain.teardown(ipt)
+
+		snatRules := snatRules(config.RuntimeConfig.PortMaps, containerIP)
+		if err := snatChain.setup(ipt, snatRules); err != nil {
```
bboreham: Same comment as above re `teardown` ... `setup`

plugins/meta/portmap/chain.go
```diff
@@ -0,0 +1,127 @@
...
+		// Iterate through the rules backwards, preserving the index
+		for idx := len(entryChainRules) - 1; idx > 0; idx-- {
+			entryChainRule := entryChainRules[idx]
+
+			if strings.HasSuffix(entryChainRule, "-j "+c.name) {
+				// Delete rules by index; iptables rules are 1-indexed, but
+				// the output from List includes either the "-N" or "-P" entry,
+				// so this works.
+				if err := ipt.Delete(c.table, entryChain, strconv.Itoa(idx)); err != nil {
+					return fmt.Errorf("Failed to delete referring rule %s %s %d: %v", c.table, entryChain, idx, err)
```
bboreham: This is the sort of code that https://github.com/coreos/go-iptables/pull/36 was intended to avoid.

plugins/meta/portmap/main.go
```diff
@@ -0,0 +1,159 @@
...
+
+type PortMapConf struct {
+	types.NetConf
+	NoSNAT        *bool     `json:"noSnat,omitempty"`
+	ConditionsV4  *[]string `json:"conditionsV4"`
+	ConditionsV6  *[]string `json:"conditionsV6"`
+	RuntimeConfig struct {
+		PortMaps []PortMapEntry `json:"portMappings,omitempty"`
+	} `json:"runtimeConfig,omitempty"`
+	RawPrevResult *map[string]interface{} `json:"prevResult,omitempty"`
```
rosenhouse: this probably doesn't need the `*` since `map` types can be nil.

plugins/meta/portmap/README.md
```diff
@@ -0,0 +1,111 @@
...
+`CNI-DN-xxxxxx` chain (where xxxx is a function of the ContainerID and network name):
+- `-p tcp --dport 8080 -j DNAT --to-destination 172.16.30.2:80`
+- `-p tcp --dport 8043 -j DNAT --to-destination 172.16.30.2:443`
+
+So, new connections to the host will have to traverse all rules, but connections
+directly to the pod will not.
+
+### SNAT
+The SNAT rule rewrites (masquerades) the source address for connections from
+localhost. If this rule did not exist, a connection to `localhost:80` would
```
bboreham: It would help my understanding to note here that the connection being discussed is originating in the host netns, not in any pod netns.

plugins/meta/portmap/portmap.go
```diff
@@ -0,0 +1,266 @@
...
+		return fmt.Errorf("could not teardown ipv4 snat: %v", err)
+	}
+
+	ip6t, err := iptables.NewWithProtocol(iptables.ProtocolIPv6)
+	if err != nil {
+		return err
+	}
+
+	if err := dnatChain.teardown(ip6t); err != nil {
+		return fmt.Errorf("could not teardown ipv6 dnat: %v", err)
```
caseydavenport: Hmm, I think we want a way to disable ip6tables calls.  There are still a number of systems which don't come with ip6 support today (e.g. GCI)

Specifically we're seeing this when we try out the plugin on GCI:

```
NetworkPlugin cni failed to teardown pod \"fluentd-gcp-v2.0-h101m_kube-system\" network: could not teardown ipv6 dnat: exit status 3: ip6tables v1.4.21: can't initialize ip6tables table `nat':
```

Some rather hastily thought out options:
- Emit error messages but continue DEL processing as normal here
- More fine-grained error handling to determine if cleanup failed, or if it's just not supported.
- A switch which can be set to enable/disable ipv6 in the plugin entirely.

plugins/meta/portmap/README.md
```diff
@@ -0,0 +1,111 @@
...
+You should use this plugin as part of a network configuration list. It accepts
+the following configuration options:
+
+* `noSnat` - boolean, default false. If true, do not set up the SNAT chains
+* `conditionsV4`, `conditionsV6` - array of strings. A list of arbitrary `iptables` 
+matches to add to the per-container rule. This may be useful if you wish to 
+exclude specific IPs from port-mapping
+
+The plugin expects to receive the actual list of port mappings via the 
+[`portMappings` capability argument](CONVENTIONS.md#plugin-specific-fields)
```
tomdee: This link is broken

plugins/meta/portmap/README.md
```diff
@@ -0,0 +1,111 @@
...
+of cleanup. So, if a single container exists on IP 172.16.30.2 with ports 
+8080 and 8043 on the host forwarded to ports 80 and 443 in the container, the 
+rules look like this:
+
+`PREROUTING`, `OUTPUT` chains:
+- `--dst-type LOCAL -j CNI-HOSTPORT-DNAT`
+
+`CNI-HOSTPORT-DNAT` chain (note: conditionsV4 goes here):
+- `-j CNI-DN-xxxxxx`
+- `-j CNI-DN-yyyyyy` (one rule per container)
```
tomdee: I'm not really sure that the -yyyyy rules is for (and it's not referenced below).

plugins/meta/portmap/README.md
```diff
@@ -0,0 +1,111 @@
...
+
+`CNI-DN-xxxxxx` chain (where xxxx is a function of the ContainerID and network name):
+- `-p tcp --dport 8080 -j DNAT --to-destination 172.16.30.2:80`
+- `-p tcp --dport 8043 -j DNAT --to-destination 172.16.30.2:443`
+
+So, new connections to the host will have to traverse all rules, but connections
```
tomdee: I don't understand this comment - can you provide a little more detail?

plugins/meta/portmap/utils.go
```diff
@@ -0,0 +1,67 @@
...
+
+import (
+	"crypto/sha512"
+	"fmt"
+	"net"
+
+	"github.com/vishvananda/netlink"
+)
+
+const maxChainLength = 28
```
tomdee: Should this be maxChain*Name*Length, or is that too long

plugins/meta/portmap/README.md
```diff
@@ -0,0 +1,111 @@
+## Port-mapping plugin
+
+This plugin will forward traffic from one or more ports on the host to the
+container. It expects to be run as a chained plugin.
+
+## Usage
+You should use this plugin as part of a network configuration list. It accepts
+the following configuration options:
+
+* `noSnat` - boolean, default false. If true, do not set up the SNAT chains
```
danehans: What do you think about naming this field `snat` (default true) instead of `noSnat`? At the surface, it seems clearer to understand than `noSnat` false.

plugins/meta/portmap/README.md
```diff
@@ -0,0 +1,111 @@
...
+There is a top-level chain, `CNI-HOSTPORT-DNAT` which is always created and
+never deleted. Each plugin execution creates an additional chain for ease
+of cleanup. So, if a single container exists on IP 172.16.30.2 with ports 
+8080 and 8043 on the host forwarded to ports 80 and 443 in the container, the 
+rules look like this:
+
+`PREROUTING`, `OUTPUT` chains:
+- `--dst-type LOCAL -j CNI-HOSTPORT-DNAT`
+
+`CNI-HOSTPORT-DNAT` chain (note: conditionsV4 goes here):
```
danehans: What about conditionsV6?

plugins/meta/portmap/README.md
danehans: What are your thoughts on simply naming this field `snat` with a default true bool? Seems clearer than a double negative of `noSnat` and false.

plugins/meta/portmap/README.md
```diff
@@ -0,0 +1,111 @@
...
+- `-j CNI-DN-xxxxxx`
+- `-j CNI-DN-yyyyyy` (one rule per container)
+
+`CNI-DN-xxxxxx` chain (where xxxx is a function of the ContainerID and network name):
```
danehans: You may want to consider `(where xxxx/yyyy is a function of the ContainerID and network name):`

plugins/meta/portmap/README.md
```diff
@@ -0,0 +1,111 @@
...
+- `-p tcp -s 127.0.0.1 -d 172.16.30.2 --dport 80 -j MASQUERADE`
+- `-p tcp -s 127.0.0.1 -d 172.16.30.2 --dport 443 -j MASQUERADE`
+
+Only new connections from the host, where the source address is 127.0.0.1 but
+not the destination will traverse this chain. It is unlikely that any packets
+will reach these rules without being SNATted, so the cost should be minimal.
+
+Because MASQUERADE happens in POSTROUTING, it means that packets with source ip
+127.0.0.1 need to pass a routing boundary. By default, that is not allowed
+in Linux. So, need to enable the sysctl `net.ipv4.conf.IFNAME.route_localnet`,
```
danehans: I see this setting in kube-proxy, but not in kubelet hostports. Is the cni hostport plugin trying to replicate kube-proxy iptables management?

plugins/meta/portmap/README.md
squeed: I'd wanted to indicate that yyyyy could be a different container, but it doesn't add clarity - removed.

plugins/meta/portmap/README.md
squeed: This plugin is heavily inspired by the rkt hostport plugin.

Rkt has to support the `127.0.0.1:8080` case, but k8s does not.

plugins/meta/portmap/chain.go
squeed: Even better: I fixed the reverse order problem (good catch)

plugins/meta/portmap/chain.go
squeed: Hah.

Yeah, that PR fell down the GH notification black hole.

plugins/meta/portmap/portmap.go
squeed: Also, there is some value in clearing it anways. We'd like this to succeed even if somehow a DEL was missed.

plugins/meta/portmap/portmap.go
squeed: Hmm, good point. Of course, we don't know the pod IPs at delete time, so we can't cover all our bases.

It's probably safe, though, to assume that if `ip6tables` doesn't exist at all, then we don't need to clean up v6 rules. Thoughts?

squeed:
Thanks for the review & feedback; updated. PTAL.

plugins/meta/portmap/README.md
```diff
@@ -0,0 +1,115 @@
...
+
+* `snat` - boolean, default true. If true or omitted, set up the SNAT chains
+* `conditionsV4`, `conditionsV6` - array of strings. A list of arbitrary `iptables` 
+matches to add to the per-container rule. This may be useful if you wish to 
+exclude specific IPs from port-mapping
+
+The plugin expects to receive the actual list of port mappings via the 
+`portMappings` [capability argument](https://github.com/containernetworking/cni/blob/master/CONVENTIONS.md)
+
+So a sample config might look like:
```
gunjan5: what do you think about adding a note or a line here mentioning that the config should have a `.conflist` extension?

plugins/meta/portmap/portmap.go
caseydavenport: I think that's safe to assume, but it's also possible that that the ip6tables tool is present but the required kernel support is not, and we don't want to fail in that case either I think.

I'm not sure there's a good solution - ideally we'd be able to know if ip6tables rules were actually created for this container.

I'd be happy with this being "best-effort" - if it can't init the table then it logs a warning and carries on.

I'd also be happy with a global "ipv6_enabled" option,  but that means people deploying this will need to make a decision.  Seems like ipv6_enabled could even default to false, and the plugin could return an error if it gets ipv6 mappings.

squeed:
@caseydavenport added some ip(6)tables error-ignoring logic, PTAL.

plugins/meta/portmap/portmap.go
```diff
@@ -0,0 +1,294 @@
...
+// So, we first check that iptables is "generally OK" by doing a check. If
+// not, we ignore the error, unless neither v4 nor v6 are OK.
+func unforwardPorts(config *PortMapConf) error {
+	dnatChain := genDnatChain(config.Name, config.ContainerID, nil)
+	snatChain := genSnatChain(config.Name, config.ContainerID)
+
+	ip4t := maybeGetIptables(false)
+	ip6t := maybeGetIptables(true)
+	if ip4t == nil && ip6t == nil {
+		return fmt.Errorf("neither iptables not ip6tables usable")
```
caseydavenport: typo: not

plugins/meta/portmap/portmap.go
```diff
@@ -0,0 +1,294 @@
...
+	if isV6 {
+		proto = iptables.ProtocolIPv6
+	}
+
+	ipt, err := iptables.NewWithProtocol(proto)
+	if err != nil {
+		return nil
+	}
+
+	_, err = ipt.List("filter", "OUTPUT")
```
caseydavenport: So it turns out this check isn't great because it checks the `filter` table, whereas the portmap plugin uses the `nat` table, and it's possible to have the one but not the other :D 

```
# ip6tables -L -t nat
ip6tables v1.6.0: can't initialize ip6tables table `nat': Table does not exist (do you need to insmod?)
Perhaps ip6tables or your kernel needs to be upgraded
# ip6tables -L -t filter
Chain INPUT (policy DROP)
target     prot opt source               destination
ACCEPT     all      anywhere             anywhere
ACCEPT     tcp      anywhere             anywhere             tcp dpt:22

Chain FORWARD (policy DROP)
target     prot opt source               destination

Chain OUTPUT (policy DROP)
target     prot opt source               destination
ACCEPT     all      anywhere             anywhere
```

squeed: Awesome. Looks like v6 nat support was added in kernel 3.7.

Fixed.

bboreham:
Should we consider adding iptables ACCEPT rules for the case where the machine has default deny? https://github.com/kubernetes/kubernetes/issues/35069

squeed:
one last TODO: need to make the teardown code not use indexes, or else we risk stepping on our own toes.

squeed:
And, updated - delete now deletes by rule directly.

caseydavenport:
@squeed LGTM!

bboreham:
LGTM

PR #11, opened 2017-05-27
dhcp: add --pidfile option to DHCP client daemon

dcbw:


squeed:
LGTM - though the manual argument parsing is kind of funny.

dcbw:
I suppose I could move it to flags if you'd prefer.

dcbw:
@squeed sigh... so flags.go doesn't really support -- args, so I guess we go to -.

squeed:
@dcbw it supports both `--` and `-`; it treats them the same.

dcbw:
@squeed updated, PTAL thanks!

plugins/ipam/dhcp/main.go
```diff
@@ -30,7 +32,18 @@ const socketPath = "/run/cni/dhcp.sock"
 
 func main() {
 	if len(os.Args) > 1 && os.Args[1] == "daemon" {
-		runDaemon()
+		var pidfilePath string
+		daemonFlags := flag.NewFlagSet("daemon", flag.ExitOnError)
+		daemonFlags.StringVar(&pidfilePath, "pidfile", "", "optional path to write daemon PID to")
+		if err := daemonFlags.Parse(os.Args[2:]); err != nil {
```
rosenhouse: With `flag.ExitOnError` provided above, this conditional will never trigger, since the `flag` library will call `os.Exit(2)` for you ([see here](https://golang.org/pkg/flag/#ErrorHandling)).  So it's ok to just drop the error in the application code.

Alternately, if you want to customize the error handling, pass `flag.ContinueOnError` into `NewFlagSet()`

dcbw: @rosenhouse thanks, repushed

dcbw:
Travis failure is a portmap plugin failure unrelated to this PR...

rosenhouse:
I kicked Travis and it's green now.  (that portmap plugin test has been flakey, cc @squeed ).  

Anyhow, merging...

squeed:
Yeah, I raised one timeout, but missed another. The flakiness should be fixed in #22

PR #17, opened 2017-06-08
pkg/ns: use unix.Setns() instead of open coding it

tklauser:
Use unix.Setns() instead of manually re-implementing it using
unix.Syscall().

### DIFF
--- a/pkg/ns/ns_linux.go
+++ b/pkg/ns/ns_linux.go
@@ -141,7 +141,7 @@ func (ns *netNS) Set() error {
 		return err
 	}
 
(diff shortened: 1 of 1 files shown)


squeed:
Wonderful! For those playing along at home, it seems Setns was finally added to the unix package in February, 2016

tklauser:
Sorry, should probably have mentioned that in the PR/commit. It was introduced by golang/sys@54535356f1d989a6072b0898d9ddc4dd7009dcd3 (CL: https://go-review.googlesource.com/20054)

Can push an updated PR if necessary.

squeed:
Not needed, IMHO.
LGTM.

PR #22, opened 2017-06-14
portmap: Don't accept a port number of 0

squeed:
This also adds more testing around configuration parsing.

This was inspired by k8s bug that passed a port number of 0, but I realized that there was some logic not under test that really should be.


dcbw:
LGTM

PR #25, opened 2017-06-16
ptp: add ipv6 support

squeed:
* Wait for addresses to leave tentative state before setting routes
* Enable forwarding correctly
* Set up masquerading according to the active protocol

### DIFF
--- a/plugins/main/ptp/ptp.go
+++ b/plugins/main/ptp/ptp.go
@@ -104,12 +104,17 @@ func setupContainerVeth(netns ns.NetNS, ifName string, mtu int, pr *current.Resu
 				return fmt.Errorf("failed to delete route %v: %v", route, err)
 			}
 
+			addrBits := 32
+			if ipc.Version == "6" {
+				addrBits = 128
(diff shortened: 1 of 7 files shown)


plugins/main/ptp/ptp.go
```diff
@@ -84,6 +84,11 @@ func setupContainerVeth(netns ns.NetNS, ifName string, mtu int, pr *current.Resu
 			return err
 		}
 
+		// Wait for addresses to leave tentative state (for up to 10 seconds)
+		if err := ip.SettleAddresses(ifName, 10); err != nil {
```
dcbw: Should we just make ConfigureIface() handle the settling?  That might reduce code in the plugins a bit; we'd still need the Settle call for standalone IP assignment like bridge.go does though.

squeed:
@dcbw updated. I added a stricter testcase that actually shells out to ping (don't hit me). Can you check that I got SettleAddresses right?

dcbw:
LGTM

PR #26, opened 2017-06-17
Set RT_SCOPE_LINK when gateway is nil

lstoll:
There's no gateway, so keep the route scoped to the link. It'll only be relevant for it.

This seems like it would be consistent with iproute2's behaviour https://kernel.googlesource.com/pub/scm/linux/kernel/git/shemminger/iproute2/+/master/ip/iproute.c#1258 . I'm unsure of a case where this behaviour would be be wanted.

(imported from https://github.com/containernetworking/cni/pull/348)

### DIFF
--- a/pkg/ip/route_linux_test.go
+++ b/pkg/ip/route_linux_test.go
@@ -1,0 +1,133 @@
+// Copyright 2016 CNI authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package ip_test
+
+import (
+	"fmt"
+	"net"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+
+	"github.com/containernetworking/plugins/pkg/ip"
+	"github.com/containernetworking/plugins/pkg/ns"
+
+	"github.com/vishvananda/netlink"
+	"github.com/vishvananda/netlink/nl"
+)
+
+var _ = Describe("Link", func() {
+	const (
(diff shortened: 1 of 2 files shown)


squeed:
This seems reasonable, but I'm trying to decipher that iproute code some more. (Can I just say that I hate the Linux coding style. Bah).

Anyways, I think you're right, but what does [this line](https://kernel.googlesource.com/pub/scm/linux/kernel/git/shemminger/iproute2/+/master/ip/iproute.c#1274) mean? It seems to imply that all ipv6 routes are UNIVERSE scoped? Is that right?

pkg/ip/route_linux_test.go
```diff
@@ -0,0 +1,133 @@
...
+			err = ip.AddRoute(vianet, net.ParseIP("1.2.3.4"), containerVeth)
+			Expect(err).To(HaveOccurred())
+
+			err = ip.AddRoute(linknet, nil, containerVeth)
+			Expect(err).NotTo(HaveOccurred())
+
+			err = ip.AddRoute(vianet, net.ParseIP("1.2.3.4"), containerVeth)
+			Expect(err).NotTo(HaveOccurred())
+
+			routes, err := netlink.RouteList(containerVeth, nl.FAMILY_ALL)
```
dcbw: Shouldn't this check the scope of each route to make sure the one added route that should have LOCAL scope actually does?  Not that hard, just loop through `routes` and find the one that's got LinkIndex = containerVeth.Attrs().Index and Dst==linknet and scope == SCOPE_LINK.

pkg/ip/route_linux_test.go
```diff
@@ -0,0 +1,133 @@
...
+		_ = containerNetNS.Do(func(ns.NetNS) error {
+			defer GinkgoRecover()
+			initRoutes, err := netlink.RouteList(containerVeth, nl.FAMILY_ALL)
+			Expect(err).NotTo(HaveOccurred())
+
+			_, linknet, err := net.ParseCIDR("1.2.3.4/32")
+			err = ip.AddHostRoute(linknet, nil, containerVeth)
+			Expect(err).NotTo(HaveOccurred())
+
+			routes, err := netlink.RouteList(containerVeth, nl.FAMILY_ALL)
```
dcbw: Same here, should ensure we have a route with SCOPE_HOST.

lstoll:
This is stale - ended up writing a standalone plugin. Will re-visit if this comes up again, but discarding for now.

PR #28, opened 2017-06-22
pkg,plugins: update for Result struct Interface index changes

dcbw:
It's a pointer now, so we need to use the helper function to set the field and also test for nil before accessing it.

@containernetworking/cni-maintainers

### DIFF
--- a/plugins/meta/portmap/main.go
+++ b/plugins/meta/portmap/main.go
@@ -164,9 +164,14 @@ func parseConfig(stdin []byte, ifName string) (*PortMapConf, error) {
 			}
 
 			// Skip known non-sandbox interfaces
-			intIdx := ip.Interface
-			if intIdx >= 0 && intIdx < len(conf.PrevResult.Interfaces) && conf.PrevResult.Interfaces[intIdx].Name != ifName {
-				continue
+			if ip.Interface != nil {
+				intIdx := *ip.Interface
+				if intIdx >= 0 &&
+					intIdx < len(conf.PrevResult.Interfaces) &&
+					(conf.PrevResult.Interfaces[intIdx].Name != ifName ||
+						conf.PrevResult.Interfaces[intIdx].Sandbox == "") {
+					continue
+				}
 			}
 			switch ip.Version {
 			case "6":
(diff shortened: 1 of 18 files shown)


plugins/meta/portmap/main.go
```diff
@@ -164,7 +164,10 @@ func parseConfig(stdin []byte, ifName string) (*PortMapConf, error) {
 			}
 
 			// Skip known non-sandbox interfaces
-			intIdx := ip.Interface
+			if ip.Interface == nil {
```
squeed: I'm not sure this is right - the interface is nil when up-converting v0.2.0 results, right?

dcbw: @squeed yeah, it would be.  maybe we should upconvert and set index to 0, since there will only ever be a single interface for 0.2.0?

dcbw: @squeed how about something like:

```
			// Skip known non-sandbox interfaces
			if ip.Interface != nil {
				intIdx := *ip.Interface
				if intIdx < 0 ||
					intIdx >= len(conf.PrevResult.Interfaces) ||
					conf.PrevResult.Interfaces[intIdx].Name != ifName ||
					conf.PrevResult.Interfaces[intIdx].Sandbox == "" {
					continue
				}
			}
```

squeed: I still don't think that's right.

Basically, I was trying to make the plugin as flexible as possible (though this might not be a good idea). In other words, it tries to skip an IP if it's definitely sure that it's not on the container. In the upgrade instructions, for example, we say that the Interface can be set to -1 if desired.

If you think the plugin should be stricter, maybe so. But that should probably be a separate PR.

dcbw: @squeed ok, I think you mean this then:

```
	// Skip known non-sandbox interfaces
	if ip.Interface != nil {
		intIdx := *ip.Interface
		if intIdx >= 0 &&
			intIdx < len(conf.PrevResult.Interfaces) &&
			(conf.PrevResult.Interfaces[intIdx].Name != ifName ||
			conf.PrevResult.Interfaces[intIdx].Sandbox == "") {
			continue
		}
	}
```

squeed: Yeah, that's it, thanks!

dcbw:
@squeed PTAL thanks.

PR #30, opened 2017-06-28
plugins/ipam: add IPv6 SLAAC plugin

dcbw:
@containernetworking/cni-maintainers @squeed @danehans

### DIFF
--- a/plugins/ipam/slaac/slaac.go
+++ b/plugins/ipam/slaac/slaac.go
@@ -1,0 +1,157 @@
+// Copyright 2015 CNI authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"fmt"
+	"io/ioutil"
+	"path/filepath"
+	"syscall"
+	"time"
+
+	"github.com/containernetworking/cni/pkg/skel"
+	"github.com/containernetworking/cni/pkg/types"
+	"github.com/containernetworking/cni/pkg/types/current"
+	"github.com/containernetworking/cni/pkg/version"
+	"github.com/containernetworking/plugins/pkg/ns"
+
+	"github.com/vishvananda/netlink"
+)
+
+func setIPV6Sysctl(ifName, sysctl, value string) error {
+	fileName := fmt.Sprintf("/proc/sys/net/ipv6/conf/%s/%s", ifName, sysctl)
+	if err := ioutil.WriteFile(fileName, []byte(value), 0644); err != nil {
+		return fmt.Errorf("failed to toggle %s/%s off: %v", ifName, filepath.Base(fileName), err)
+	}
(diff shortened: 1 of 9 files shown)


dcbw:
@squeed updated with suggestions from the maintainer meeting.  Though my memory was fuzzy, did we agree on just turning off RA once we had the details?

plugins/ipam/slaac/slaac.go
```diff
@@ -0,0 +1,157 @@
...
+			return fmt.Errorf("failed to acquire IPv6 address")
+		}
+
+		// Look for an IPv6 default route through the container interface,
+		// from which we grab the gateway
+		routes, err := netlink.RouteList(link, netlink.FAMILY_V6)
+		if err != nil {
+			return fmt.Errorf("failed to list %q IPv6 routes: %v", link.Attrs().Name, err)
+		}
+		for _, r := range routes {
```
squeed: Do we want to grab any further prefixes? Maybe look for any other routes with a protocol of ra (`RTPROT_RA = 9`)?

squeed: Duh, nevermind, they'll be on-link prefixes and therefore have a corresponding SLAAC address.

Honestly, I should probably take up llama farming.

squeed:
Tested this with a local radvd, got this:
```
failed to add IP addr {Version:6 Interface:2 Address:{IP:2001:470:6d:7bf:8054:98ff:febf:ca38 Mask:ffffffffffffffff0000000000000000} Gateway:fe80::f0d3:13ff:fe73:8c86} to "eth0": file exists
```

dcbw:
> Tested this with a local radvd, got this:

@squeed I guess we turn of RA, remove the addresses and routes, and then pass them back?  To replace an existing address you have to use netlink.AddrReplace() instead of AddrAdd().  Perhaps we should just change ConfigureIface() to use AddrReplace() instead, since it'll add the address if not present, and replace if present?

squeed:
That should work, I would think. One thing to be careful for is whether or not the addresses and routes still have a lifetime.

BTW, I managed to get it to work by - this is extreme - disabling ipv6, disabling slaac, then re-enabling ipv6. I actually have a bridge running radvd with the same prefix as my access network, and I currently have an online container with v6 'net access and no nat! However, something's not right, and the container can't reach other hosts on my local 'net.

dcbw:
@squeed did the AddrReplace/RouteReplace thing

squeed:
reviving this PR.

AddrReplace didn't fix it: `failed to add IP addr {Version:6 Interface:0xc420011388 Address:{IP:2001:470:6d:7bf:d098:e5ff:febb:bae9 Mask:ffffffffffffffff0000000000000000} Gateway:fe80::5cc8:daff:fec8:7e77} to "eth0": file exists`

squeed:
Reviving this PR.

@dcbw I still get the same error: `failed to add IP addr {Version:6 Interface:0xc420011388 Address:{IP:2001:470:6d:7bf:d098:e5ff:febb:bae9 Mask:ffffffffffffffff0000000000000000} Gateway:fe80::5cc8:daff:fec8:7e77} to "eth0": file exists`

benjumanji:
What does thi pr need to push it along?

dcbw:
FWIW, this is blocked because CNI (and some runtimes that use it) have no way to asynchronously indicate changes in the IP addressing back. eg, if a new RA is received that changes the prefix, or DNS info, or routes, that is not able to be pushed back to the runtime, leading to the container state and the runtime state disagreeing. Not good.

squeed:
It's not so disastrous :-).

This PR does SLACC *once*, then disables it on the container interface. That means that we don't necessarily need to support any notion of dynamic addresses.

This just needs to be revived and any bugs squashed

Issue #29, opened 2017-06-27
bridge plugin veth not cleaned up upon ipam address failure 


vzcambria:
When bridge plugin is told that no more addresses are available, a veth isn't cleaned up.

This can be reproduced with a simple test using rangeStart/rangeEnd to simulate e.g. IPAM address exhaustion:

```
$ sudo ip netns add test0
$ sudo ip netns add test1  
$ cat test.conf           
{
  "name": "test",
  "cniVersion": "0.3.1",
      "type": "bridge",
      "bridge": "mynet",
      "ipMasq": true,
      "isGateway": true,
      "ipam": {
      "type": "host-local",
      "subnet": "10.244.10.0/24",
      "rangeStart": "10.244.10.2",
      "rangeEnd": "10.244.10.2",
      "routes": [
          { "dst": "0.0.0.0/0"  }
      ]
      }
}

$ ip addr show | grep veth
$ sudo cnitool add test /var/run/netns/test0
{
    "interfaces": [
        {
            "name": "mynet",
            "mac": "0a:58:0a:f4:0a:01"
        },
        {
            "name": "vethcf0a0b90",
            "mac": "42:80:89:85:27:d6"
        },
        {
            "name": "eth0",
            "mac": "0a:58:0a:f4:0a:02",
            "sandbox": "/var/run/netns/test0"
        }
    ],
    "ips": [
        {
            "version": "4",
            "interface": 2,
            "address": "10.244.10.2/24",
            "gateway": "10.244.10.1"
        }
    ],
    "routes": [
        {
            "dst": "0.0.0.0/0"
        }
    ],
    "dns": {}
}
$ ip addr show | grep veth
23: vethcf0a0b90@if5: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue master mynet state UP group default 
$ sudo cnitool add test /var/run/netns/test1
no IP addresses available in network: test
$ ip addr show | grep veth
23: vethcf0a0b90@if5: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue master mynet state UP group default 
24: veth2d72b1c3@if3: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue master mynet state UP group default 
$ sudo cnitool del test /var/run/netns/test0
$ ip addr show | grep veth
[...]

PR #32, opened 2017-06-28
Change veth allocation to after IPAM.

thanset:
Related to Issue #29, this was replicated and tested with a small script.

```
#!/bin/bash

ip netns del test0
ip netns del test1

export DEBUG=1
export NETCONFPATH=/etc/cni/net.d/
export CNI_PATH=/home/tim/go/src/plugins/bin

ip netns add test0
ip netns add test1

./cnitool add test /var/run/netns/test0

echo "BEFORE"
ip addr show | grep veth
echo ""

./cnitool add test /var/run/netns/test1

echo "AFTER"
ip addr show | grep veth

./cnitool del test /var/run/netns/test0

echo "AFTER DELETION"
ip addr show | grep veth


unset DEBUG
unset NETCONFPATH
unset CNI_PATH
```

Right now the behavior is such that a veth will be allocated even after IPAM exhaustion but it will never be cleaned up on it's own, simply altering the order of operations (Veth allocation after IPAM is called) prevents this veth orphan from ever being created.

### DIFF
--- a/plugins/main/bridge/bridge.go
+++ b/plugins/main/bridge/bridge.go
@@ -33,6 +33,7 @@ import (
 	"github.com/containernetworking/plugins/pkg/ns"
 	"github.com/containernetworking/plugins/pkg/utils"
 	"github.com/vishvananda/netlink"
+	"os"
 )
 
 const defaultBrName = "cni0"
@@ -308,6 +309,15 @@ func enableIPForward(family int) error {
 }
 
 func cmdAdd(args *skel.CmdArgs) error {
+	var success bool = false
+	defer func(ok *bool) {
+		if !*ok {
+			os.Setenv("CNI_COMMAND", "DEL")
+			cmdDel(args)
+			os.Setenv("CNI_COMMAND", "ADD")
+		}
+	}(&success)
+
 	n, cniVersion, err := loadNetConf(args.StdinData)
 	if err != nil {
 		return err
@@ -442,6 +452,8 @@ func cmdAdd(args *skel.CmdArgs) error {
 	brInterface.Mac = br.Attrs().HardwareAddr.String()
 
 	result.DNS = n.DNS
+
+	success = true
 
 	return types.PrintResult(result, cniVersion)
 }


thanset:
Looks like the tests for arch `Linux x86_64` have failed for port mapping, getting an Ip-tables error https://travis-ci.org/containernetworking/plugins/jobs/248013051#L669, looks like it's not finding the chain created by the test?  I'm not sure if this test is flakey?  Anyone have any input regarding this?

squeed:
Hmm, sorry about that. (I wrote the portmap tests). I've kicked the job. Something is flaky about them.

thanset:
@squeed Thanks for the review, I was unaware that was part of the spec-  I will add in tear down logic for this in case of IPAM exhaustion.  cc'ing @vzcambria so he's aware.

squeed:
It is, indeed, barely mentioned in the spec. I'll try and make it clearer.

thanset:
@squeed Updated my branch, I'm simply deleting the `args.IfName` in the namespace as is done in `cmdDel()` via `ns.WithNetNSPath()`.  However, the code is duplicated, would you suggest moving this into a separate function (style wise question- I'm unsure if a small bit of duplication is frowned upon here)?

thanset:
@squeed The approach I took here was to call `cmdDel` if there was an error according to the spec. The particular section I was looking at was:  

```
An ADD action fails, when the runtime decides to handle the failure it should execute the DEL action (in reverse order from the ADD as specified above) for all plugins in the list, even if some were not called during the ADD action.
```

I wrapped this up in the function `cleanUpOnFailure()`.

thanset:
@dcbw  Thank you kindly for the review and more elegant solution.  I've taken your approach and updated the branch.

thanset:
@squeed Can you kick the travis test again to fix the build? Thanks! Seems to be the flakey test.

rosenhouse:
Kicked

squeed:
The test failed because I fixed a flake; can you squash and and rebase your commits on master? Thanks!

thanset:
@squeed Done, squashed commits and rebased on master- but another version of Go on travis is breaking on the same test (Go v1.8 vs v1.7 before), still on amd64.

thanset:
@squeed @dcbw Is there anything else holding this PR back from merge?

Thanks!

squeed:
Changes look good to me - @dcbw, anything to add?

thanset:
@rosenhouse @dcbw @squeed anything else to change before a merge? Thanks!

thanset:
@squeed checking in a last time to see if there's anything else you'd like me to clean up before a merge, thanks.

mauriciovasquezbernal:
@squeed could I know what is the reason it wans't merged?

PR #38, opened 2017-07-05
Add release process

tomdee:
Copied from containernetworking/cni

The only difference is that the release archives also include the the
word "plugin"

### DIFF
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -1,0 +1,35 @@
+# Release process
+
+## Resulting artifacts
+Creating a new release produces the following artifacts:
+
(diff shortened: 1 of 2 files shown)


squeed:
Seems basically good.

One thought: Do we want to make any statement about the version of cni / libcni we'll vendor? It would probably be good to add some kind of manual check that we're always vendoring a named release.

tomdee:
I amended the commit (sorry) to add the s390x platform, and to add a note about vendoring containernetworking/cni

tomdee:
The CI failure isn't caused by this PR

PR #39, opened 2017-07-05
ptp: fix testcase for Interface int-pointer changes

dcbw:


dcbw:
Test failure is portmap flake.  @squeed we need to fix that...

squeed:
post-merge LGTM.

PR #40, opened 2017-07-06
plugins/portmap: fix test flake

squeed:
The source address selection was random, and sometimes we picked a
source address that the container didn't have a route to. Adding a
default route fixes that!

### DIFF
--- a/plugins/meta/portmap/portmap_integ_test.go
+++ b/plugins/meta/portmap/portmap_integ_test.go
@@ -16,6 +16,7 @@ package main
 
 import (
 	"fmt"
+	"math/rand"
 	"net"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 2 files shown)


squeed:
Also, I honestly think I see fewer flakes because I generally push during EU hours.

squeed:
Ah, finally found the real flake.

dcbw:
> Ah, finally found the real flake.

@squeed what was the real flake?

bboreham:
@squeed could you clarify what things are changed here, and what they relate to?  Seems you have changed timeout, changed port, added route, checked address better, maybe other things.

We need the commit message to reflect what is changed, and preferably split different fixes into different commits.

tomdee:
Discussed and we're going to merge this now, and can tidy it up later if needed

PR #42, opened 2017-07-14
bridge: setup ebtables dedup rules if promisc mode is specified.

dnardo:
This is required to prevent duplicate packets when promiscuous mode is set on the bridge interface.
This is related to moving kubernetes "kubenet" functionality out of kubernetes and into the bridge plugin.
https://github.com/kubernetes/kubernetes/issues/38890

Original issue requiring the deduplication: https://github.com/kubernetes/kubernetes/issues/25793
Original Kubernetes PR fixing the issue: https://github.com/kubernetes/kubernetes/pull/28717

### DIFF
--- a/pkg/bridge/bridge.go
+++ b/pkg/bridge/bridge.go
@@ -1,0 +1,100 @@
+// Copyright 2017 CNI authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package bridge
+
+import (
+	"fmt"
+	"strings"
+	"syscall"
+
+	"github.com/containernetworking/plugins/pkg/bridge/ebtables"
+	utilexec "github.com/containernetworking/plugins/pkg/exec"
+)
+
+const (
+	dedupChain  = "CNI-DEDUP"
+	filterTable = "filter"
+	outputChain = "OUTPUT"
+)
+
+// AddDedupRules adds two ebtables rules providing the following:
+// 1) Allow the bridge IP address when using the src bridge mac
+// 2) Block any other IP in the network sourced with the bridge interface's mac
+//    address.
+// For example if the bridge address is 192.16.1.1/24 the following ebtables
+// rules will be setup.
+//  -t filter -A CNI-DEDUP -p IPv4 -s <bridge-mac> -o veth+ --ip-src 192.168.1.1 -j ACCEPT
+//  -t filter -A CNI-DEDUP -p IPv4 -s <bridge-mac> -o veth+ --ip-src 192.168.1.0/24 -j DROP
+func AddDedupRules(mac, ip, nw string, family uint16) error {
+	ebt, err := ebtables.New(utilexec.New())
+	if err != nil {
+		return err
+	}
+	if err := ebt.NewChain(filterTable, dedupChain); err != nil {
+		// Make chain creation idempotent.
+		if !strings.Contains(err.Error(), "already exists") {
+			return err
+		}
+	}
+
+	var common_args []string
+	switch family {
+	case syscall.AF_INET:
+		common_args = []string{"-p", "IPv4", "-s", mac, "-o", "veth+", "--ip-src"}
+	case syscall.AF_INET6:
+		common_args = []string{"-p", "IPv6", "-s", mac, "-o", "veth+", "--ip6-src"}
+	default:
+		return fmt.Errorf("unsupported protocol family %q", family)
+	}
+
+	if err := ebt.AppendUnique(filterTable, dedupChain, append(common_args, ip, "-j", "ACCEPT")...); err != nil {
+		return err
+	}
+	if err := ebt.AppendUnique(filterTable, dedupChain, append(common_args, nw, "-j", "DROP")...); err != nil {
+		return err
+	}
+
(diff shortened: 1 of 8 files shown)


pkg/bridge/bridge.go
```diff
@@ -0,0 +1,89 @@
...
+		return err
+	}
+	if err := ebt.NewChain(filterTable, dedupChain); err != nil {
+		// Make chain creation idempotent.
+		if !strings.Contains(err.Error(), "already exists") {
+			return err
+		}
+	}
+
+	common_args := []string{"-p", "IPv4", "-s", mac, "-o", "veth+", "--ip-src"}
```
dcbw: Hmm, how about IPv6, should we do the same there?

pkg/bridge/ebtables/ebtables.go
```diff
@@ -0,0 +1,180 @@
...
+
+func (e *Error) Error() string {
+	return fmt.Sprintf("exit status %v: %v", e.ExitStatus(), e.msg)
+}
+
+type EBTables struct {
+	path string
+}
+
+func New() (*EBTables, error) {
```
dcbw: Could we pass an 'execer' interface here instead of just using os/exec directly, so that we can mock the whole thing rather than using live ebtables on the host?

pkg/bridge/ebtables/ebtables_test.go
```diff
@@ -0,0 +1,95 @@
...
+// limitations under the License.
+
+package ebtables
+
+import (
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("EBTables Operations", func() {
```
dcbw: Might want to consider running these in their own network namespace so we don't pollute the host's?

squeed:
Might have to add another deb to the .travis.yml file.

dnardo:
@dcbw I think I've address your comments.  Added IPv6 support, and updated the test cases.  I added a lightweight os.exec interface and the test should now run in their own NS.
I wasn't sure if there some particular tests that you wanted mocked out, since the tests I added are mostly e2e/integration tests.

pkg/exec/exec.go
```diff
@@ -0,0 +1,66 @@
...
+// limitations under the License.
+
+package exec
+
+import (
+	"io"
+	osexec "os/exec"
+)
+
+type cmdWrapper osexec.Cmd
```
bboreham: Why do we need this wrapper?  (I.e. the entire file)

squeed:
I know this issue isn't exactly the right place to discuss it, but is this still needed? From what I can gather from kubernetes/kubernetes#20096, this issue is still sometimes happening even with hairpin turned off, and it's not really clear if hairpin was the trigger.

I would love to get rid of the promisc + ebtables fix - they introduce a ton of code complexity. But, if it's needed, it's needed.

pkg/exec/exec.go
dnardo: It was requested that it be wrapped by dcbw, I imagine for testing purposes.

dnardo:
@squeed 
I agree with what you are saying.  We don't know that it was the cause, we do know that for the most part mitigated the issue although.  I'm hesitant to add this code in CNI as well, which is why I've not been pushing hard on this PR.  The task that was given to me was to deprecate "kubenet" and migrate existing bridge plugin work-arounds into the CNI bridge plugin.  That said one direction we could go with, is to just make p2p the default for Kubernetes plugin and don't put the work-arounds into CNI bridge plugin (and remove the already merged promisc stuff).  I'm not, however, comfortable with keeping the bridge plugin the default and not having the work arounds since we should have parity with what is part of kubenet today.

dnardo:
@thockin for his thoughts on this as well.

squeed:
@dnardo awesome, glad to see we're on the same page. However, I do suspect we'll have to add it, given that we already have released a version of bridge with promiscuous. Here's hoping Tim has the magic answer.

dnardo:
@thockin @squeed 
I think at the next sig-network for k8s I'll just float the idea of changing the default plugin to ptp.  That may be the quickest way to get this moving forward.

pkg/bridge/ebtables/ebtables.go
```diff
@@ -0,0 +1,180 @@
...
+	"strconv"
+	"strings"
+	"syscall"
+
+	utilexec "github.com/containernetworking/plugins/pkg/exec"
+	"github.com/containernetworking/plugins/pkg/utils/sysctl"
+)
+
+const (
+	sysctlBridgeCallIPTables = "net/bridge/bridge-nf-call-iptables"
```
rpothier: Can we also set the v6 iptables?
sysctlBridgeCallIP6Tables = "net/bridge/bridge-nf-call-ip6tables"

pkg/bridge/ebtables/ebtables.go
```diff
@@ -0,0 +1,180 @@
...
+	cmd := ebt.exec.Command(modprobe, "br-netfilter")
+	cmd.SetStderr(&stderr)
+
+	if err := cmd.Run(); err != nil {
+		return nil, &Error{(err.(*osexec.ExitError)), stderr.String()}
+	}
+
+	// Allow iptables to see bridged traffic, do this best effort.
+	sysctl.Sysctl(sysctlBridgeCallIPTables, "1")
+
```
rpothier: here also

rpothier: Also, just an implementation question, why is this set here in the plugin and not in the proxier that actually uses it?

m1093782566:
@dnardo @squeed 

From my test, CNI ptp plugin does not resolve the hairpin issue. In other words, container A -> VIP -> container A is not reachable in ptp model by default. 

So, if ptp plugin does no help to hairpin issue, why we don't support veth(not promisc-bridge mode) type hairpin for bridge plugin?

squeed:
@m1093782566 CNI ptp definitely *should* support hairpin. Can you open an issue with a test case?

I've successfully done some hairpin traffic in connection with hostport mapping, so this may be a bad interaction with kube-proxy.

m1093782566:
@squeed , I created an issue: #139.

PR #46, opened 2017-07-24
Greatly reduce build time

aaronjwood:
Before:
```
time ./build.sh 
Building plugins
  flannel
  portmap
  tuning
  bridge
  ipvlan
  loopback
  macvlan
  ptp
  vlan
  dhcp
  host-local
  sample

real	1m3.570s
user	1m55.950s
sys	0m19.707s
```

After:
```
time ./build.sh 
Building plugins
  flannel
  portmap
  tuning
  bridge
  ipvlan
  loopback
  macvlan
  ptp
  vlan
  dhcp
  host-local
  sample

real	0m27.567s
user	2m28.666s
sys	0m25.343s
```

The tests above were done on OS X with an `Intel(R) Core(TM) i7-4770HQ CPU @ 2.20GHz` (4 cores/8 threads) and 16 GB of RAM. I'm sure this will be a bit faster (and take less resources) on Linux since I'm going through the cross compiler here.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -20,12 +20,16 @@ mkdir -p "${PWD}/bin"
 
 echo "Building plugins"
 PLUGINS="plugins/meta/* plugins/main/* plugins/ipam/* plugins/sample"
+ppids=""
 for d in $PLUGINS; do
 	if [ -d "$d" ]; then
 		plugin="$(basename "$d")"
 		echo "  $plugin"
 		go build -o ${PWD}/bin/$plugin -i -pkgdir "$GOPATH/pkg" "$@" "$REPO_PATH/$d" &
+		ppids="$ppids $!"
 	fi
 done
 
-wait
+for ppid in $ppids; do
+	wait $ppid
+done
--- a/build.sh
+++ b/build.sh
@@ -24,7 +24,7 @@ for d in $PLUGINS; do
 	if [ -d "$d" ]; then
 		plugin="$(basename "$d")"
 		echo "  $plugin"
-		go build -o "${PWD}/bin/$plugin" -pkgdir "$GOPATH/pkg" "$@" "$REPO_PATH/$d" &
+		go build -o ${PWD}/bin/$plugin -i -pkgdir "$GOPATH/pkg" "$@" "$REPO_PATH/$d" &
 	fi
 done
 
--- a/build.sh
+++ b/build.sh
@@ -24,13 +24,7 @@ for d in $PLUGINS; do
 	if [ -d "$d" ]; then
 		plugin="$(basename "$d")"
 		echo "  $plugin"
-		# use go install so we don't duplicate work
-		if [ -n "$FASTBUILD" ]
-		then
-			GOBIN=${PWD}/bin go install -pkgdir $GOPATH/pkg "$@" $REPO_PATH/$d
-		else
-			go build -o "${PWD}/bin/$plugin" -pkgdir "$GOPATH/pkg" "$@" "$REPO_PATH/$d" &
-		fi
+		go build -o "${PWD}/bin/$plugin" -pkgdir "$GOPATH/pkg" "$@" "$REPO_PATH/$d" &
 	fi
 done
 
--- a/build.sh
+++ b/build.sh
@@ -29,7 +29,9 @@ for d in $PLUGINS; do
 		then
 			GOBIN=${PWD}/bin go install -pkgdir $GOPATH/pkg "$@" $REPO_PATH/$d
 		else
-			go build -o "${PWD}/bin/$plugin" -pkgdir "$GOPATH/pkg" "$@" "$REPO_PATH/$d"
+			go build -o "${PWD}/bin/$plugin" -pkgdir "$GOPATH/pkg" "$@" "$REPO_PATH/$d" &
 		fi
 	fi
 done
+
+wait


squeed:
Interesting :-)

If you look, I added a FASTBUILD var which uses `go install` instead, so that all of the libraries are only compiled once. It too is pretty fast, especially for subsequent builds. I just didn't turn it on for release builds in an abundance of caution.

aaronjwood:
Ah, just noticed that. One thing I ran into with `$FASTBUILD` on OS X:
```
./build.sh 
Building plugins
  flannel
go install: cannot install cross-compiled binaries when GOBIN is set
```
Is there any reason to use `go install` now if everything is done in parallel? The Go devs talk about sticking to `go build -o` when cross-compiling https://github.com/golang/go/issues/11778#issuecomment-122545763

Might be nicer to have one fast way to build that works on all platforms that doesn't require a bunch of conditionals :)

aaronjwood:
Any interest in making this change? This complements a nice CI setup with decent hardware too!

build.sh
```diff
@@ -24,12 +24,8 @@ for d in $PLUGINS; do
...
-		then
-			GOBIN=${PWD}/bin go install -pkgdir $GOPATH/pkg "$@" $REPO_PATH/$d
-		else
-			go build -o "${PWD}/bin/$plugin" -pkgdir "$GOPATH/pkg" "$@" "$REPO_PATH/$d"
-		fi
+		go build -o "${PWD}/bin/$plugin" -pkgdir "$GOPATH/pkg" "$@" "$REPO_PATH/$d" &
 	fi
 done
+
+wait
```
bboreham: The function of `wait` is described as follows in `man sh`:
> wait for all jobs to complete and the return an exit status of zero.

We would need to know if any of the background processes failed.

bboreham:
> The Go devs talk about sticking to `go build -o`

looks like that conversation is about binaries, not libraries.

bboreham:
Thinking that changing `go install` to `go build -i` should solve the error message you hit

build.sh
aaronjwood: They should still be reported and not cause any stalls. For example:
```
echo "test" &
sleep 2
ls /t &
sleep 5 &
echo "test"
wait 
```
`ls /t &` will fail for me but everything is still waited upon (and executed) as expected.

aaronjwood:
Do you want to stick with `go install -i` then instead of `go build -o`? Or do you still want the original conditional that does either an install or build depending on the logic?

squeed:
We can get rid of the branch and standardize on `go build -i` without the backgrounding; I'd added it to scratch an itch but it was always superfluous.

aaronjwood:
Sure, I added in `-i` but I'd like to propose that we keep the parallelism. It's still handy for the first build and for CI environments where a clean build is always done (or when the workspace is cleaned up on some schedule/trigger). Also, after the first build (once you have your libs already compiled) it's still faster :)

Without parallelism:
```
time ./build.sh 
Building plugins
  flannel
  portmap
  tuning
  bridge
  ipvlan
  loopback
  macvlan
  ptp
  vlan
  dhcp
  host-local
  sample

real	0m4.949s
user	0m3.831s
sys	0m1.732s
```

With parallelism:
```
time ./build.sh 
Building plugins
  flannel
  portmap
  tuning
  bridge
  ipvlan
  loopback
  macvlan
  ptp
  vlan
  dhcp
  host-local
  sample

real	0m1.520s
user	0m5.603s
sys	0m2.949s
```

build.sh
bboreham: we want to get the failed status code reported up to the caller of `build.sh`

build.sh
```diff
@@ -24,12 +24,8 @@ for d in $PLUGINS; do
...
-		then
-			GOBIN=${PWD}/bin go install -pkgdir $GOPATH/pkg "$@" $REPO_PATH/$d
-		else
-			go build -o "${PWD}/bin/$plugin" -pkgdir "$GOPATH/pkg" "$@" "$REPO_PATH/$d"
-		fi
+		go build -o ${PWD}/bin/$plugin -i -pkgdir "$GOPATH/pkg" "$@" "$REPO_PATH/$d" &
 	fi
 done
+
+wait
```
bboreham: The function of `wait` is described as follows in `man sh`:
> wait for all jobs to complete and the return an exit status of zero.

We would need to know if any of the background processes failed.
[Here](https://github.com/weaveworks/weave/blob/dff5b6d3f113105311fd5178ea5845150aba3360/test/setup.sh#L56) is an example how this can be done

aaronjwood: I've adjusted the script so it should do just that.

rosenhouse:
For what it's worth, this runs slower on my 2-core VirtualBox.

rosenhouse:
Running this in VirtualBox on my Mac (using the local `Vagrantfile`):

```
./build.sh
Building plugins
  flannel
  portmap
  tuning
  bridge
  ipvlan
  loopback
  macvlan
  ptp
  vlan
  dhcp
  host-local
  sample
go install internal/cpu: build output "/go/src/github.com/containernetworking/plugins/gopath/pkg/internal/cpu.a" already exists and is not an object file
go install internal/cpu: build output "/go/src/github.com/containernetworking/plugins/gopath/pkg/internal/cpu.a" already exists and is not an object file
go install internal/race: build output "/go/src/github.com/containernetworking/plugins/gopath/pkg/internal/race.a" already exists and is not an object file
go install sync/atomic: build output "/go/src/github.com/containernetworking/plugins/gopath/pkg/sync/atomic.a" already exists and is not an object file
# io
/usr/local/go/src/io/pipe.go:12:2: can't open import: "sync": open /go/src/github.com/containernetworking/plugins/gopath/pkg/sync.a: no such file or directory
# os
/usr/local/go/src/os/exec.go:12:2: import /go/src/github.com/containernetworking/plugins/gopath/pkg/time.a: reading input: EOF
# fmt
/usr/local/go/src/fmt/print.go:11:2: can't open import: "reflect": open /go/src/github.com/containernetworking/plugins/gopath/pkg/reflect.a: no such file or directory
go install unicode/utf16: build output "/go/src/github.com/containernetworking/plugins/gopath/pkg/unicode/utf16.a" already exists and is not an object file
```
```
echo $?
2
```

rosenhouse:
Also, Travis CI only uses a single core.  With this change, CI actually runs a bit *slower*.
Example: this PR [build on s390](https://travis-ci.org/containernetworking/plugins/jobs/296367445) takes 2:42, whereas [the same architecture on `master`](https://travis-ci.org/containernetworking/plugins/jobs/302327541) takes 2:14.

squeed:
Indeed, we chatted about this and decided that the change to the `go build` command is good, we just would rather not have the parallelism. Would you mind making that change, then we can merge this? Thanks!

dcbw:
@squeed @aaronjwood turns out that Go 1.10+ fix this issue for us in a different way (by caching everything during the build process).  So we'd prefer to just suggest using Go 1.10+ and to remove the FASTBUILD stuff, rather than waiting for each plugin's build to complete manually.

Issue #45, opened 2017-07-23
Host-local should support disjoint sets of ranges.


selvakumarnms:
Hi ,
I was testing the multiple ranges support for the IP and looks like it tries to get the IP from all the subnets at a time rather than one by one range.

Is this the expected behaviour ?

echo '{
"cniVersion": "0.3.1",
"name": "examplenet",
"ipam": {
"type": "host-local",
"ranges": [{
"subnet": "10.10.10.0/24",
"rangeStart": "10.10.10.20",
"rangeEnd": "10.10.10.22"

}, {
"subnet": "20.20.20.0/24",
"rangeStart": "20.20.20.20",
"rangeEnd": "20.20.20.21"

}],
"dataDir": "/tmp/cni-example"
}
}' | CNI_COMMAND=ADD CNI_CONTAINERID=example CNI_NETNS=/dev/null CNI_IFNAME=dummy0 CNI_PATH=. ./host-local

selvakumarnms:
My expectation would be first 10.10.10. network should be exhausted and then it should give ip from 20.20.20.20 network rather than at a time allocating ip

squeed:
This is the intended behavior as written. The initial goal was to support dual-stack networks, so the plugin treats each range independently.

I haven't really seen the need for a single "pool" of disparate ranges - can you provide a use-case?

dcbw:
@squeed we actually have one with OpenShift, sometimes the size of the subnet that people can use isn't that large so they want to use a couple on a given node.  Like say they use 172.16.x.x but there are some legacy systems that use some /24s, etc.

squeed:
Thought for a bit about the configuration format. Couldn't really come up with anything particularly clever:

```json
{
    "ipam": {
        "type": "host-local",
        "ranges": [
            [
                        {"subnet": "10.0.0.0/24"},
                        {"subnet": "172.16.5.0/24"}
            ],
            [
                        {"subnet": "2001:db8:96:abd::/64"}
            ]
        ]
    }
}
```

I don't love it, but it's OK. One question is if we're going to want to add some sort of "pool-level" property. It will add a lot of verbosity though.

squeed:
Hah, @dcbw, this will break our cute last_allocated_ip filename prefix scheme (we used an encoding of the first IP).

I might switch it back to index-based.

PR #50, opened 2017-07-31
ipam/host-local: support sets of disjoint ranges

squeed:
In real-world address allocations, disjoint address ranges are common. Therefore, the host-local allocator should support them.

This change still allows for multiple IPs in a single configuration, but also allows for a "set of subnets."

I'm not entirely happy with the configuration format - currently an array of arrays. I could be convinced to write a custom JSON deserializer that accepts either an array or a single object.

Fixes: #45

### DIFF
--- a/plugins/ipam/host-local/backend/allocator/allocator.go
+++ b/plugins/ipam/host-local/backend/allocator/allocator.go
@@ -15,11 +15,11 @@
 package allocator
 
 import (
-	"encoding/base64"
 	"fmt"
 	"log"
 	"net"
 	"os"
+	"strconv"
 
 	"github.com/containernetworking/cni/pkg/types/current"
 	"github.com/containernetworking/plugins/pkg/ip"
@@ -27,29 +27,16 @@ import (
 )
 
 type IPAllocator struct {
-	netName string
-	ipRange Range
-	store   backend.Store
-	rangeID string // Used for tracking last reserved ip
+	rangeset *RangeSet
+	store    backend.Store
+	rangeID  string // Used for tracking last reserved ip
 }
 
-type RangeIter struct {
-	low   net.IP
-	high  net.IP
-	cur   net.IP
-	start net.IP
-}
-
-func NewIPAllocator(netName string, r Range, store backend.Store) *IPAllocator {
-	// The range name (last allocated ip suffix) is just the base64
-	// encoding of the bytes of the first IP
-	rangeID := base64.URLEncoding.EncodeToString(r.RangeStart)
-
+func NewIPAllocator(s *RangeSet, store backend.Store, id int) *IPAllocator {
 	return &IPAllocator{
-		netName: netName,
-		ipRange: r,
-		store:   store,
-		rangeID: rangeID,
+		rangeset: s,
+		store:    store,
+		rangeID:  strconv.Itoa(id),
 	}
 }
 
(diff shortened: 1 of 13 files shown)


plugins/ipam/host-local/backend/allocator/allocator.go
```diff
@@ -151,39 +150,67 @@ func (a *IPAllocator) GetIter() (*RangeIter, error) {
...
-		i.start = a.ipRange.RangeStart
+		iter.rangeIdx = 0
+		iter.startRange = 0
+		iter.startIP = (*a.pool)[0].RangeStart
 	}
-	return &i, nil
+	return &iter, nil
 }
 
 // Next returns the next IP in the iterator, or nil if end is reached
```
dcbw: Update the comment now that the return args have changed?

plugins/ipam/host-local/backend/allocator/allocator.go
```diff
@@ -151,39 +150,67 @@ func (a *IPAllocator) GetIter() (*RangeIter, error) {
...
-	//  we returned .high last time, since we're inclusive
-	if i.cur.Equal(i.high) {
-		i.cur = i.low
+
+	// If we've reached the end of this range, we need to advance the range
+	// RangeEnd is inclusive as well
+	if i.cur.Equal(r.RangeEnd) {
+		i.rangeIdx += 1
+		i.rangeIdx %= len(*i.pool)
+		r = (*i.pool)[i.rangeIdx]
```
dcbw: What guarantees that we don't index out-of-range here?  eg that when we have exhausted all IPs in all ranges in the pool, that this line doesn't crash?  Shouldn't we check len(*i.pool) or something?

plugins/ipam/host-local/backend/allocator/pool.go
```diff
@@ -0,0 +1,99 @@
...
+		if r.Contains(addr) == nil {
+			return nil
+		}
+	}
+
+	return fmt.Errorf("%s not in range set", addr.String())
+}
+
+// RangeFor finds the range that contains an IP, or nil if not found
+func (p *Pool) RangeFor(addr net.IP) *Range {
```
dcbw: Should we canonicalize the IP here too?  If so, could just implement Contains() on top of RangeFor()

plugins/ipam/host-local/backend/allocator/pool_test.go
```diff
@@ -0,0 +1,36 @@
...
+		p := Pool{
+			Range{Subnet: mustSubnet("192.168.0.0/24")},
+			Range{Subnet: mustSubnet("172.16.1.0/24")},
+		}
+
+		err := p.Canonicalize()
+		Expect(err).NotTo(HaveOccurred())
+
+		Expect(p.Contains(net.IP{192, 168, 0, 55})).To(BeNil())
+	})
```
dcbw: Maybe some overlap tests and pool-family-consistency checks?

plugins/ipam/host-local/backend/allocator/allocator.go
squeed: Pool exhaustion and end-of-range are not the same, since we can start in the "middle" of the range set. This code wont out-of-range since we wrap it back around with a modulo.

During Canonicalize, we reject empty pools.

plugins/ipam/host-local/backend/allocator/pool.go
squeed: Yeah, that would be a lot neater.

plugins/ipam/host-local/backend/allocator/pool_test.go
squeed: Good call.

squeed:
@dcbw updated, PTAL. Thanks!

plugins/ipam/host-local/backend/allocator/config.go
```diff
@@ -113,9 +115,10 @@ func LoadIPAMConfig(bytes []byte, envArgs string) (*IPAMConfig, string, error) {
 	numV6 := 0
 	for i, _ := range n.IPAM.Ranges {
 		if err := n.IPAM.Ranges[i].Canonicalize(); err != nil {
-			return nil, "", fmt.Errorf("Cannot understand range %d: %v", i, err)
+			return nil, "", fmt.Errorf("invalid range set %d: %s", i, err)
 		}
-		if len(n.IPAM.Ranges[i].RangeStart) == 4 {
+
+		if len(n.IPAM.Ranges[i][0].RangeStart) == 4 {
```
dcbw: Maybe we should change this to To4()... I know we canonicalize the start so it'll be len=4 for v4, but I don't think we use len() anywhere else for v4/v6 checks (though we do use it later for canonicalization and range checking, but only to compare lengths to each other).

squeed:
@dcbw Updated. The `Pool` object is now a `RangeSet` object. Clarified the README a bit.

PR #56, opened 2017-08-21
Don't run tests in parallel

squeed:
Go runs tests in parallel by default, one of the tests was switching namespaces.

Turn of parallelism.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,8 +21,8 @@ matrix:
 
 script:
  - |
-   if [ "${TARGET}" == "amd64" ]; then
-     GOARCH="${TARGET}" ./test.sh
+   if [ "${TARGET}" == $( go env GOARCH ) ]; then
+     FASTBUILD=1 GOARCH="${TARGET}" ./test.sh
    else
      GOARCH="${TARGET}" ./build.sh
    fi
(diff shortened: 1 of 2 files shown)


squeed:
Woah, something is clearing all the iptables rules in the whole travis container. Hm.

rosenhouse:
Not sure how I missed this before.  My understanding is that the `go test` runner uses multiple goroutines in a single OS process, but the `ginkgo` test runner, when run with the `-p` flag, actually [creates a separate OS process for each "parallel node"](http://onsi.github.io/ginkgo/#parallel-specs), and then distributes test cases across those nodes.  This means that each node runs one test at a time, and in theory could have namespace affinity that doesn't affect other nodes.

Therefore, I'd suggest we try using `ginkgo -p` to run tests first.  That may address this namespace flakiness thing without forcing us down to non-parallel

.travis.yml
```diff
@@ -21,8 +21,8 @@ matrix:
 
 script:
  - |
-   if [ "${TARGET}" == "amd64" ]; then
-     GOARCH="${TARGET}" ./test.sh
+   if [ "${TARGET}" == $( go env GOARCH ) ]; then
```
bboreham: Could you explain how this change relates to the overall thrust of the PR?

squeed:
No longer needed.

PR #61, opened 2017-08-25
Update README.md

dengyi1996:
I noticed that if I use above example in a 3 worknodes cluster. Some pod will have same IP address, so I add this notice


squeed:
I'm not sure this is necessary or the right place; anyone else?

rosenhouse:
Yeah, this guidance:
- assumes that IPAM config is static, which isn't always true
- isn't scoped just to the `macvlan` plugin -- it could apply to basically everything
- isn't always applicable: sometimes an operator wants to re-use IP address space

@dengyi1996 you may be highlighting a different deficiency-- our lack of good "getting started" documentation for CNI and our core plugins.  If you'd like to start something like that, feel free to open a separate PR.   I don't think this patch is quite right.

Issue #60, opened 2017-08-24
bridge: mac address unstable for v6-only bridges


squeed:
We set the bridge's mac address (a.k.a. bridge id) based on a deterministic permutation of the bridge's v4 address. When the bridge doesn't have a v4 address, there is no explicit mac set, so it chooses the lowest mac on the bridge. This changes as containers come and go.

We probably should just set the mac to itself in all cases and stop changing it.

Issue #59, opened 2017-08-24
Don't touch ipv6 sysctls when there are no v6 addresses


squeed:
Sometimes ipv6 is disabled (e.g. via kernel command line)

squeed:
This was fixed.

qrpike:
What version of CNI was this released in? I have ipv6 disabled at the kernel. I'm am using CNI `v0.6.0` 
 with flannel `v0.10.0` and I'm getting the following:

```
Jan 29 22:42:34 blade02 kubelet-wrapper[2062]: E0130 03:42:34.547096    2062 remote_runtime.go:92] RunPodSandbox from runtime service failed: rpc error: code = Unknown desc = NetworkPlugin cni failed to set up pod "node-problem-detector-xqp5k_kube-system" network: open /proc/sys/net/ipv6/conf/eth0/accept_dad: no such file or directory
Jan 29 22:42:34 blade02 kubelet-wrapper[2062]: E0130 03:42:34.547148    2062 kuberuntime_sandbox.go:54] CreatePodSandbox for pod "node-problem-detector-xqp5k_kube-system(7d585c3f-fbee-11e7-b6ee-1eb535d7075a)" failed: rpc error: code = Unknown desc = NetworkPlugin cni failed to set up pod "node-problem-detector-xqp5k_kube-system" network: open /proc/sys/net/ipv6/conf/eth0/accept_dad: no such file or directory
Jan 29 22:42:34 blade02 kubelet-wrapper[2062]: E0130 03:42:34.547165    2062 kuberuntime_manager.go:647] createPodSandbox for pod "node-problem-detector-xqp5k_kube-system(7d585c3f-fbee-11e7-b6ee-1eb535d7075a)" failed: rpc error: code = Unknown desc = NetworkPlugin cni failed to set up pod "node-problem-detector-xqp5k_kube-system" network: open /proc/sys/net/ipv6/conf/eth0/accept_dad: no such file or directory
[...]

Issue #52, opened 2017-08-02
bridge plugin should not disable IPv6 DAD in certain cases


dcbw:
1) when hairpin mode is not enabled, DAD should not be disabled
2) when enhanced_dad is available in the kernel, that should be used instead of disabling DAD

karampok:
We have similar issues:

Because of this [line](https://github.com/containernetworking/plugins/blob/20bc33abc565e75c0f9de488a020d1ef5b7d9b56/plugins/main/bridge/bridge.go#L299), the plugin errors. That file does not even exists in our only IPv4 stack.

It would also make sense to skip any IPv6 related tasks when the machine uses only IPv4.

squeed:
@karampok filed separate issue #59

squeed:
This is fixed.

PR #62, opened 2017-08-25
bridge: various fixes

squeed:
* Don't set the MAC, send gratuitous arp instead
* Set the bridge's MAC to itself
* Only disable DAD when necessary

Fixes: #60 #59 #52

### DIFF
--- a/plugins/main/bridge/bridge.go
+++ b/plugins/main/bridge/bridge.go
@@ -32,6 +32,7 @@ import (
 	"github.com/containernetworking/plugins/pkg/ipam"
 	"github.com/containernetworking/plugins/pkg/ns"
 	"github.com/containernetworking/plugins/pkg/utils"
+	"github.com/j-keck/arping"
 	"github.com/vishvananda/netlink"
 )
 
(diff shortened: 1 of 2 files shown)


rosenhouse:
Is there any way we can add test coverage for these changes?  

I've been thinking about refactoring this plugin, but its scarier to attempt when there are important features that might break while the tests remain green.

squeed:
Yeah, good point. Most of this should be testable.

squeed:
@rosenhouse added a test for the "stable mac" case. Not sure how to reasonably test the other changes.

rosenhouse:
@squeed looks like it went red

squeed:
Once #56 is merged we should see fewer test failures.

A lot of these tests are dancing around with namespaces, making them non-deterministic.

rosenhouse:
Yeah, I don't see how to add functional tests either.  (all the more reason to refactor to a more dependency-injected style where we can have unit tests of the application-object).

LGTM

wanghaibo:
can we set the MAC（GenerateByIp） in container and send gratuitous arp at the same time? in some case,we need container mac address stable when ip not change

PR #63, opened 2017-08-28
Fix ipmasq teardown on v6-only interfaces

squeed:
V6-only interfaces were throwing an error on teardown.


pkg/ip/link.go
```diff
@@ -158,6 +158,9 @@ func SetupVeth(contVethName string, mtu int, hostNS ns.NetNS) (net.Interface, ne
 func DelLinkByName(ifName string) error {
 	iface, err := netlink.LinkByName(ifName)
 	if err != nil {
+		if err.Error() == "Link not found" {
```
dcbw: Could you test instead for netlink.LinkNotFoundError instead of matching the string?

dcbw:
LGTM

PR #64, opened 2017-08-29
travis: run with ginkgo -p instead of go test

rosenhouse:
may help reduce test-pollution due to namespace-affinity

see http://onsi.github.io/ginkgo/#parallel-specs

ref: https://github.com/containernetworking/plugins/pull/56#issuecomment-325675507

cc: @squeed

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ go:
 
 env:
   global:
-  - PATH=$GOROOT/bin:$PATH
+  - PATH=$GOROOT/bin:$GOPATH/bin:$PATH
   matrix:
   - TARGET=amd64
   - TARGET=arm
@@ -18,6 +18,9 @@ env:
 
 matrix:
   fast_finish: true
+
+install:
+  - go get github.com/onsi/ginkgo/ginkgo
 
 script:
  - |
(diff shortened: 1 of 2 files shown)


squeed:
Good catch. I wish ginkgo detected a test that ended in a different namespace.

rosenhouse:
This is red on account of differences between how test packages are specified to `go test` and `ginkgo`.

My current plan to fix this is:
- #65: Upgrade to Go 1.9 so that we get the benefits of [ignoring `vendor` by default](https://golang.org/doc/go1.9#vendor-dotdotdot)
- #66: fix up lingering `go vet` warnings in `host-local/backend/disk`
- #67: just test `./...` instead of a long list of packages that could get stale
- finally return to this and use `ginkgo -p` instead of `go test`

squeed:
Sounds like a good plan. Want to merge #56 in the meantime?

rosenhouse:
@squeed updated now that the `go vet` and `TESTABLE` things are merged.  I think this is good to go.

This went green twice in a row.  So at least it's no worse than `go test` !

rosenhouse:
Aaaand it's red: https://travis-ci.org/containernetworking/plugins/jobs/270394612

I may try to look at this more today.  I'd really like to figure out how to run the tests in parallel without pollution.

PR #65, opened 2017-08-30
Golang versions: add 1.9, drop 1.7

rosenhouse:
Go 1.9 is out, 1.7 is now no longer maintained.

Reference: https://golang.org/doc/devel/release.html#policy

See also: https://github.com/containernetworking/cni/pull/499

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,8 +3,8 @@ sudo: required
 dist: trusty
 
 go:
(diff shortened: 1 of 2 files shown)


squeed:
We discussed this in the meeting; let's let this sit for a bit in case we need to cut a 0.6.1;

PR #66, opened 2017-08-30
host-local FileLock is used by value instead of by reference

rosenhouse:
Without this change

```
go vet ./plugins/ipam/host-local/backend/disk
```

returns
```
plugins/ipam/host-local/backend/disk/backend.go:54: literal copies lock value from *lk: disk.FileLock
exit status 1
```

see the [go vet heuristic here](https://github.com/golang/go/blob/d77d4f509c61a8e1eaadeabca86ee48710bd8030/src/cmd/vet/copylock.go#L230-L232).

### DIFF
--- a/plugins/ipam/host-local/backend/disk/backend.go
+++ b/plugins/ipam/host-local/backend/disk/backend.go
@@ -31,7 +31,7 @@ var defaultDataDir = "/var/lib/cni/networks"
 // Store is a simple disk-backed store that creates one file per IP
 // address in a given directory. The contents of the file are the container ID.
 type Store struct {
-	FileLock
+	*FileLock
 	dataDir string
 }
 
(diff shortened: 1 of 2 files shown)


squeed:
Go vet is complaining about something that looks unsafe, but totally is. We're copying the lock value exactly once, which is safe in this case because it's unused, but might not be safe in all cases.

Also: non-pointer receivers in golang should basically never be used. They also result in call-by-value duplication. I don't understand how that passes govet.

In any case, the real fix is to change like 34 to type `*FileLock`.

plugins/ipam/host-local/backend/disk/backend.go
```diff
@@ -51,7 +51,7 @@ func New(network, dataDir string) (*Store, error) {
 	if err != nil {
 		return nil, err
 	}
-	return &Store{*lk, dir}, nil
+	return &Store{lk, dir}, nil
```
squeed: Store should embed `*FileLock` instead.

rosenhouse:
@squeed good point, updated.

PR #71, opened 2017-09-02
portmap integration test: echo server runs in separate process

rosenhouse:
this way we're not mixing goroutines and namespaces

when running 
```
ginkgo -untilItFails -timeout 3s -p
```
with the single integration spec focused, it used to fail after about 5 runs.  with this change, i got 95 successes in a row.

however, this won't fix everything.   when I unfocus the integration spec, I still get failures due to namespace affinity, but now after about 10 runs.

~~UPDATE: travis says `"nsenter": executable file not found in $PATH` :-(~~

Update: it's green! 🍏 💚 🥗 !

### DIFF
--- a/plugins/meta/portmap/echosvr/echosvr_test.go
+++ b/plugins/meta/portmap/echosvr/echosvr_test.go
@@ -1,0 +1,74 @@
+package main_test
+
+import (
+	"fmt"
+	"io/ioutil"
+	"net"
+	"os/exec"
+	"strings"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	"github.com/onsi/gomega/gbytes"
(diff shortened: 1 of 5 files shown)


squeed:
Hard to believe that we don't have `util-linux`, but I guess you have to apt-get it?

rosenhouse:
Turns out `util-linux` on Ubuntu 14.04 doesn't have `nsenter`.  So rather than building it from source, I used `ip netns exec`.

Anyhow, it's green now.

squeed:
Looks fine to me.

If you wanted to be _super mega awesome_, you could move it to testutils - seems like a handy thing to have around. But that's a bonus suggestion. LGTM regardless.

dcbw:
LGTM.

PR #73, opened 2017-09-12
Newly added `host-device` doesn't compile.

Random-Liu:
`host-device` is using non-exist packages.

Signed-off-by: Lantao Liu <lantaol@google.com>

### DIFF
--- a/plugins/host-device/host-device.go
+++ b/plugins/host-device/host-device.go
@@ -24,9 +24,9 @@ import (
 	"runtime"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


trusch:
I'm sorry for that, I will figure out and fix the root source of the problem as soon as possible.

I wrote this new plugin a few month ago and did not rebased it recently.

squeed:
Ah, good catch, sorry.

@trusch, could you add a README in the plugin directory too?

trusch:
@sqeed Yes I'll do that. I hope that I will find time this week.

PR #74, opened 2017-09-13
Host-device fixes

rosenhouse:
Fixes errors in the `host-device` plugin, and its test, which were introduced in #3.

Includes change from #73.

@containernetworking/cni-maintainers a couple other issues that I encountered while fixing this:
- should this plugin, and loopback, go to the trouble of returning a `cniVersion` that matches the config version?  Since neither one returns any kind of interesting result, I say no.  But its technically a spec violation...
- what should be the meaning of the `CNI_IFNAME` environment variable in this plugin?    The plugin ignores it today... but it could potentially substitute for the `device` config field.

Anyhow, this should at least get CI green again.

### DIFF
--- a/plugins/host-device/host-device.go
+++ b/plugins/host-device/host-device.go
@@ -24,9 +24,9 @@ import (
 	"runtime"
 	"strings"
 
-	"github.com/containernetworking/cni/pkg/ns"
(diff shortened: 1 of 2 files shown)


squeed:
Thanks, and sorry for the premature merge. I'll merge this now, and we can follow up.

PR #77, opened 2017-10-07
host-local: Update host-local IPAM to support Windows

rakelkar:
Get host-local to work for Windows. This is required for Flannel VXLAN mode on Windows.

### DIFF
--- a/plugins/ipam/host-local/backend/disk/backend.go
+++ b/plugins/ipam/host-local/backend/disk/backend.go
@@ -22,6 +22,7 @@ import (
 	"strings"
 
 	"github.com/containernetworking/plugins/plugins/ipam/host-local/backend"
+	"runtime"
 )
 
 const lastIPFilePrefix = "last_reserved_ip."
@@ -55,7 +56,8 @@ func New(network, dataDir string) (*Store, error) {
 }
 
 func (s *Store) Reserve(id string, ip net.IP, rangeID string) (bool, error) {
-	fname := filepath.Join(s.dataDir, ip.String())
+	fname := GetEscapedPath(s.dataDir, ip.String())
+
 	f, err := os.OpenFile(fname, os.O_RDWR|os.O_EXCL|os.O_CREATE, 0644)
 	if os.IsExist(err) {
 		return false, nil
@@ -73,7 +75,7 @@ func (s *Store) Reserve(id string, ip net.IP, rangeID string) (bool, error) {
 		return false, err
 	}
 	// store the reserved ip in lastIPFile
-	ipfile := filepath.Join(s.dataDir, lastIPFilePrefix+rangeID)
+	ipfile := GetEscapedPath(s.dataDir, lastIPFilePrefix+rangeID)
 	err = ioutil.WriteFile(ipfile, []byte(ip.String()), 0644)
 	if err != nil {
 		return false, err
@@ -83,7 +85,7 @@ func (s *Store) Reserve(id string, ip net.IP, rangeID string) (bool, error) {
 
 // LastReservedIP returns the last reserved IP if exists
 func (s *Store) LastReservedIP(rangeID string) (net.IP, error) {
-	ipfile := filepath.Join(s.dataDir, lastIPFilePrefix+rangeID)
+	ipfile := GetEscapedPath(s.dataDir, lastIPFilePrefix+rangeID)
 	data, err := ioutil.ReadFile(ipfile)
 	if err != nil {
 		return nil, err
@@ -92,7 +94,7 @@ func (s *Store) LastReservedIP(rangeID string) (net.IP, error) {
 }
 
 func (s *Store) Release(ip net.IP) error {
-	return os.Remove(filepath.Join(s.dataDir, ip.String()))
+	return os.Remove(GetEscapedPath(s.dataDir, ip.String()))
 }
 
 // N.B. This function eats errors to be tolerant and
@@ -114,4 +116,11 @@ func (s *Store) ReleaseByID(id string) error {
 		return nil
 	})
 	return err
+}
+
+func GetEscapedPath(dataDir string, fname string) string {
+	if runtime.GOOS == "windows" {
+		fname = strings.Replace(fname, ":", "_", -1)
+	}
+	return filepath.Join(dataDir, fname)
 }
--- a/plugins/ipam/host-local/backend/disk/lock.go
+++ b/plugins/ipam/host-local/backend/disk/lock.go
@@ -1,4 +1,3 @@
-// +build !windows
 // Copyright 2015 CNI authors
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
@@ -16,18 +15,28 @@
 package disk
 
 import (
+	"github.com/alexflint/go-filemutex"
 	"os"
-	"syscall"
+	"path"
 )
 
 // FileLock wraps os.File to be used as a lock using flock
 type FileLock struct {
-	f *os.File
+	f *filemutex.FileMutex
 }
 
 // NewFileLock opens file/dir at path and returns unlocked FileLock object
-func NewFileLock(path string) (*FileLock, error) {
-	f, err := os.Open(path)
+func NewFileLock(lockPath string) (*FileLock, error) {
+	fi, err := os.Stat(lockPath)
+	if err != nil {
+		return nil, err
+	}
+
+	if fi.IsDir() {
+		lockPath = path.Join(lockPath, "lock")
+	}
+
+	f, err := filemutex.New(lockPath)
 	if err != nil {
 		return nil, err
 	}
@@ -35,17 +44,16 @@ func NewFileLock(path string) (*FileLock, error) {
 	return &FileLock{f}, nil
 }
 
-// Close closes underlying file
 func (l *FileLock) Close() error {
 	return l.f.Close()
 }
 
 // Lock acquires an exclusive lock
 func (l *FileLock) Lock() error {
-	return syscall.Flock(int(l.f.Fd()), syscall.LOCK_EX)
+	return l.f.Lock()
 }
 
 // Unlock releases the lock
 func (l *FileLock) Unlock() error {
-	return syscall.Flock(int(l.f.Fd()), syscall.LOCK_UN)
+	return l.f.Unlock()
 }
--- a/plugins/ipam/host-local/host_local_test.go
+++ b/plugins/ipam/host-local/host_local_test.go
@@ -28,6 +28,7 @@ import (
 	"github.com/containernetworking/cni/pkg/types/current"
 	"github.com/containernetworking/plugins/pkg/testutils"
 
+	"github.com/containernetworking/plugins/plugins/ipam/host-local/backend/disk"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 )
@@ -37,7 +38,7 @@ var _ = Describe("host-local Operations", func() {
 		const ifname string = "eth0"
 		const nspath string = "/some/where"
 
-		tmpDir, err := ioutil.TempDir("", "host_local_artifacts")
+		tmpDir, err := getTmpDir()
 		Expect(err).NotTo(HaveOccurred())
 		defer os.RemoveAll(tmpDir)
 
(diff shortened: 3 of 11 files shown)


plugins/ipam/host-local/backend/disk/backend.go
```diff
@@ -56,6 +57,10 @@ func New(network, dataDir string) (*Store, error) {
 
 func (s *Store) Reserve(id string, ip net.IP, rangeID string) (bool, error) {
 	fname := filepath.Join(s.dataDir, ip.String())
+	if runtime.GOOS == "windows" {
+		fname = strings.Replace(fname, ":", "_", -1)
```
squeed: Hah, good catch!

pkg/ip/route_unspecified.go
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// +build !linux
+// +build !linux,!windows
```
squeed: Is there supposed to be a Windows version of this file? Or am I just missing something?

pkg/ipam/ipam_windows.go
```diff
@@ -0,0 +1,38 @@
...
+
+package ipam
+
+import (
+	"github.com/containernetworking/cni/pkg/invoke"
+	"github.com/containernetworking/cni/pkg/types"
+	"github.com/containernetworking/cni/pkg/types/current"
+)
+
+func ExecAdd(plugin string, netconf []byte) (types.Result, error) {
```
squeed: It looks like these functions are duplicated - Can you just move ConfigureIface to a build-specific file?

plugins/ipam/host-local/backend/disk/lock.go
```diff
@@ -1,3 +1,5 @@
+// +build linux
```
squeed: Just do !windows here - this "should" work on all Unix-like systems.

pkg/ip/addr.go
```diff
@@ -1,3 +1,5 @@
+// +build linux
```
squeed: Out of curiosity, does Windows also need to worry about "tentative" addresses?

squeed:
Seems pretty reasonable to me. I added a first pass of changes - you're right, not too many changes needed.

When doing build-tagged files, it's considered good practice to only duplicate the minimum amount of os-dependent code, and put the rest in non-build-tagged files.

pkg/ipam/ipam_windows.go
rakelkar: Moved the execXX funcs to exec.go - let me know if thats not what you had in mind

plugins/ipam/host-local/backend/disk/lock.go
rakelkar: Done

pkg/ip/route_unspecified.go
rakelkar: Cant make a windows specific version of these methods since they reference the netlink package in their method signature "dev netlink.Link" ... Not sure why this file is present in the repo.

pkg/ip/addr.go
rakelkar: Not sure, asking around

rakelkar: So it seems Windows doesn't need to worry about it right now since HNS doesn't support IPv6 yet - so ja bitte for now I guess ;-)

rakelkar:
@squeed Have addressed your comments, could you take a look? The test failure on CI is unrelated to this change (possibly a flaky test?).

squeed:
Kicked that CI job and it's green now, dammit.

Seems reasonable to me. I want to think for one more day if we want to have a specific conditional build policy - the maintainer meeting is tomorrow.

pkg/ipam/ipam_windows.go
```diff
@@ -0,0 +1,28 @@
+// +build windows
```
rosenhouse: I don't believe this line is necessary, since the filename ends in `_windows.go`.

[[Reference](https://golang.org/pkg/go/build/#hdr-Build_Constraints)]

plugins/ipam/host-local/backend/disk/lock_windows.go
```diff
@@ -0,0 +1,61 @@
...
+	fi, err := os.Stat(lockPath)
+	if err != nil {
+		return nil, err
+	}
+
+	if fi.IsDir() {
+		lockPath = path.Join(lockPath, "lock")
+	}
+
+	f, err := filemutex.New(lockPath)
```
rosenhouse: this code seems identical to the POSIX version, except it makes use of this nice cross-platform adapter.  Could we make this file the one true `lock.go` and remove the POSIX-only implementation entirely?

plugins/ipam/host-local/backend/disk/lock_windows.go
```diff
@@ -0,0 +1,61 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package disk
+
+import (
+	"github.com/alexflint/go-filemutex"
```
rosenhouse: Is this package added to the `vendor` directory somewhere?  I don't see it.

rosenhouse: [meta comment: I'm surprised this passed Travis without doing that.    Maybe we should open a separate PR to enforce vendoring at CI time.]

rakelkar:
Travis doesn't build windows... I can add a build.ps1 and an appveyor yml
to kick it off? The ps1 would only include plugins that work for Windows -
so only host-local for now...

pkg/ipam/ipam_windows.go
rakelkar: removed

plugins/ipam/host-local/backend/disk/lock_windows.go
rakelkar: Vendored

plugins/ipam/host-local/backend/disk/lock_windows.go
rakelkar: Done

plugins/ipam/host-local/backend/disk/lock.go
```diff
@@ -15,36 +15,47 @@
...
 		return nil, err
 	}
 
 	return &FileLock{f}, nil
 }
 
-// Close closes underlying file
 func (l *FileLock) Close() error {
-	return l.f.Close()
+	return nil
```
rosenhouse: why not close the file?

plugins/ipam/host-local/backend/disk/lock.go
```diff
@@ -15,36 +15,47 @@
...
 }
 
 // Lock acquires an exclusive lock
 func (l *FileLock) Lock() error {
-	return syscall.Flock(int(l.f.Fd()), syscall.LOCK_EX)
+	l.f.Lock()
+	return nil
```
rosenhouse: I didn't realize before that this library `panic`s on errors rather than returning them. 
Effective Go says that ["real library functions should avoid panic"](https://golang.org/doc/effective_go.html#panic).  I wonder if that library deserves a PR...

plugins/ipam/host-local/backend/disk/lock.go
rosenhouse: oh, because the library doesn't expose a `Close()` method at all.  uh oh....

rosenhouse: Although apparently we're not using `Close()` anyhow.  When I delete the `Close()` method from the `backend.Store` interface, the program still compiles.  So I guess you could just do that.

rakelkar: Nice catch - will remove, though I must say with a touch of disquiet.

rakelkar: So it seems like main.go:51 creates a Store struct and not the interface and calls defer store.Close so close is called after all...

rakelkar: There is an open PR on the library to add a close method, will create a PR to update this plugin once that has been merged. alexflint/go-filemutex#2

rakelkar: Updated with shiny new close method!

plugins/ipam/host-local/host_local_test.go
```diff
@@ -37,34 +38,34 @@ var _ = Describe("host-local Operations", func() {
...
+					[{ "subnet": "2001:db8:1::0/64" }]
+				],
+				"routes": [
+					{"dst": "0.0.0.0/0"},
+					{"dst": "::/0"},
+					{"dst": "192.168.0.0/16", "gw": "1.1.1.1"},
+					{"dst": "2001:db8:2::0/64", "gw": "2001:db8:3::1"}
+				]
+			}
+		}`, tmpDir, tmpDir)
```
rakelkar: I ran gofmt - and it felt indents were in order!

plugins/ipam/host-local/host_local_test.go
```diff
@@ -517,6 +518,15 @@ var _ = Describe("host-local Operations", func() {
 	})
 })
 
+func getTmpDir() (string, error) {
+	tmpDir, err := ioutil.TempDir("", "host_local_artifacts")
+	if err == nil {
+		tmpDir = filepath.ToSlash(tmpDir)
```
rakelkar: so without this the un-escaped backslashes in the windows path cause JSON parsing to barf

plugins/ipam/host-local/backend/disk/backend.go
```diff
@@ -56,10 +56,7 @@ func New(network, dataDir string) (*Store, error) {
 }
 
 func (s *Store) Reserve(id string, ip net.IP, rangeID string) (bool, error) {
-	fname := filepath.Join(s.dataDir, ip.String())
-	if runtime.GOOS == "windows" {
-		fname = strings.Replace(fname, ":", "_", -1)
-	}
+	fname := GetEscapedPath(s.dataDir, ip.String())
```
rakelkar: Turns out I missed a bunch, also escaping the colon in the dataDir doesnt work so well when the path includes a drive letter - duh!

rakelkar:
@rosenhouse Have addressed all comments, got the shiny new changes from alexflint and made fixes to get tests to run on Windows. Let me know if you see something anything don't like. The current build failure is an unrelated flaky test.

rosenhouse:
@rakelkar thanks for all your hard work on this!  we're excited to merge this, just as soon as we have Continuous integration set up for Windows in this repo.  I've got PR #84 open, but its not yet green.  Once that's merged, then we can return to this PR.

rosenhouse:
@rakelkar now that #84 is merged, please rebase and squash this down to a single commit, e.g. something like https://github.com/rosenhouse/plugins/commit/799776fec6216a76310df96ec486b7ca3dd44cbf

rakelkar:
@rosenhouse rebased and squashed!

plugins/ipam/host-local/backend/disk/lock.go
```diff
@@ -16,36 +15,45 @@
...
-	f, err := os.Open(path)
+func NewFileLock(lockPath string) (*FileLock, error) {
+	fi, err := os.Stat(lockPath)
+	if err != nil {
+		return nil, err
+	}
+
+	if fi.IsDir() {
+		lockPath = path.Join(lockPath, "lock")
+	}
```
rosenhouse: is this line necessary for windows?  If so, maybe it deserves a unit test?

rakelkar: yes it is required - added unit tests for filepath and folderpath

rakelkar: LMK if you like what you see, will wait for LGTM to squash!

rakelkar:
squashed commits

plugins/ipam/host-local/backend/disk/lock.go
rakelkar: squashed commits

PR #78, opened 2017-10-11
Don't let DHCP IPAM plugin fail on missing lease

rmohr:
There are at least two reasons why a lease is not present:

 * The dhcp ipam daemon was restarted
 * ON add the IPAM plugin failed

Don't fail the IPAM invocation when the lease is not present, to allow
proper device cleanup on CNI delete invocations.


rmohr:
I also wonder if the macvlan plugin should remove an already created device if the IPAM invocation on ADD fails, like it is done in #32 for the bridge plugin.

squeed:
Looks fine to me. Super minor request: Can you change the commit description to be "Don't let DHCP delete fail on missing lease"? Thanks!

rmohr:
@squeed ammended. Thx.

Issue #68, opened 2017-08-30
portmap doesn't work with hairpin


squeed:
When a container tries to talk to itself (over the portmapped "host" ip + port), it doesn't work. It should.

squeed:
The most obvious solution is to add an additional MASQ rule that rewrites the source IP to the host itself. This works, but I'm not sure how to set up the rule structure efficiently. @freehan maybe you have some suggestions?

dcbw:
@squeed word on the street is you want to twiddle accept_local or route_localnet, not sure which.

squeed:
There's no way to do this without a host-MASQ, because the souce port of the responding packet (i.e. a SYN/ACK) will be wrong.

PR #81, opened 2017-10-19
portmap: support hairpin, improve performance

squeed:
This change improves the performance of the portmap plugin and fixes hairpin, when a container is mapped back to itself.

Performance is improved by using a multiport test to reduce rule traversal, and by using a masquerade mark.

Hairpin is fixed by enabling masquerading for hairpin traffic.

Fixes: #68

### DIFF
--- a/plugins/meta/portmap/README.md
+++ b/plugins/meta/portmap/README.md
@@ -8,6 +8,8 @@ You should use this plugin as part of a network configuration list. It accepts
 the following configuration options:
 
 * `snat` - boolean, default true. If true or omitted, set up the SNAT chains
+* `markMasqBit` - int, (0-31), default 13. The mark bit to use for masquerading (see section SNAT). Cannot be set when `externalSetMarkChain` is used.
+* `externalSetMarkChain` - string, default nil. If you already have a Masquerade mark chain (e.g. Kubernetes), specify it here. This will use that instead of creating a separate chain. When this is set, `markMasqBit` must be unspecified.
 * `conditionsV4`, `conditionsV6` - array of strings. A list of arbitrary `iptables` 
 matches to add to the per-container rule. This may be useful if you wish to 
 exclude specific IPs from port-mapping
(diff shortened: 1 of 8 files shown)


squeed:
Of course, the tests pass on my machine. Looking in to why.

squeed:
@containernetworking/cni-maintainers This is finally ready for review.

plugins/meta/portmap/README.md
```diff
@@ -8,6 +8,8 @@ You should use this plugin as part of a network configuration list. It accepts
 the following configuration options:
 
 * `snat` - boolean, default true. If true or omitted, set up the SNAT chains
+* `markMasqBit` - int, (0-31), default 13. The mark bit to use for masquerading (see section SNAT)
+* `externalSetMarkChain` - string, default nil. If you already have a Masquerade mark chain (e.g. Kubernetes), specify it here. This will use that instead of creating a separate chain. This cannot be used with `markMasqBit`
```
matthewdupre: I think the last sentence would be better if it said something like "If this is specified, markMasqBit is ignored".  Cannot be used isn't super clear - will the server catch fire?

plugins/meta/portmap/README.md
```diff
@@ -39,7 +41,9 @@ look like:
                 {
                         "type": "portmap",
                         "capabilities": {"portMappings": true},
-                        "snat": false,
+                        "snat": true,
+                        "markMasqBit": 4,
```
matthewdupre: Prefer to use the default value here to avoid polluting bit 4 with people using it accidentally.

plugins/meta/portmap/README.md
squeed: Right now the plugin errors out if both are set; thoughts?

matthewdupre: I'm a little uneasy about that - if someone wants to set the chain, they have to go and look that the bit is pointed and they can nil it - nothing about this description suggests it can also be nilled.  Maybe just an extra line that tells people to nil / omit it?

PR #84, opened 2017-11-01
Enable Windows CI (Appveyor)

rosenhouse:
~Work in progress.~ DONE!

[Windows test output here](https://ci.appveyor.com/project/rosenhouse/plugins)


rakelkar:
@rosenhouse I get an access denied when I click on the detail link above. Maybe need to open up perms on appveyor a bit so that folks can debug build failures?

rosenhouse:
Hmm, we seem to be in a chicken & egg scenario here.  Host local is breaking on windows here, and you have #77 to make it work on Windows, which I've said is blocked on this.

One option is to continue adding build-tags here.  And then later rebase #77 on top of master?

Thoughts?

rakelkar:
If you mean add +build Linux in 77 and then have me rebase once 77 is in,
sure I can do that.

rakelkar:
Not 77 I mean #84

rakelkar:
Argh, you know what I mean.. I mean yes I can rebase, makes sense.

rosenhouse:
ok @rakelkar, @containernetworking/cni-maintainers this is now green on:
- Travis (Linux)
- Appveyor (Windows), using a limited set of packages (see `plugins/linux_only.txt` in the diff)

Please take a look!

Plan for post-merge:
- (re-)enable the Appveyor integration on this repo, so that we can detect regression in Windows compatibility.
- rebase #77, so we can remove `ipam/host-local` from the `linux_only` list.

squeed:
LGTM.

plugins/ipam/host-local/backend/disk/lock.go
```diff
@@ -1,3 +1,4 @@
+// +build !windows
```
rakelkar: >+build !windows [](start = 3, length = 15)

what made you include this instead of renaming it to lock_linux.go like you did with the other files?

rosenhouse: because this way the `host-local` tests can run on my Mac :-)

rakelkar:
LGTM

Issue #82, opened 2017-10-21
Creating Vlan tagging for the physical interface in a pod


rkamudhan:
I like get the suggestion from CNI maintainers on the idea, modifying the vlan plugin to add vlan tagging if the `master` interface is in the container network namespace.

```
eth0     Link encap:Ethernet  HWaddr 0e:6b:9a:06:a5:a0
          inet addr:192.168.116.3  Bcast:0.0.0.0  Mask:255.255.252.0
          inet6 addr: fe80::c6b:9aff:fe06:a5a0/64 Scope:Link
          UP BROADCAST RUNNING MULTICAST  MTU:1450  Metric:1
          RX packets:96 errors:0 dropped:0 overruns:0 frame:0
          TX packets:9 errors:0 dropped:0 overruns:0 carrier:0
          collisions:0 txqueuelen:0
          RX bytes:10232 (10.2 KB)  TX bytes:690 (690.0 B)

lo        Link encap:Local Loopback
          inet addr:127.0.0.1  Mask:255.0.0.0
          inet6 addr: ::1/128 Scope:Host
          UP LOOPBACK RUNNING  MTU:65536  Metric:1
          RX packets:0 errors:0 dropped:0 overruns:0 frame:0
          TX packets:0 errors:0 dropped:0 overruns:0 carrier:0
          collisions:0 txqueuelen:1
          RX bytes:0 (0.0 B)  TX bytes:0 (0.0 B)

net0     Link encap:Ethernet  HWaddr d2:70:b0:fd:ce:5a
          inet addr:10.56.217.131  Bcast:0.0.0.0  Mask:255.255.255.0
          inet6 addr: fe80::d070:b0ff:fefd:ce5a/64 Scope:Link
          UP BROADCAST RUNNING MULTICAST  MTU:1500  Metric:1
          RX packets:11484 errors:0 dropped:0 overruns:0 frame:0
          TX packets:4 errors:0 dropped:0 overruns:0 carrier:0
          collisions:0 txqueuelen:1000
          RX bytes:3927528 (3.9 MB)  TX bytes:288 (288.0 B)

net0.33  Link encap:Ethernet  HWaddr d2:70:b0:fd:ce:5a
          inet addr:10.56.217.132  Bcast:0.0.0.0  Mask:255.255.255.0
          inet6 addr: fe80::d070:b0ff:fefd:ce5a/64 Scope:Link
          UP BROADCAST RUNNING MULTICAST  MTU:1500  Metric:1
[...]

PR #87, opened 2017-11-01
adding vlan tagging for cont ns interface

rkamudhan:
Support VLAN tagging for the interface inside the container network namespace
This PR is related to the issue #82

### DIFF
--- a/plugins/main/vlan/vlan.go
+++ b/plugins/main/vlan/vlan.go
@@ -19,6 +19,7 @@ import (
 	"errors"
 	"fmt"
 	"runtime"
+	"strconv"
 
 	"github.com/containernetworking/cni/pkg/skel"
 	"github.com/containernetworking/cni/pkg/types"
@@ -32,9 +33,10 @@ import (
 
 type NetConf struct {
 	types.NetConf
-	Master string `json:"master"`
-	VlanId int    `json:"vlanId"`
-	MTU    int    `json:"mtu,omitempty"`
+	Master            string `json:"master"`
+	MasterInContainer bool   `json:"masterInContainer"`
+	VlanId            int    `json:"vlanId"`
+	MTU               int    `json:"mtu,omitempty"`
 }
 
 func init() {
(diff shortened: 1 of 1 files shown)


rosenhouse:
Sorry about the appveyor CI failure, I configured that integration for this repo before it was ready.  Please ignore that failure.

squeed:
I'm not exactly sure what this does - can you please explain? Also, all PRs that add a config flag need to have a corresponding update in the README for that plugin.

rkamudhan:
@squeed  Current VLAN plugin creates VLAN tagging for the master interface in the host namespace. We have a use case in which application inside a pod needs to connect to the master interface, which is in the container network namespace and connect to the application instances on two separate VLAN. The `isContNs` config flag in the PR is used to identify the master interface in the container network namespace.

Will create PR for the VLAN plugin README file.

squeed:
Ah, I get it now. A few thoughts.
1) Could you change the flag to "masterInContainer" - it makes it clearer
2) Could you refactor the code a bit? Make a common function that creates the tagged vlan interface, then call it appropriately?

Something like
```
if masterInContainer {
    containerNS.Do({
        makeVlanInterface()
    })
} else {
    makeVlanInterface()
}
````

**Comment by rkamudhan** [comment]
@squeed Thanks for the feedback, I will do it.

File: plugins/main/vlan/vlan.go [hunk 1]
```diff
@@ -32,9 +33,10 @@ import (
 
 type NetConf struct {
 	types.NetConf
-	Master string `json:"master"`
-	VlanId int    `json:"vlanId"`
-	MTU    int    `json:"mtu,omitempty"`
+	Master            string `json:"master"`
+	MasterInContainer bool   `json:"masterInContainer"`
```
rkamudhan: Changed the parameter name to "masterInContainer"

plugins/main/vlan/vlan.go
```diff
@@ -60,10 +62,29 @@ func loadConf(bytes []byte) (*NetConf, string, error) {
 
 func createVlan(conf *NetConf, ifName string, netns ns.NetNS) (*current.Interface, error) {
 	vlan := &current.Interface{}
+	var curNetns ns.NetNS
+	var m netlink.Link
+	var err error
+
+	if conf.MasterInContainer != false {
```
rkamudhan: checking the "MasterInContainer" flag and applying container network namespace to the find the master and add the vlan link to it.

rkamudhan:
@squeed Added the changes to the code.

rkamudhan:
@squeed Please let me know, whether the PR is OK?

rkamudhan:
@squeed @dcbw Please let me know, Whether I want to make changes for the PR.

jellonek:
@squeed @dcbw @bboreham ping

rkamudhan:
@squeed  @dcbw @bboreham Could you please review this PR.

plugins/main/vlan/vlan.go
```diff
@@ -60,10 +62,29 @@ func loadConf(bytes []byte) (*NetConf, string, error) {
...
+		curNetns = netns
+	} else {
+		curNetns, err = ns.GetCurrentNS()
+		if err != nil {
+			return nil, fmt.Errorf("failed to get init netns: %v", err)
+		}
+	}
+
```
dcbw: I feel like the code could be tightened up here, but also note that the "RandomVethName()" function must be called in the namespace that veth will be created in, so it must also be under curNetns.Do().  So the first half of the function from LinkByName(conf.Master) until LinkAdd(v) should be within a curNetns.Do() block.

plugins/main/vlan/vlan.go
```diff
@@ -127,6 +159,10 @@ func cmdAdd(args *skel.CmdArgs) error {
 	}
 	defer netns.Close()
 
+	if n.MasterInContainer != false {
+		args.IfName = getIfName(n.Master, n.VlanId)
+	}
```
dcbw: We shouldn't be doing this here or in cmdDel().  CNI requires that the plugin use the name passed as CNI_IFNAME, and if that's eth0 then the vlan interface must be called eth0.  We can't override it from the plugin per the spec.

rkamudhan:
@dcbw I will go through the changes and get back to you.

jellonek:
I've contacted guys from Intel working on SR-IOV and looks like this is no longer needed by anyone.

PR #88, opened 2017-11-01
ipvlan: support enslaving an interface returned by ipam

paulnivin:
For IP allocation schemes that cannot be interface agnostic, master can be set to "ipam". In this configuration, the IPAM plugin is required to return a single interface name for the ipvlan plugin to enslave.

For additional background, we've developed an IPAM plugin that dynamically scales elastic network interfaces (ENI) in AWS as we allocate and deallocate IP addresses for containers. In the ENI design, IP allocation is dependent on the subnet assigned to each interface and IPs returned are not contiguous. As such, we've delegated the decision for which master interface to enslave to our IPAM plugin. Per the CNI specification:

> Note that unlike regular CNI plugins, IPAM plugins should return an abbreviated Result structure that does not include the interfaces key, since IPAM plugins should be unaware of interfaces configured by their parent plugin except those specifically required for IPAM (eg, like the dhcp IPAM plugin).

In this case, the interface is specifically required for IPAM and our IPAM plugin returns a Result structure that contains a single "interfaces" key containing the name of the master interface for the ipvlan plugin to enslave. This approach is intended to be simple and straightforward without violating the CNI spec.


squeed:
Interesting. This is more straightforward than I expected - nice job.

There is an ordering question raised here. We generally create the interface *before* executing IPAM, in case the IPAM plugin wishes to use the interface or its properties (e.g. mac) when making the allocation. The typical example for this is the DHCP plugin, which needs to use the interface to do IPAM.

However, this is not a hard rule in the spec. Given that an ipvlan interface is completely useless without an IP (and the MAC is the same as the "master"), it is probably okay to do the reordering. Thoughts?

paulnivin:
I agree that reordering should be fine. If you'd prefer, I can update the PR so we only reorder if master is set to `ipam` and the existing behavior is retained if master specifies an interface. Let me know.

rosenhouse:
Hi @paulnivin We talked about this one for a while at the maintainers meeting today.  There are a few things that weigh against merging it:

- It appears to violate the "should"s in the CNI spec
- It is designed to support an IPAM plugin that we don't have in the open source
- It is for a pretty narrow use case

Any thoughts on these issues?

paulnivin:
Thanks for following up. To address each point:

> It appears to violate the "should"s in the CNI spec

Per RFC2119:

> SHOULD   This word, or the adjective "RECOMMENDED", mean that there may exist valid reasons in particular circumstances to ignore a particular item, but the full implications must be understood and carefully weighed before choosing a different course.

In this case, we'd like for the IPAM plugin to communicate an interface decision to the CNI plugin as the interface cannot be reasonably determined in advance. There appears to be no other way in the CNI spec to have arbitrary data (ex. "args") that can be passed via the Result structure from an IPAM plugin to the calling CNI plugin. We could alternatively modify the pre-CNI calling portion of our Kubernetes stack to call to a shim to decide on an interface, dynamically create a configuration for the ipvlan plugin, and then make the call out to CNI to use the pre-determined/allocated interface and IPs. Ignoring the "should" request in the CNI spec with respect to returning an "interfaces" key greatly simplifies any approach, as the interface is the only additional piece of data that our IPAM plugin needs to return to to the CNI plugin.

> It is designed to support an IPAM plugin that we don't have in the open source

We're in the process of open sourcing the IPAM plugin. This is the one upstream change that's needed to make use of our IPAM plugin, so we put in the pull request early with the hope that it would land by the time we've completed the open source process.

> It is for a pretty narrow use case

We'd argue it's not a narrow use case as CNI is heavily tied to Kubernetes and there are significant downsides to currently recommended and widely-used networking configurations for Kubernetes in AWS. We've developed our ENI/IPvlan-based IPAM plugin to support fast networking, be conformant with all Kubernetes networking requirements for pod/cluster communication, and be easy to debug if issues arise. This is all done via a set of CNI plugins with no additional software or runtimes.

We're open to any option the CNI maintainers feel is the best path for moving forward.

paulnivin:
We've open sourced our CNI plugins. The [announcement](https://eng.lyft.com/announcing-cni-ipvlan-vpc-k8s-ipvlan-overlay-free-kubernetes-networking-in-aws-95191201476e) includes an overview on why we developed this implementation and the source code is at https://github.com/lyft/cni-ipvlan-vpc-k8s.

We'd welcome input on how to proceed. If our approach sounds sane, we could also propose a modification to the CNI spec where IPAM plugins may return an `interfaces` key if an IPAM plugin needs to be aware of interfaces.

bboreham:
I would like to suggest using the chaining mechanism rather than special-casing the IPAM call on a magic interface name.

You would define a chain which first calls your ENI/IPAM plugin then chains to the ipvlan plugin.  This PR would then change to examine `prevResult` and, if it contains a suitable interface and IP address, use that.

This feels cleaner to me because it is using chaining as designed rather than bending the ipam mechanism.

paulnivin:
I've updated the PR to use chaining instead of the special-cased IPAM call. For testing, I've been using cni-ipvlan-vpc-k8s and this setup is working well. Let me know your thoughts. Thanks.

paulnivin:
@bboreham @rosenhouse @squeed Any thoughts on the updated PR now that we're using chaining? Is this suitable for merging or should additional changes be made?

squeed:
@paulnivin it's awesome, thanks!

Issue #94, opened 2017-11-16
Traffic shaping plugin


squeed:
A motivating example is the kubenet deprecation, but that certainly isn't the only use. It would be nice to have a chained plugin that allowed for some basic bandwidth limiting / traffic shaping.

This would be a new capability arg.

squeed:
Oh, that's awesome. The other implementation I was looking at was the one in k8s, but I'm not particularly wedded to it. In fact, it doesn't support bursting, and since it doesn't use IFB, I suspect it doesn't rate same-host bandwidth (which might be OK...).

squeed:
Also, I think @dnardo said he had a shaping plugin almost done?

dnardo:
I do have something that has parity with the existing stuff in kubenet.  I was waiting for the GET call to be available since I need that for proper cleanup.  I'll clean it up and send it out for review.

PR #96, opened 2017-11-27
plugins/meta/bandwidth: traffic shaping plugin

DennisDenuto:
As mentioned in https://github.com/containernetworking/plugins/issues/94

An *attempt* to port `github.com/cloudfoundry/cf-networking-release` traffic control tbf (ingress/egress) logic into a chained plugin


build.sh
```diff
@@ -19,7 +19,7 @@ export GOPATH=${PWD}/gopath
 mkdir -p "${PWD}/bin"
 
 echo "Building plugins"
-PLUGINS="plugins/meta/* plugins/main/* plugins/ipam/* plugins/sample"
+PLUGINS="plugins/meta/* plugins/main/* plugins/ipam/* plugins/sample plugins/tc/*"
```
rosenhouse: cool.   you'll need to add it to [`plugins/linux_only.txt`](https://github.com/containernetworking/plugins/blob/master/plugins/linux_only.txt) also to fix the AppVeyor (Windows CI) build.

plugins/tc/tbf/main.go
```diff
@@ -0,0 +1,235 @@
...
+			if *cip.Interface == pidx {
+				continue
+			}
+			nonContainerNetNs, err := ns.GetNS(prevIface.Sandbox)
+			if err != nil {
+				return err
+			}
+			err = nonContainerNetNs.Do(func(netNS ns.NetNS) error {
+				nonContainerIfaceName := getRoutableHostIF(cip.Address.IP)
+				if nonContainerIfaceName != "" {
```
rosenhouse: I *think* that you can assume that this plugin is being called from the "host" namespace.  In other words, the current NS when the plugin starts is the place to put host devices.  That should remove the need for these nested loop and conditionals.

plugins/tc/tbf/main.go
```diff
@@ -0,0 +1,235 @@
...
+
+// cmdAdd is called for ADD requests
+func cmdAdd(args *skel.CmdArgs) error {
+	conf, err := parseConfig(args.StdinData)
+	if err != nil {
+		return err
+	}
+
+	if conf.Rate == 0 && conf.Burst == 0 {
+		return types.PrintResult(conf.PrevResult, conf.CNIVersion)
```
rosenhouse: maybe add a comment here, e.g. `no traffic shaping was requested, so just no-op and quit`

plugins/tc/tbf/main.go
```diff
@@ -0,0 +1,235 @@
...
+				continue
+			}
+			nonContainerNetNs, err := ns.GetNS(prevIface.Sandbox)
+			if err != nil {
+				return err
+			}
+			err = nonContainerNetNs.Do(func(netNS ns.NetNS) error {
+				nonContainerIfaceName := getRoutableHostIF(cip.Address.IP)
+				if nonContainerIfaceName != "" {
+					ifbDeviceName := os.Getenv("CNI_IFNAME")
```
rosenhouse: Ah, also, this is not correct.  According to [the CNI spec](https://github.com/containernetworking/cni/blob/869505bebdf96fc1176bc29ad0893b867e717f6a/SPEC.md), `CNI_IFNAME` is the desired name for the primary network interface *inside the container*, i.e. `eth0`.  It is guaranteed unique only inside the container network namespace, not the host.  You'll need to generate a unique name for the host-side interface -- perhaps by hashing the `(network name, container id)` pair (see [here for some considerations](https://github.com/containernetworking/cni/blob/869505bebdf96fc1176bc29ad0893b867e717f6a/SPEC.md#general-considerations)), using a random name (see [here](https://github.com/containernetworking/plugins/blob/92c634042c3888d942199c396c1d6f42d82785f3/pkg/ip/link_linux.go#L87-L96)), or by using an auto-incrementing integer on the system.  You may need to store this name on the local filesystem in order to properly clean it up when `DEL` is called.

plugins/tc/tbf/main.go
```diff
@@ -0,0 +1,235 @@
...
+
+// cmdDel is called for DELETE requests
+func cmdDel(args *skel.CmdArgs) error {
+	conf, err := parseConfig(args.StdinData)
+	if err != nil {
+		return err
+	}
+	_ = conf
+
+	for _, prevIface := range conf.PrevResult.Interfaces {
```
rosenhouse: Ah, I'm afraid `PrevResult` will not be populated on a `DEL` call.  The [runtime will execute the plugins in reverse order](https://github.com/containernetworking/cni/blob/869505bebdf9/libcni/api.go#L147).  (also, to top it off, most plugins don't return any data on a successful `DEL`).

So to know the interface name to teardown, you may need some other tricks.  One way is that `ADD` may store some state (i.e. the interface name) on the filesystem, and `DEL` reads it back out.  See the other comment about the IFB device name.

plugins/tc/tbf/main.go
squeed: You're right - plugins are always called from the host.

plugins/tc/tbf/main.go
DennisDenuto: I like the idea of generating a deterministic unique name for the host side interface. I have pushed a change to compute the sha1 of the (network name, container id) and only using the first 15 characters of the hex digest.
Thus on cmdDel i am able to determine the device name to delete.

On the off chance that the sha1 conflicts with an interface that already exists, this plugin will fail... If this is an issue then I can change to the random name approach + stored state. (i.e. if the random name conflicts, generate a new random name until no conflict)

plugins/tc/tbf/main.go
DennisDenuto: Dealt with this with generating a deterministic unique name (sha1 of network name and container id).

DennisDenuto:
I like the idea of adding a measure test and asserting ingress/egress shaping. I can chain it with some plugins too. I can work on that next.

plugins/tc/tbf/main.go
```diff
@@ -0,0 +1,208 @@
...
+}
+
+func getIfbDeviceName(networkName string, containerId string) (string, error) {
+	hash := sha1.New()
+	_, err := hash.Write([]byte(networkName + containerId))
+	if err != nil {
+		return "", err
+	}
+
+	return fmt.Sprintf("%x", hash.Sum(nil))[:15], nil
```
rosenhouse: I think this is ok, but I'd like other @containernetworking/cni-maintainers to weigh in.

plugins/tc/tbf/main.go
```diff
@@ -0,0 +1,208 @@
...
+		// for an unknown interface (which we'll assume to be Container-side
+		// Skip all IPs we know belong to an interface with the wrong name.
+		if intIdx >= 0 && intIdx < len(conf.PrevResult.Interfaces) && conf.PrevResult.Interfaces[intIdx].Name != args.IfName {
+			continue
+		}
+		containerIPConfigs = append(containerIPConfigs, ip)
+	}
+
+	if len(containerIPConfigs) == 0 {
+		return fmt.Errorf("got no container IPs")
```
rosenhouse: Why bother with this check?  I don't see `containerIPConfigs` used later in this function.

plugins/tc/tbf/main.go
```diff
@@ -0,0 +1,208 @@
...
+	}
+
+	for _, prevIface := range conf.PrevResult.Interfaces {
+		if prevIface.Sandbox != "" {
+			continue
```
rosenhouse: This check makes sense.  I wonder about readability of this code though.
One option is to extract a private function with a signature like 
```go
func getHostInterface(ifaces []*current.Interface)` (*current.Interface, error) { ... }
```
You could keep the `Sandbox` emptiness check in there, and do stuff like have it error if no matching interfaces are found. 

That would make it more clear what this code is doing.  It would also have the side benefit of removing the for-loop indentation from the main code-path here, so your top-level function would read
```go
hostInterfaceface, err := getHostInterface(conf.PrevResult.Interfaces)
if err != nil {
   return err
}
err = CreateIngressQdisc(conf.Rate, conf.Burst, hostInterface.Name)
...
```

plugins/tc/tbf/main.go
```diff
@@ -0,0 +1,208 @@
...
+	return types.PrintResult(conf.PrevResult, conf.CNIVersion)
+}
+
+// cmdDel is called for DELETE requests
+func cmdDel(args *skel.CmdArgs) error {
+	conf, err := parseConfig(args.StdinData)
+	if err != nil {
+		return err
+	}
+	_ = conf
```
rosenhouse: ?

plugins/tc/tbf/tbf_linux_test.go
```diff
@@ -0,0 +1,471 @@
...
+				runtimeWithoutLimit = b.Time("without tbf", func() {
+					result, err := current.GetResult(containerWithoutTbfRes)
+					Expect(err).NotTo(HaveOccurred())
+
+					addr := result.IPs[0].Address.IP.String() + ":8080"
+					makeTcpClientInNS(hostNs, addr, packetInBytes)
+				})
+			})
+
+			Expect(runtimeWithLimit).To(BeNumerically(">", runtimeWithoutLimit+1000*time.Millisecond))
```
rosenhouse: Awesome.  I love this.

plugins/tc/tbf/tbf_linux_test.go
```diff
@@ -0,0 +1,471 @@
...
+			{
+				"version": "4",
+				"address": "%s/24",
+				"gateway": "10.0.0.1",
+				"interface": 1
+			}
+		],
+		"routes": []
+	}
+}`
```
rosenhouse: Interesting.  I see you've got a more integration-type test with the PTP plugin below.  Is there a reason you want to test with a literal result JSON as well?  (Not opposed, just curious.)

plugins/tc/tbf/tbf_suite_test.go
```diff
@@ -0,0 +1,193 @@
...
+		}
+
+		reply := make([]byte, 1)
+		_, err = conn.Read(reply)
+
+		return err
+	})).To(Succeed())
+}
+
+func startTcpServerInNS(containerNS ns.NetNS, laddr string) {
```
rosenhouse: At a glance, this sounds a lot like the [`echosvr` we use in the `portmap` integration tests](https://github.com/containernetworking/plugins/blob/92c634042c3888d942199c396c1d6f42d82785f3/plugins/meta/portmap/portmap_suite_test.go#L66-L78).

Perhaps there's an opportunity for code-reuse?

plugins/tc/tbf/main.go
matthewdupre: I think I'd prefer a short prefix (<= 4 characters).  That should still leave enough entropy (>=2^44) even just using the basic hex encoding.

plugins/tc/tbf/tbf_suite_test.go
DennisDenuto: Ah, thanks for pointing this out. I like the idea of code-reuse. 

I am finding using echosvr a bit tricky to use for this particular test. Mainly because the echosvr [tries to read a fixed amount of bytes](https://github.com/containernetworking/plugins/blob/92c634042c3888d942199c396c1d6f42d82785f3/pkg/testutils/echosvr/main.go#L34) and then continues to write back what it read and close the connection. However in the measure case I want to ensure that the entire message is sent and read by the server.  I did this previously by sending a message with a trailing new-line character as a signal to the server that the entire message had been sent, and that it could then write back a response so the client could know that the server received the entire message.

I have tried to get this to work by tweaking the num of packets sent to the 'echo server' and tbf's burst/limit sent to fit into the buffer size of 512. However this results in the time taken with/without tbf to be negligible (~100 millisecond difference) 

I have tried to increase the `echosvr` buffer size  to allow sending/reading a larger payload, but `conn.Read` seems like it reads as much as the `read syscall` is set to (which when I run is consistently 61232 bytes). If i send < 61232 to the `echosvr` it results in a `broken pipe error`. Haven't looked too much into it but I suspect that since ingress and egress traffic are both shaped, that when the `echo svr`  tries to write back a large payload it gets shaped, and it closes the connection prematurely.

Your thoughts are much appreciated.

squeed: I'd be happy with replacing the echoserver with this code. It really should be in another process, if only because Go namespace crappiness has caused us a lot of grief and flaky tests.

DennisDenuto: ok thanks for the reply. will do

DennisDenuto:
Thanks for the comments.
fyi:
- renamed to `plugins/meta/bandwidth`
- use `echosvr`
- ifb device name is 4 characters long
- added a `readme.md`

plugins/tc/tbf/main.go
DennisDenuto: copy-pasta from the plugin template example :-(

plugins/meta/bandwidth/README.md
```diff
@@ -0,0 +1,35 @@
...
+The following network configuration file
+```
+{
+  "name": "slowdown",
+  "type": "bandwidth",
+  "rate": 123,
+  "burst": 456
+}
+```
+
```
rosenhouse: It might be useful to have a more fleshed out example of chaining, e.g. with bridge first and then this.

plugins/meta/bandwidth/README.md
```diff
@@ -0,0 +1,35 @@
+# bandwidth plugin
+
+## Overview
+
+This plugin provides a way to use and configure Linux's Traffic control (tc) subystem. tc encompasses the sets of mechanisms and operations by which packets are queued for transmission/reception on a network interface.
+
+This plugin configures a token bucket filter (tbf) queing discipline (qdisc) on both ingress and egress traffic. Resulting in traffic being shaped when reading / writing.
+
+Due to limitations on tc shaping rules for ingress, this plugin creates an Intermediate Functional Block device (ifb) to redirect packets from the target interface. tc tbf is then applied to the ifb device. The packets that were redirected to the ifb devices, are written OUT (and shaped) to the target interface.
```
rosenhouse: You say "target" interface but it may be more clear to distinguish between host-side and container-side interfaces here.

DennisDenuto:
fyi
- Added a chaining example in the readme
- renamed 'target' -> 'host interface'
- Updated the commit message

plugins/meta/bandwidth/README.md
```diff
@@ -0,0 +1,112 @@
+# bandwidth plugin
+
+## Overview
+
+This plugin provides a way to use and configure Linux's Traffic control (tc) subystem. tc encompasses the sets of mechanisms and operations by which packets are queued for transmission/reception on a network interface.
+
+This plugin configures a token bucket filter (tbf) queing discipline (qdisc) on both ingress and egress traffic. Resulting in traffic being shaped when reading / writing.
```
matthewdupre: queing -> queuing

DennisDenuto: good catch. fixed typo.

plugins/meta/bandwidth/README.md
```diff
@@ -0,0 +1,112 @@
...
+
+The following is an example [json configuration list](https://github.com/containernetworking/cni/blob/master/SPEC.md#network-configuration-list-runtime-examples) for creating a `ptp` between the host -> container via veth interfaces, whereby traffic is shaped by the `bandwidth` plugin:
+
+```json
+{
+  "cniVersion": "0.3.1",
+  "name": "mynet",
+  "plugins": [
+    {
+      "name": "my-net",
```
squeed: The "name" field here (and below) isn't needed.

plugins/meta/bandwidth/README.md
```diff
@@ -0,0 +1,112 @@
...
+      "type": "bandwidth",
+      "rate": 123,
+      "burst": 456
+    }
+  ]
+}
+```
+
+
+1. The following is the json payload the container runtime would need to perform during the ADD action:
```
squeed: You probably don't need to explain plugin chaining in this README.

plugins/meta/bandwidth/main.go
```diff
@@ -0,0 +1,199 @@
...
+	// parse this in two passes. If your plugin is not chained, this can be
+	// removed (though you may wish to error if a non-chainable plugin is
+	// chained.
+	// If you need to modify the result before returning it, you will need
+	// to actually convert it to a concrete versioned struct.
+	RawPrevResult *map[string]interface{} `json:"prevResult"`
+	PrevResult    *current.Result         `json:"-"`
+
+	// Add plugin-specifc flags here
+	Rate  int `json:"rate"`  //Bandwidth rate in Kbps for traffic through container. 0 for no limit. If rate is set, burst must also be set
```
squeed: Can you add separate ingress and egress rates?

plugins/meta/bandwidth/main.go
```diff
@@ -0,0 +1,199 @@
...
+		if err != nil {
+			return nil, fmt.Errorf("could not convert result to current version: %v", err)
+		}
+	}
+	// End previous result parsing
+
+	switch {
+	case conf.Burst < 0 || conf.Rate < 0:
+		return nil, fmt.Errorf("rate and burst must be a positive integer")
+	case conf.Burst == 0 && conf.Rate != 0:
```
squeed: I would rather infer when `burst == 0` that `burst = rate`. But this is subjective.

plugins/meta/bandwidth/main.go
```diff
@@ -0,0 +1,199 @@
...
+	}
+
+	ifbDevice, err := netlink.LinkByName(ifbDeviceName)
+	if err != nil {
+		return err
+	}
+
+	conf.PrevResult.Interfaces = append(conf.PrevResult.Interfaces, &current.Interface{
+		Name:    ifbDeviceName,
+		Sandbox: hostInterface.Sandbox,
```
squeed: Sandbox should not be set, since the IFB device isn't in the sandbox / container netns.

plugins/meta/bandwidth/ifb_creator.go
```diff
@@ -0,0 +1,165 @@
+package bandwidth
+
+// Copyright 2015 CNI authors
```
squeed: 2018 :-)

DennisDenuto: happy new year :-)

plugins/meta/bandwidth/main.go
DennisDenuto: done. pushed.

plugins/meta/bandwidth/README.md
DennisDenuto: removed plugin chaining examples. pushed.

DennisDenuto:
fyi @matthewdupre @squeed the requested changes have been pushed.

plugins/meta/bandwidth/main.go
```diff
@@ -0,0 +1,218 @@
...
+	types.NetConf
+	RuntimeConfig *struct{} `json:"runtimeConfig"`
+
+	// This is the previous result, when called in the context of a chained
+	// plugin. Because this plugin supports multiple versions, we'll have to
+	// parse this in two passes. If your plugin is not chained, this can be
+	// removed (though you may wish to error if a non-chainable plugin is
+	// chained.
+	// If you need to modify the result before returning it, you will need
+	// to actually convert it to a concrete versioned struct.
```
bboreham: This comment should be removed - it is advice to the plugin author.
It could be replaced with a comment about how you resolved the decisions.

plugins/meta/bandwidth/main.go
```diff
@@ -0,0 +1,218 @@
...
+	RawPrevResult *map[string]interface{} `json:"prevResult"`
+	PrevResult    *current.Result         `json:"-"`
+
+	// Add plugin-specifc flags here
```
bboreham: This comment should be removed - it is advice to the plugin author.

plugins/meta/bandwidth/main.go
```diff
@@ -0,0 +1,218 @@
...
+	return link.Attrs().MTU, nil
+}
+
+func getHostInterface(interfaces []*current.Interface) (*current.Interface, error) {
+	for _, prevIface := range interfaces {
+		if prevIface.Sandbox != "" {
+			continue
+		}
+
+		return prevIface, nil
```
rosenhouse: So in the maintainers meeting someone pointed out that this will break for the bridge plugin, since bridge returns two non-sandbox interfaces: the bridge itself, and the container veth, in that order.

We're considering how best to resolve the issue: perhaps by adding a `peerIndex` field to the result structure so that the prior plugin can specify which interface is the container's peer veth.

Alternately, you may be able to detect whether a particular interface is a peer of the sandbox one, cc @dcbw and @bboreham for possible tricks here.

dcbw: @rosenhouse implementing the "get peer index" library code for #75 brought up the following question: what do we consider a "peer"?  Of course we know what that should be for veth devices.  But what other device types does this plugin support?

eg, does it support ipvlan or macvlan?  In that case, there is only a "parent" device that may have multiple "child" ipvlan/macvlan devices attached.  The child devices all have the IFLA_LINK attribute (which is what veth sets to indicate its peer) set to the parent device that they all share.

We can obviously define a `peerIndex` field however we like in the CNI spec, but it would be a shame if that was veth-only or relied on some implied behavior of the LInux kernel.

rosenhouse: Ok, lets do the hack to detect the peer on the host side, and then update the readme to call out that this only works with plugins that put a peer veth on the host.

plugins/meta/bandwidth/bandwidth_linux_test.go
```diff
@@ -0,0 +1,565 @@
...
+	"github.com/containernetworking/cni/pkg/skel"
+	"github.com/containernetworking/cni/pkg/types"
+	"github.com/containernetworking/cni/pkg/types/current"
+	"github.com/containernetworking/plugins/pkg/ns"
+	"github.com/containernetworking/plugins/pkg/testutils"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	"github.com/onsi/gomega/gexec"
+	"github.com/vishvananda/netlink"
+	"net"
```
m1093782566: nits:

move "net" and "time" right below "fmt"?

m1093782566:
Hi all,

Any chance to push it forward? I have been thinking about supporting traffic shaping in kubelet/cni and this work relies on it.

rosenhouse:
Yeah, since #75 is still not merged, I would suggest we copy over the peer-discovery function from there into here.  I can probably get to that this week if @DennisDenuto doesn't beat me too it.

DennisDenuto:
apologies for the delay, i have been a bit preoccupied.  @rosenhouse If you have the peer-discovery function, please feel free to add it. I won't be able to look into it this week. I would probably have time on the weekend.

rosenhouse:
I've pushed a couple commits here to address the nits, and opened #120 so we can use that library function from this PR.

dcbw:
#120 has been merged so this can get rebased and updated to use GetVethPeerIfindex().

m1093782566:
Friendly ping @DennisDenuto ~

rosenhouse:
Dennis, feel free to pick this up if you want, but if not, I can pair with @chinanglea and team on it next week.

Lion-Wei:
Looking forward for this to be merged. :+1:

DennisDenuto:
@rosenhouse yeah that would be great if you could pair on it with the c2c team. Thanks

caseydavenport:
Coming to this a bit late so apologies if it's already been discussed, but have folks considered using the [capabilities mechanism + convention](https://github.com/containernetworking/cni/blob/master/CONVENTIONS.md#dynamic-plugin-specific-fields-capabilities--runtime-configuration) to allow runtime configuration of the plugin, similar to how the portmap plugin works today?

squeed:
@DennisDenuto / @rosenhouse is this ready to go?

DennisDenuto:
@m1093782566 Cool. I personally think submitting that as a separate PR makes sense.

plugins/meta/bandwidth/main.go
```diff
@@ -112,15 +113,16 @@ func getMTU(deviceName string) (int, error) {
...
+			if err == nil {
+				return iface, err
+			}
 		}
-
-		return prevIface, nil
 	}
-
-	return nil, errors.New("no host interface found")
+	return nil, errors.New("no host interface found: " + err.Error())
```
rosenhouse: if the loop never runs, then `err` won't be set, right?

integration/integration_linux_test.go
```diff
@@ -20,64 +20,181 @@ import (
...
+
+			By(fmt.Sprintf("adding %s to %s\n\n", "chained-bridge-bandwidth", contNS1.ShortName()), func() {
+				chainedBridgeBandwidthEnv.runInNS(hostNS, cnitoolBin, "add", "network-chain-test", contNS1.LongName())
+				chainedBridgeIP = ipRegexp.FindString(chainedBridgeBandwidthEnv.runInNS(contNS1, "ip", "addr"))
+				Expect(chainedBridgeIP).To(ContainSubstring("10.11.2."))
+			})
 
-	It("supports add and del with bridge + bandwidth", func() {
-		basicAssertion("chained-bridge-bandwidth", "10.11.2.")
+			By(fmt.Sprintf("adding %s to %s\n\n", "basic-bridge", contNS2.ShortName()), func() {
```
rosenhouse: nit: I slightly prefer the `By(...)` syntax where there isn't another nested `func()`, e.g. https://onsi.github.io/ginkgo/#documenting-complex-its-by

This makes it so you don't have to do weird things like lift variables outside.

But no big deal.

tylerschultz: We removed the nesting and localized some vars.

plugins/meta/bandwidth/main.go
tylerschultz: Good catch, that would have nil pointed. Fixed.

rosenhouse:
@squeed I think this is ready to go.  WDYT?
any way to validate that this meets the K8s-community's needs for kubelet?

m1093782566:
It looks very good now and can meet the basic needs of kubelet.

plugins/meta/bandwidth/README.md
```diff
@@ -0,0 +1,64 @@
...
+      "ipam": {
+          "type": "host-local",
+          "subnet": "10.0.0.0/24"
+      },
+      "dns": {
+        "nameservers": [ "10.1.0.1" ]
+      }
+    },
+    {
+      "name": "slowdown",
```
squeed: minor nit: this `name` isn't needed.

PR #97, opened 2017-11-27
delete link and ip if err when cmdAdd to avoid resource leak.

oilbeater:
We encounter link and ip leak in production, one reason on cni side is that it not recycle resource when err happens. 

We first meet the issue in macvlan plugin, however this issue appears in other plugins as well, if this modification is reasonable I will apply the change to other plugins.

### DIFF
--- a/plugins/main/macvlan/macvlan.go
+++ b/plugins/main/macvlan/macvlan.go
@@ -161,11 +161,28 @@ func cmdAdd(args *skel.CmdArgs) error {
 		return err
 	}
 
+	// Delete link if err to avoid link leak in this ns
+	defer func() {
+		if err != nil {
+			netns.Do(func(_ ns.NetNS) error {
+				return ip.DelLinkByName(args.IfName)
+			})
+		}
+	}()
+
 	// run the IPAM plugin and get back the config to apply
 	r, err := ipam.ExecAdd(n.IPAM.Type, args.StdinData)
 	if err != nil {
 		return err
 	}
+
+	// Invoke ipam del if err to avoid ip leak
+	defer func() {
+		if err != nil {
+			ipam.ExecDel(n.IPAM.Type, args.StdinData)
+		}
+	}()
+
 	// Convert whatever the IPAM result was into the current Result type
 	result, err := current.NewResultFromResult(r)
 	if err != nil {


rosenhouse:
I've seen others add this sort of code to plugins as well.

The CNI spec says "[plugin DEL should be idempotent](https://github.com/containernetworking/cni/blob/master/SPEC.md#general-considerations)".  The intention was that **runtimes** could run `DEL` after any failed ADD.  That way plugins would not need to contain this sort of self-cleanup logic.  It seems like your runtime (kubernetes? mesos? something else?) doesn't do this.

I'm wondering if the right solution is maybe one of the following:
1. fix your runtime and *also* add a line to the spec recommending that runtimes do this
2. make this sorts of change to all our reference plugins and *also* add a line to the spec to explicitly recommend this behavior from other plugins

maybe other @containernetworking/cni-maintainers have thoughts too...

oilbeater:
We encounter this problem in Kubernetes. As our investigation many parts will lead resource leak (kubelet failed to call cni delete, docker leak netns, network partition and other inconsistence between other components ...) We have an above-all gc script that monitors all stats and deal with the inconsistence.

From my point, every components should do their best to provide atomic operations and keep a consistent state, as other parts the components rely on for gc may fail. If a create operation failed in the middle, it should not make the system to an unknown state, which also lead create not an idempotent operation. 

I'd like to change other reference plugins to recycle resources if failed in middle if this is solution is preferred.

oilbeater:
any update on this? @rosenhouse

bboreham:
On principle, choice 2 seems better.  Although it is more code for us to add.

squeed:
We briefly discussed this in the maintainer's meeting. My position:
- Plugins that "wrap" other plugins (most notably: ipam) should try to call DEL on failure during ADD
- Runtimes should call DEL if ADD fails.

This is a change to libcni and the spec.

squeed:
Also, we can probably get this in to spec version v0.4.0

rosenhouse:
This gets complicated when you have a chain of plugins.  Is it the responsibility of the library to call DEL immediately on ADD failure?  I'm a bit worried about hiding errors from the runtime.  Runtimes should have the opportunity to inspect and log each error.

@bboreham had some more thoughts on this in the call today.

Issue #95, opened 2017-11-20
Host-local: add a capability arg for desired address range


squeed:
The host-local (and maybe the DHCP) plugin should support range configuration via capability args.

Questions:
1) Do we support the full pool datastructure? Or just a CIDR?
2) What do we do if there is both a range specified in configuration *and* a capability arg?

karampok:
Hi, following our discussion to slack, what I was thinking that would be useful to my case
is to be able to provide a such chain list
```
{
  "name": "aname",
  "plugins": [
     {
         "type": "host-local",
         "config":{
            "subnet": "10.10.20.0/24",
            "routes": [
                { "dst": "0.0.0.0/0" }
            ],
            "dataDir": "/run/ipam-out-net"
         },
         "capabilities": {"overWriteConfig": true}
     },
     {
        "type": "bridge",
        "bridge": "cnt-bridge",
        "isGateway": false,
        "ipMasq": false,
        "hairpinMode": true,
        "dns": { "nameservers": [ "8.8.8.8" ] }
    },
    {
      "type": "portmap",
      "capabilities": {"portMappings": true},
      "snat": false
      }
  ]
}
```

the runtime would be able (if allowed by the operator) to modify the config of the host-local.

What do you think?

If I run like that (given host-local works as chained plugin), do I expect bridge to work as it is now?

squeed:
Hey there, @karampok.
I don't think we'll change the IPAM plugin to work as a chained plugin, but that shouldn't matter. We should be able to support what you want without any troubles.

squeed:
My thoughts to my own two questions:

1) We should support the full IPAM structure. If it's needed statically, it will be needed dynamically
2) If the administrator configures both the capability args *and* a static config, do both. Capability IPs should come first.

karampok:
My suggestion for the IPAM to work also as chained is because I think it will make the implementation in the runtime simpler,  but I can see also many reason why not.

One other case that should be thought is that after we dynamically change the subnet, we might need to dynamically change the bridge name.

PR #100, opened 2017-12-11
ipam/host-local: Accept ip ranges as a runtime argument

squeed:
This allows for the runtime to dynamically request IP ranges.

Fixes: #95


dcbw:
LGTM

PR #103, opened 2017-12-26
pkg/ipam: Skip ip.SettleAddresses if only IPv4 is used

alice02:
This change improves the performance of the ipam.ConfigureIface.
Some plugins are slow because of the ip.SettleAddress in ipam.ConfigureIface.
It seems to be only needed for IPv6, so should be skipped if only IPv4 is used.

### DIFF
--- a/pkg/ipam/ipam_linux.go
+++ b/pkg/ipam/ipam_linux.go
@@ -65,7 +65,9 @@ func ConfigureIface(ifName string, res *current.Result) error {
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


dcbw:
Yeah, it's true that the conditions the code currently checks for (eg IFA_F_TENTATIVE) are only used for IPv6 addresses.  So LGTM.  @squeed ?

squeed:
I dream of a day of IPv4 link-local autoconfiguration. Oh, wait, I don't.  LGTM.

Issue #51, opened 2017-08-02
DHCP: correctly handle route fields


squeed:
From the RFC:
> If the DHCP server returns both a Classless Static Routes option and
   a Router option, the DHCP client MUST ignore the Router option.

>   Similarly, if the DHCP server returns both a Classless Static Routes
   option and a Static Routes option, the DHCP client MUST ignore the
   Static Routes option.

We do neither: we ignore the Router option and merge the Classless and Static sections.

mzahorik:
I ran into this bug.  I've submitted https://github.com/containernetworking/plugins/pull/104 to correct.  I've spun up some nginx containers under Kubernetes with "classic" static routes (DHCP option 33) and all looks good including the containers getting a default route with the router as gateway.

This is the first time I've touched Go, and my Infobloxes don't allow me to specify classless static routes, so please check that everything looks right for adding the routes returned by parseCIDRRoutes()

PR #104, opened 2018-01-03
Append default route and process route options in compliance with RFC 3442

mzahorik:
Append a default route to the CNI reply if there's a gateway advertised.

Classless static routes (DHCP option 121) are now processed first.
If CSRs exist, static routes (DHCP option 33) and the gateway default
route are ignored as per RFC 3442.

Fixes: #51

### DIFF
--- a/plugins/ipam/dhcp/lease.go
+++ b/plugins/ipam/dhcp/lease.go
@@ -292,8 +292,26 @@ func (l *DHCPLease) Gateway() net.IP {
 }
 
 func (l *DHCPLease) Routes() []*types.Route {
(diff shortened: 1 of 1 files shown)


s1061123:
@mzahorik +1
I also met the issue. Currently dnsmasq DHCP server just sends Router option, hence current dhcp ipam cannot put default route into container (i.e. Pod in k8s)...

squeed:
Looks good to me. Can you squash your two commits, please?

mzahorik:
*squish* Done.  All tests pass.  Enjoy!

PR #107, opened 2018-01-11
pkg/ip: don't write to /proc/sys if ipforward enabled

zhsj:
This enables setup in a container env like systemd nspawn
where /proc/sys is mouted as read only.

Signed-off-by: Shengjing Zhu <i@zhsj.me>

### DIFF
--- a/pkg/ip/ipforward_linux.go
+++ b/pkg/ip/ipforward_linux.go
@@ -15,6 +15,7 @@
 package ip
 
 import (
+	"bytes"
 	"io/ioutil"
 
 	"github.com/containernetworking/cni/pkg/types/current"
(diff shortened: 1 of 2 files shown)


rosenhouse:
Thanks!  Is there any way you could add test coverage for this change?

zhsj:
Hi, I'd love to write tests, but could you give me some hints. Is there a way to emulate a procfs file and ensure it's not been written?

rosenhouse:
I was thinking just a unit test of `echo1`.  The test could run `echo1(someTempFile)` and verify that the file is written to only if it doesn't already contain `1`.    Could verify this by checking last modified time, or filemask, or something.

What do you think?

zhsj:
ok, I add a test based on the ctime. I previously thought to test `echo1` with a file `0400` mode, but I find these tests will be run as root. So I think only ctime can tell.

dcbw:
LGTM

PR #111, opened 2018-01-23
Add -hostprefix in DHCP daemon to run the daemon as container

s1061123:
This diff adds -hostprefix option in dhcp daemon. This option could be used to run dhcp daemon as container because container cannot touch host's procfs directly. 

The diff changes dhcp daemon to touch procfs mounted to another path, like '/hostfs/proc'.

### DIFF
--- a/plugins/ipam/dhcp/daemon.go
+++ b/plugins/ipam/dhcp/daemon.go
@@ -39,8 +39,9 @@ const resendCount = 3
 var errNoMoreTries = errors.New("no more tries")
 
 type DHCP struct {
-	mux    sync.Mutex
-	leases map[string]*DHCPLease
+	mux             sync.Mutex
+	leases          map[string]*DHCPLease
+	hostNetnsPrefix string
 }
 
 func newDHCP() *DHCP {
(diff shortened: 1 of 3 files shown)


squeed:
This should probablly be called `hostprefix`, for a few reasons:

- The netns path may not be in `/proc`, it could also be in `/run/netns`
- A containerized daemon will also need to write the socket file on the host.

But it's a good idea. @s1061123, can you make the changes? Thanks!

s1061123:
@squeed Thank you for your comments!
Only `ip netns` command creates netns path in "/run/netns". Docker and k8s does not create it, hence
we need to touch procfs netns, that is automatically create by kernel (like /proc/<pid>/ns/net).
Actually k8s invokes CNI plugin with '/proc/<pid>/ns/net'.

But I agree with you that from CNI point of view, this prefix name should be changed to "hostprefix".

I will update diffs tomorrow!

plugins/ipam/dhcp/daemon.go
```diff
@@ -37,6 +37,7 @@ const listenFdsStart = 3
 const resendCount = 3
 
 var errNoMoreTries = errors.New("no more tries")
+var hostNetnsPrefix = ""
```
bboreham: This should be a member of the `DHCP` struct, not a package variable

s1061123: Currently one DHCP daemon serves one DHCP structure hence I don't think so, but I am okey to do that because in the future DHCP daemon may serves multiple DHCP structure.

I will update the diff.

PR #112, opened 2018-01-23
pkg/ns: enable non-linux builds

vbatts:
Signed-off-by: Vincent Batts <vbatts@hashbangbash.com>

### DIFF
--- a/pkg/ns/ns_linux.go
+++ b/pkg/ns/ns_linux.go
@@ -1,3 +1,5 @@
+// +build linux
+
 // Copyright 2015-2017 CNI authors
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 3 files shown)


pkg/ns/ns_linux.go
```diff
@@ -1,3 +1,5 @@
+// +build linux
```
rosenhouse: Since this file already ends with `_linux.go`, adding the build constraint comment is superfluous.  See [docs here](https://golang.org/pkg/go/build/#hdr-Build_Constraints), specifically:
> If a file's name, after stripping the extension and a possible _test suffix, matches any of the following patterns:
> ```
> *_GOOS
> *_GOARCH
> *_GOOS_GOARCH
> ```
> (example: source_windows_amd64.go) where GOOS and GOARCH represent any known operating system and architecture values respectively, then the file is considered to have an implicit build constraint requiring those terms (in addition to any explicit constraints in the file).

rosenhouse:
@vbatts thanks for the PR.  What is your use case for these changes?

The reason I ask is that we used to have this kind of thing in one form or another (interface in a platform-agnostic file, stub implementations in the non-linux file).  But we removed as we were adding Windows support.

Put another way: what kind of automated test could you add that would prevent us from making this regression, again?

pkg/ns/ns_linux.go
vbatts: true. I was just being explicit. I can remove the build tag.

vbatts:
@rosenhouse good information. I didn't have that history. Just finding that libraries that even import this library on non-linux platforms, even if only for referencing the `NetNS` interface/struct fail to find source present in this package.
It is completely possible to work around this in the packages doing the import.

vbatts:
@rosenhouse that is to say, I can close this issue, just figured that others have hit this as well

rosenhouse:
sounds good.  If you're just using the interface type elsewhere, I'd highly recommend you define your own interface where it is being used, and only expose the methods that you need.

PR #113, opened 2018-02-01
Explicitly enable IPv6 sysctl

nyren:
If CNI is about to configure an IPv6 address, make sure IPv6 is not disabled through the "disable_ipv6" sysctl setting.

This is a workaround for Docker 17.x which sets disable_ipv6=1 for all interfaces even if ipv6 is enabled in Docker.

Please see containernetworking/cni#531 for a more detail discussion of the issue.

### DIFF
--- a/pkg/ipam/ipam_linux.go
+++ b/pkg/ipam/ipam_linux.go
@@ -21,8 +21,13 @@ import (
 
 	"github.com/containernetworking/cni/pkg/types/current"
 	"github.com/containernetworking/plugins/pkg/ip"
+	"github.com/containernetworking/plugins/pkg/utils/sysctl"
 
 	"github.com/vishvananda/netlink"
+)
+
+const (
+	DisableIPv6SysctlTemplate = "net.ipv6.conf.%s.disable_ipv6"
 )
 
 // ConfigureIface takes the result of IPAM plugin and
@@ -42,6 +47,7 @@ func ConfigureIface(ifName string, res *current.Result) error {
 	}
 
 	var v4gw, v6gw net.IP
+	var has_enabled_ipv6 bool = false
 	for _, ipc := range res.IPs {
 		if ipc.Interface == nil {
 			continue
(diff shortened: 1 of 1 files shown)


pkg/ipam/ipam_linux.go
```diff
@@ -52,6 +58,21 @@ func ConfigureIface(ifName string, res *current.Result) error {
 			return fmt.Errorf("failed to add IP addr %v to %q: invalid interface index", ipc, ifName)
 		}
 
+		// Make sure sysctl "disable_ipv6" is 0 if we are about to add
+		// an IPv6 address to the interface
+		if !has_enabled_ipv6 && ipc.Version == "6" {
+			// Enable IPv6 for the interface being configured as well as
+			// loopback "lo", "all" and "default"
```
dcbw: Any particular reason to do this for lo, all, and default?  It may not be wanted/needed on other interfaces created by other CNI chains executing for the same container.

nyren: You are right, for `all` and `default` I do not have any particularly strong reason. My reasoning was that it would be better to have the default to be "IPv6-enabled" but this is not something CNI need to enforce.

Regarding the loopback interface `lo` I could not get the kube-dns pod to run without IPv6 enabled on `lo`. If I understood correctly, kube-dns uses 3 containers sharing the same network NS, and the containers use the loopback device for internal communication. When running kube-dns with IPv6 it expects `::1` (IPv6 on loopback) to be available.

I think it is fairly reasonable to assume that applications running IPv6 would expect IPv6 to be available on loopback.

I have updated the pull-request to only enable IPv6 for `lo` and the interface being configured by CNI. Thanks for the feedback!

squeed:
Man, it's frustrating that we have to fix this in CNI... Bah. Thanks for taking care of this.

Anyways, could you add some code to only touch the sysctl if needed? There are some cases when the sysctls are read-only, but it's OK because we don't need to change anything.

squeed:
Looking at our `sysctl` library, I seen now that it's really weird. That varadic function doesn't really make sense. I think it needs a rewrite. I can take care of this, if you like.

nyren:
Yes, totally agree, really frustrating.

The code should only hit the sysctl if an IPv6 address is about to be configured, no unit tests unfortunately but I tested it manually. The implementation also make sure to only write the sysctl once per interface.

Regarding read-only /proc/sys the code should then just silently fail (I intentionally do not check the return value of the sysctl function) with the reasoning that we will fail later anyway. A way to log a warning would be nice though.

I am quite the n00b when it comes to go programming so if there is anything that needs to be fixed in the sysctl library I would gladly hand it over to you :)

The patch seems to work at least, I have a 20+ nodes cluster which happily runs IPv6-only kubernetes with CNI bridge on Docker 17.12 now.

nyren:
Ah, you mean reading back the sysctl first and only writing if necessary? Sure, I can fix that. Do you think it is safe to error out if the write fails then?

nyren:
Pull request updated to only write the `disable_ipv6` sysctl if necessary. The implementation will now return an error if it is unable to write the sysctl. Hope it does not break things.

I had a look at the Sysctl() implementation and the variadic looks OK to me. If you want to read a value you run `value, err := sysctl.Sysctl("net.ipv6.conf.whatever")` and to write a new value `_, err := sysctl.Sysctl("net.ipv6.conf.whatever", "new_value")`. Seems to work that way at least :)

Btw, there is no logging framework available in CNI at the moment, right?

abhijitherekar:
Hey Nyren,

I am still not following even if the sysctl is present how can you write to the /proc/sys file-system of a POD as its only read-only.??

nyren:
@abhijitherekar let us continue the discussion on why things works as they do in containernetworking/cni#531. Best to keep the pull request comments focused on the implementation details.

However, if you did test the patch and it failed, then of course let me know here.

pkg/ipam/ipam_linux.go
```diff
@@ -52,6 +58,30 @@ func ConfigureIface(ifName string, res *current.Result) error {
...
+		if !has_enabled_ipv6 && ipc.Version == "6" {
+			// Enabled IPv6 for loopback "lo" and the interface
+			// being configured
+			for _, iface := range [2]string{"lo", ifName} {
+				ipv6SysctlValueName := fmt.Sprintf(DisableIPv6SysctlTemplate, iface)
+
+				// Read current sysctl value
+				value, err := sysctl.Sysctl(ipv6SysctlValueName)
+				if err != nil || value == "0" {
+					// FIXME: log warning if unable to read sysctl value
```
squeed: Indeed, we don't currently have a logging strategy. It's not a big deal right now; if this fails, then the address will either apply or fail.

squeed:
Need a second maintainer to approve.

squeed:
Gonna try and pull this in to the v0.7.0 release.

PR #117, opened 2018-02-07
dhcp plugin testcases and bug fixes

dcbw:
@containernetworking/cni-maintainers @squeed


squeed:
Dan. Daaaaaaaan.

In all seriousness, Travis comes with some firewall rules - and is very slow, so race conditions love to happen. Make sure there aren't any iptables rules getting you down.

dcbw:
> Dan. Daaaaaaaan.
> 
> In all seriousness, Travis comes with some firewall rules - and is very slow, so race conditions love to happen. Make sure there aren't any iptables rules getting you down.

@squeed hey, it wasn't that, bear with me here when I cannot reproduce things locally.  The problem turned out to be Ginkgo Expect() in a goroutine not being trapped properly, which blocked the AfterEach() from cleaning up properly, combined with the testcase not waiting until the DHCP server was actually shut down due to the above.

But that's all fixed now, so please review.

squeed:
> The problem turned out to be Ginkgo Expect() in a goroutine not being trapped properly despite a GinkgoRecover(), which blocked the AfterEach() from cleaning up properly, combined with the testcase not waiting until the DHCP server was actually shut down due to the above.

Ouch. Nice find.

LGTM.

PR #118, opened 2018-02-07
packaging/containernetworking-cni.spec - COPR

baude:
I would like to start building RPMs in my COPY repo
for CNI as part of "Upstream_CRIO_Family". We would build
new packages on each merge based on a webhook between
COPR and github.

https://copr.fedorainfracloud.org/coprs/baude/Upstream_CRIO_Family/

This does not replace the Fedora RPMs but does allow us to work
with the most recent CNI code in an easier fashion.

Signed-off-by: baude <bbaude@redhat.com>

### DIFF
--- a/.copr/Makefile
+++ b/.copr/Makefile
@@ -1,0 +1,28 @@
+#!/usr/bin/make -f
+
+spec := .copr/containernetworking-cni.spec
+outdir := $(CURDIR)
+tmpdir := build
+gitdir := $(PWD)/.git
+
+rev := $(shell sed 's/\(.......\).*/\1/' $(gitdir)/$$(sed -n '/^ref:/{s/.* //;p}' $(gitdir)/HEAD))
+date := $(shell date +%Y%m%d.%H%M)
+
+version := $(shell sed -n '/Version:/{s/.* //;p}' $(spec))
+release := $(date).git.$(rev)
+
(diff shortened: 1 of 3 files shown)


squeed:
Seems fine. Would you mind adding some documentation - either in the packaging directory or in the RELEASING document - about how to build the RPMs? I _always_ forget how to do it.

baude:
sure, i also need to add a little more to this ... hang tight, will bump it when ready.

baude:
@squeed should be good to go now.

dcbw:
LGTM

baude:
nudge

baude:
ping

rosenhouse:
I'm ok with this, but I'm curious: is it common to do this for open source projects -- mixing the RPM packaging into the upstream source code, rather than maintaining it in a separate repo?  Is there a reason it needs to live here?

dcbw:
@rosenhouse it's common to have a "deb" directory in projects to make .debs for Debian-based distros, but less so with "cloudy"-type projects I think since they don't get packaged as much as in the old days.  It would help us to create RPMs as an upstream project if we wanted to do so.

That said I'm happy to take responsibility for the specfile.

baude:
Working around this by not storing files with upstream.

PR #120, opened 2018-02-17
Adds GetVethPeerIfindex library function

rosenhouse:
Just cherry-picking commits from #75 so that we can use them in #96 also.

cc @dcbw @squeed


dcbw:
lgtm

PR #126, opened 2018-02-27
Don't umount with MNT_DETACH

mrunalp:
MNT_DETACH could take time and the subsequent removal
of the file could fail with mount busy.

@dcbw @rajatchopra PTAL

Signed-off-by: Mrunal Patel <mrunalp@gmail.com>

### DIFF
--- a/pkg/ns/ns_linux.go
+++ b/pkg/ns/ns_linux.go
@@ -125,7 +125,7 @@ func (ns *netNS) Close() error {
 	ns.closed = true
 
 	if ns.mounted {
-		if err := unix.Unmount(ns.file.Name(), unix.MNT_DETACH); err != nil {
+		if err := unix.Unmount(ns.file.Name(), 0); err != nil {
 			return fmt.Errorf("Failed to unmount namespace %s: %v", ns.file.Name(), err)
 		}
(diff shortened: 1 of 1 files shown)


squeed:
Are you seeing any issues with this code? Or is this more of a theoretical issue?

runcom:
we're are actually seeing containers hanging forever trying to umount/rm the netns, probably because we wait for umount to be successful (and we enter in a loop where we cannot recover)

dcbw:
@mrunalp in the maintainers meeting we decided that CNI's pkg/ns should probably get out of the business of mount points entirely.  There are a number of cases where external projects want to handle the mount point thing differently and moving it out of pkg/ns would let CRIO and any other projects that use pkg/ns can get the mount point behavior they want.  

Instead, we'd like to move the mounting specific stuff from pkg/ns to pkg/testutils and only use it for our testcases, while leaving the rest of the logic in pkg/ns.

Does that sound like something you'd be willing to take a first pass at?

squeed:
In #131 I've removed namespace creation entirely :-). In testutils there is a correct implementation of namespace creation that should be copied to your own project.

squeed:
Going to close this; we've decided to get out of the namespace management business.

PR #129, opened 2018-03-10
bridge: release IP in case of error

mauriciovasquezbernal:
In case of error ip addresses are not being released and hence there is a possible leakage.

### DIFF
--- a/plugins/main/bridge/bridge.go
+++ b/plugins/main/bridge/bridge.go
@@ -37,6 +37,9 @@ import (
 	"github.com/vishvananda/netlink"
 )
 
+// For testcases to force an error after IPAM has been performed
+var debugPostIPAMError error
+
 const defaultBrName = "cni0"
 
 type NetConf struct {
@@ -465,6 +468,11 @@ func cmdAdd(args *skel.CmdArgs) error {
 	brInterface.Mac = br.Attrs().HardwareAddr.String()
 
 	result.DNS = n.DNS
+
+	// Return an error requested by testcases, if any
+	if debugPostIPAMError != nil {
+		return debugPostIPAMError
+	}
 
 	success = true
 
--- a/plugins/main/bridge/bridge.go
+++ b/plugins/main/bridge/bridge.go
@@ -19,6 +19,7 @@ import (
 	"errors"
 	"fmt"
 	"net"
+	"os"
 	"runtime"
 	"syscall"
 
@@ -323,6 +324,8 @@ func enableIPForward(family int) error {
 }
 
 func cmdAdd(args *skel.CmdArgs) error {
+	var success bool = false
+
 	n, cniVersion, err := loadNetConf(args.StdinData)
 	if err != nil {
 		return err
@@ -357,6 +360,15 @@ func cmdAdd(args *skel.CmdArgs) error {
 	if err != nil {
 		return err
 	}
+
+	// release IP in case of failure
+	defer func() {
+		if !success {
+			os.Setenv("CNI_COMMAND", "DEL")
+			ipam.ExecDel(n.IPAM.Type, args.StdinData)
+			os.Setenv("CNI_COMMAND", "ADD")
+		}
+	}()
 
 	// Convert whatever the IPAM result was into the current Result type
 	result, err := current.NewResultFromResult(r)
@@ -453,6 +465,8 @@ func cmdAdd(args *skel.CmdArgs) error {
 	brInterface.Mac = br.Attrs().HardwareAddr.String()
 
 	result.DNS = n.DNS
+
+	success = true
 
 	return types.PrintResult(result, cniVersion)
 }
--- a/plugins/main/bridge/bridge_test.go
+++ b/plugins/main/bridge/bridge_test.go
@@ -221,6 +221,24 @@ func ipVersion(ip net.IP) string {
 	return "6"
 }
 
+func countIPAMIPs(path string) (int, error) {
+	count := 0
+	files, err := ioutil.ReadDir(path)
+	if err != nil {
+		return -1, err
+	}
+	for _, file := range files {
+		if file.IsDir() {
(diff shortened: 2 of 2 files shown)


dcbw:
I'm not so hot on the environment munging, but there isn't a great way to do it that preserves the inputs that IPAM ADD got, without a lot more code in the bridge plugin...  @squeed what do you think?

Otherwise looks OK to me.

rosenhouse:
Yeah.  Also, can we get some test coverage here?

dcbw:
@mauriciovasquezbernal any chance you could add a few testcases too?  Thanks!

mauriciovasquezbernal:
@dcbw sure, Will do next week when back to the office.

mauriciovasquezbernal:
I am trying to create a test case but I am not sure how to do it. 
A first idea is to simulate an error after [1], then access to the host-local IPAM folder structure to determine if the IP was released. However I don't know how to simulate such error.

Does anyone have a better idea of how to test it?

[1] https://github.com/mauriciovasquezbernal/plugins/blob/391eccf0245b0df4e6a22f1521ff12c661d2d85f/plugins/main/bridge/bridge.go#L374

dcbw:
@mauriciovasquezbernal I took a look at this and yeah, it's not as simple as I thought.  Here's what I'd do...

1) create a function in bridge.go like:
```
// For testcases to force an error after IPAM has been performed
var debugPostIPAMError error

func cmdAdd(args *skel.CmdArgs) error {
    ...
	}

        // Return an error requested by testcases, if any
        if debugPostIPAMError != nil {
            return debugPostIPAMError
        }

	if n.IsGW {
    ...
```

2) then in your testcase code, you would do:

```
                debugPostIPAMError = fmt.Errorf(<your error message here>)
		r, raw, err := testutils.CmdAddWithResult(tester.targetNS.Path(), IFNAME, tester.args.StdinData, func() error {
			return cmdAdd(tester.args)
		})
		Expect(err).To(MatchError(<your error message here>)
                <ensure the IPAM allocation has been released>
```

For ensuring the IPAM has been released, I guess you would ensure that the host-local IPAM datadir does not contain any files that look like IP addresses.  If you get stuck on this part, don't hesitate to ask.

dcbw:
@mauriciovasquezbernal any chance you could update based on comment?  Thanks!

jellonek:
@mauriciovasquezbernal ping

mauriciovasquezbernal:
@dcbw @jellonek A big sorry from my side! I've forgotten about it and I was not receiving the notifications. 

I pushed a new commit with the test, in order to check if the IPAM module released the IP I count the files inside the IPAM datadir. Do you think this is good enough or a more sophisticated mechanism should be implemented?
I also had to define as a constant the default datadir of the `host-local` IPAM plugin as this is not exported. Any ideas to improve it?

Thanks.

plugins/main/bridge/bridge_test.go
```diff
@@ -906,4 +909,36 @@ var _ = Describe("bridge Operations", func() {
...
+			Expect(err).NotTo(HaveOccurred())
+
+			args := tc.createCmdArgs(originalNS)
+
+			// get number of allocated IPs before asking for a new one
+			files, _ := ioutil.ReadDir(ipamDatadir)
+			before := len(files)
+
+			debugPostIPAMError = fmt.Errorf("debugPostIPAMError")
+			_, _, err = testutils.CmdAddWithResult(originalNS.Path(), IFNAME, args.StdinData, func() error {
```
jellonek: `CmdAddWithResult` was renamed to `CmdAdd` in 731298003c55418f15a73044757a5e36985fae0f. It now has a slightly different signature.

mauriciovasquezbernal: I rebased to master and changed this function by `CmdAddWithArguments`
Thanks!

mauriciovasquezbernal:
travis tests are failing randomly, I think parallel tests cause issues because the way check if the IP have been released (it is possible that other tests allocate and release IPs at the same time).

Other failure is `/var/lib/cni/networks/testConfig/` not found. I have no idea why.

Should we disable parallel tests?, Is there a way to disable parallel testing only for the bridge?

dcbw:
@mauriciovasquezbernal retriggered the tests, seem to pass now.

plugins/main/bridge/bridge_test.go
```diff
@@ -106,6 +107,8 @@ const (
 
 	ipamEndStr = `
     }`
+
+	ipamDatadir = "/var/lib/cni/networks/testConfig/"
```
dcbw: One thing you could do here avoid race issues is to create your own datadir for the test.  I've done an implementation of that in https://github.com/dcbw/plugins/commit/2ee31427f9156b57bfb6ec754fed419099eb9e9f .  Cherry-pick that commit from my fork into your PR and then pass `dataDir` to tc.createCmdArgs() and then *also* to countIPAMIPs().  That'll make sure nothing else messes with the datadir during the test.

mauriciovasquezbernal: Thanks @dcbw for the suggestion, this solves the issue.
I updated the commits with this.

plugins/main/bridge/bridge.go
```diff
@@ -358,6 +364,15 @@ func cmdAdd(args *skel.CmdArgs) error {
 		return err
 	}
 
+	// release IP in case of failure
+	defer func(ok *bool) {
```
jellonek: Small question - why it's passed there as a pointer value? Inside of this func it's already accessible as `success` and it will have a correct value in time of deferred execution, so there is no need for this `ok` variable as `success` can be used directly.

mauriciovasquezbernal: You're right. Fixed.

dcbw:
LGTM now, thanks!

PR #132, opened 2018-03-22
allow building with custom Go binary

lsm5:
I need to build deb packages using golang-1.10 on Ubuntu Xenial which is
present in a non-default location while /usr/bin/go is at golang-1.6.

With this commit I can simply run:
`GO=/usr/lib/go-1.10/bin/go ./build.sh`
to use golang-1.10.

Signed-off-by: Lokesh Mandvekar <lsm5@fedoraproject.org>

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -15,6 +15,7 @@ fi
 
 export GO15VENDOREXPERIMENT=1
 export GOPATH=${PWD}/gopath
+export GO="${GO:-go}"
 
 mkdir -p "${PWD}/bin"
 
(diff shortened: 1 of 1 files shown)


build.sh
```diff
@@ -16,6 +16,10 @@ fi
 export GO15VENDOREXPERIMENT=1
 export GOPATH=${PWD}/gopath
 
+if [ -z ${GO+x} ]; then
```
squeed: Is there any reason not to just do `GO="${GO:-go}"` ?

lsm5: sure, i'll give it a try, this was just a dumb noobie way to get it going I guess :\

lsm5:
@squeed ping, PTAL at change.

squeed:
LGTM. Just needs a second maintainer's OK.

dcbw:
LGTM

PR #136, opened 2018-03-29
Add static CNI plugin

s1061123:
This changes to add 'static' CNI plugin, which provides to assign IPv4/v6 address statically from given config file. See README.md for the details.

### DIFF
--- a/plugins/ipam/static/README.md
+++ b/plugins/ipam/static/README.md
@@ -1,0 +1,45 @@
+# static IP address management plugin
+
+## Overview
+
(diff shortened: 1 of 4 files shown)


squeed:
Might there be any reason for this to have locking? In other words, only allow the IP to be issued to one running container. I suppose if you're using this plugin, you are in debug mode. But it may be a good idea to add.

dcbw:
@squeed I'd say locking could be added later, perhaps?  Also, because it's static, even locking wouldn't prevent assigning the same IP to two containers on different nodes.  The locking would only help for containers on the same node.

dcbw:
@squeed said in the maintainers meeting he had no problem with the locking issue.

PR #137, opened 2018-04-03
Get rid of gateway IP assignment to veth in host namespace for PTP plugin

jingax10:
In this PR, we get rid of gateway IP assignment to veth in host namespace for PTP plugin by enabling Proxy ARP and Proxy NDP for IPv4 and IPv6, respectively. Specifically,

- Enable Proxy ARP and Proxy NDP if applicable for veth in host namespace;
- Refactor the ptp tests with two network namespaces pinging one to the other;
- Minor comments and logging changes;

### DIFF
--- a/plugins/main/ptp/ptp.go
+++ b/plugins/main/ptp/ptp.go
@@ -30,14 +30,15 @@ import (
 	"github.com/containernetworking/plugins/pkg/ipam"
 	"github.com/containernetworking/plugins/pkg/ns"
 	"github.com/containernetworking/plugins/pkg/utils"
+	"github.com/containernetworking/plugins/pkg/utils/sysctl"
 	"github.com/j-keck/arping"
 	"github.com/vishvananda/netlink"
 )
 
 func init() {
-	// this ensures that main runs only on main thread (thread group leader).
-	// since namespace ops (unshare, setns) are done for a single thread, we
-	// must ensure that the goroutine does not jump from OS thread to thread
+	// This ensures that main runs only on main thread (thread group leader).
+	// Since namespace ops (unshare, setns) are done for a single thread, we
+	// must ensure that the goroutine does not jump from OS thread to thread.
 	runtime.LockOSThread()
 }
 
@@ -74,7 +75,7 @@ func setupContainerVeth(netns ns.NetNS, ifName string, mtu int, pr *current.Resu
 		containerInterface.Sandbox = netns.Path()
 
 		for _, ipc := range pr.IPs {
-			// All addresses apply to the container veth interface
+			// All addresses apply to the container veth interface.
 			ipc.Interface = current.Int(1)
 		}
 
@@ -90,7 +91,7 @@ func setupContainerVeth(netns ns.NetNS, ifName string, mtu int, pr *current.Resu
 		}
 
 		for _, ipc := range pr.IPs {
-			// Delete the route that was automatically added
+			// Delete the route that was automatically added.
 			route := netlink.Route{
 				LinkIndex: contVeth.Index,
 				Dst: &net.IPNet{
@@ -136,7 +137,7 @@ func setupContainerVeth(netns ns.NetNS, ifName string, mtu int, pr *current.Resu
 			}
 		}
 
-		// Send a gratuitous arp for all v4 addresses
+		// Send a gratuitous arp for all v4 addresses.
 		for _, ipc := range pr.IPs {
 			if ipc.Version == "4" {
 				_ = arping.GratuitousArpOverIface(ipc.Address.IP, *contVeth)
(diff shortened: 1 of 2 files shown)


squeed:
Just making sure I understand this: When this is enabled, the gateway address is expected to be external (i.e. a real router)? Thus making the container "seem" directly connected to the network? If so, I like it.

plugins/main/ptp/ptp.go
```diff
@@ -136,13 +137,6 @@ func setupContainerVeth(netns ns.NetNS, ifName string, mtu int, pr *current.Resu
 			}
 		}
 
-		// Send a gratuitous arp for all v4 addresses
```
squeed: We needed this for the case when an IP is reused very quickly, so the arp tables are stale. I'm not sure we can get rid of it.

jingax10:
> Just making sure I understand this: When this is enabled, the gateway address is expected to be external (i.e. a real router)? Thus making the container "seem" directly connected to the network? If so, I like it.

Yes. Based on the route configuration in container namespace, the imaginary gateway (with .1 address) is the direct connected next-hop for container to reach external world. Therefore, we can expect to have one and only one ARP/Neigh entry on behalf of gateway address in container namespace. On the other hand, proxy ARP & NDP are techniques to resolve MAC address on behalf of such a gateway by not requiring veth in host namespace to own the gateway IP. Given IP forwarding has been enabled in host namespace (node is configured as a router), packets egressed from container namespace would be routed once enters into host namespace. Moreover, without assigning gateway address to all veth devs in host namespace, we can reduce gateway route entries per veth in local route table.

plugins/main/ptp/ptp.go
jingax10: Based on the route configuration in container namespace, we only need to resolve MAC address for gateway (with .1 address). I'm not sure why sending gratuitous ARP on behalf of container IP could help when IP is reused.

I actually ran the test with and without gratuitous ARP and both passed. Please let me know you want me to revert the change on gratuitous ARP. Thanks!

squeed: The gratuitous arp is to prevent a timing bug with arp caching - it's not necessary for reachability in the normal case. This was an issue reported long-ago in containernetworking/cni#88 and containernetworking/cni#194.

So, yes, it will work without it, but it does help solve that bug in degenerate cases. And sorry for not writing a better comment :-)

squeed: Correct - we send the gratuitous arp from the container to the host.

jingax10: I also updated the comments accordingly. Please let me know any further comment. Thanks!

jingax10:
@squeed I just synced to the master. Please take another look. Thanks!

jingax10:
@squeed friendly ping

jingax10:
@squeed I just synced to the master. Please take another look. We plan to use PTP on GCP so we want to merge this pull request. Thanks!

squeed:
Hi there,
We talked about this at the maintainer's meeting. How would you feel about this being a separate "tethering" plugin?

jingax10:
Thanks @squeed! Could you please explain more on "tethering" plugin?

Though I introduced a new option for PTP plugin, I would see no issue to make it as the default setting while removing the current one (e.g., assigning gateway IP to veth in host namespace etc.).

jingax10:
@squeed, could you please clarify your point? I really want to move it forward. Thanks!

thockin:
@jingax10 What I don't see here is docs explaining that this feature exists, what type the field is, what it does, or how to use it.

@squeed Can you explain what you mean by "tethering" ?  To my eyes, this seems like an all-around improvement - am I missing somethign?

thockin:
Ping on this?  I am not an L2 expert, so my main question is whether there's any real reason to WANT to do things the old way vs this way?

squeed:
Apologies for not getting back sooner. I had this on the top of my to-do list for weeks, and we all know how that works.

Anyways, I've been looking at this for some time now, and I realize now that I jumped to a false conclusion regarding the goal of this plugin. I had previously spent some time building (and ultimately throwing away as undesired) a plugin that emulated a layer-2 attachment without bridging via arp and ndp proxy. I see now that this isn't the goal of this change, rather something smaller in scope.

Just to make sure I understand this correctly, the goal of this change is to eliminate assigning the gateway address to the host-side of the veth, to cut down on routing table entries, right? I haven't really run in to situations where that would ever matter, but I trust you that can help.

To merge this, I'd like to test with flannel, the firewall, and the bandwidth shaping plugin. I think it should work, but will need to test.

jingax10:
@squeed Tests are fixed by merging containernetworking/cni#562. PTAL. Thanks!

rosenhouse:
Am I understanding correctly that this is now a behavior change to the plugin which results in ARP messages now flowing when they didn't before?

dcbw:
@jingax10 would you mind rebasing this PR on top of master so we can get a clean Travis run? Looks like it's almost ready to merge, thanks!

jingax10:
Sorry for late reply. I just rebased the PR. I still need to fix bandwidth's test as it may assume the gateway IP if I remember correctly. I will ping when I get it done.

squeed:
Awesome! Let us know if you need anything.

malozanoff:
Hi, we were looking at how other people handled this, we do this same thing in our CNI plugin and it just broke with kube-proxy 1.13 (https://github.com/kubernetes/kubernetes/issues/71555) just letting you know so you can test this with that before merging. We are going back to adding the link-local ip on host side probably.

squeed:
Hey @jingax10 we'd like to get this in the v0.8.0 release - do you have time to work on this?

jingax10:
Sorry for the late reply. I actually tried almost a month back and I found kubernetes/kubernetes#71555, where net.ipv4.conf.all.arp_ignore=1, would make my solution stop working :-(

Not sure this PR is still worth to move forward (e.g., add a new option in CNI spec, e.g., proxy_arp: enabled). Please advise. Thanks!

BSWANG:
In the [Terway](https://github.com/AliyunContainerService/terway) plugin. We use another solution, set a permanent arp record to the `host-veth` as fake gateway in container, eg:

In container netns:
```
ip neigh add 169.254.1.1 dev eth0 lladdr <host-veth-mac-address> nud permanent
ip route add 169.254.1.1 dev eth0
ip route add default dev eth0
```

In host netns:
```
ip route add <container-ip> dev <host-side-veth-of-container>
```

The `host-side veth` no need to config gateway ip address or `proxy_arp`.

mars1024:
I think @BSWANG has provided a way to move forward, if possible, could you pull a new request based on this? Thanks @BSWANG  cc @jingax10

PR #144, opened 2018-04-12
build: some small improvements; bump to go1.10

squeed:
- bump to go 1.10
- Add a linker tag with the build version
- Remove fastbuild, go builds are cached now
- use better ginkgo suite names


squeed:
@dcbw now I know what those "weird symlinks" are for :-)

PR #159, opened 2018-06-11
Introduce iplink plugin to change netlink attr (i.e. 'ip link' command)

s1061123:
This change introduces iplink plugin, used to change interface attributes with netlink, like 'ip link' command. Currently MTU, MAC and promiscas are supported.

### DIFF
--- a/plugins/meta/iplink/README.md
+++ b/plugins/meta/iplink/README.md
@@ -1,0 +1,54 @@
+# iplink plugin
+
+## Overview
+
+This plugin changes some interface attributes, like 'ip link' command.
+It does not create any network interfaces and therefore does not bring connectivity by itself.
+This iplink plugin applies changes to interfaces created by previously applied plugins as meta plugin.
+It is only useful when used in addition to other plugins.
+
+## Example configuration
+
+```json
+{
+  "cniVersion": "0.3.1",
+  "name": "mynet",
+  "plugins": [
+    {
+      "type": "ptp",
+      "ipMasq": true,
+      "mtu": 512,
+      "ipam": {
+          "type": "host-local",
+          "subnet": "10.0.0.0/24"
+      },
+      "dns": {
+        "nameservers": [ "10.1.0.1" ]
+      }
+    },
+    {
+      "name": "myiplink",
+      "type": "iplink",
+      "promisc": true,
+      "mac": "c2:b0:57:49:47:f1",
+      "mtu": 1454
+    }
+  ]
+}
+```
+
+## Network configuration reference
+
+* `type` (string, required): "iplink"
+* `mac` (string, optional): MAC address (i.e. hardware address) of interface
+* `mtu` (integer, optional): MTU of interface
+* `promisc` (bool, optional): Change the promiscas mode of interface
+
+## Supported arguments
+The following [CNI_ARGS](https://github.com/containernetworking/cni/blob/master/SPEC.md#parameters) are supported:
+
+* `MAC`: request a specific MAC address for the interface 
+
+    (example: CNI_ARGS="IgnoreUnknown=true;MAC=c2:11:22:33:44:55")
+
+Note: You may add `IgnoreUnknown=true` to allow loose CNI argument verification (see CNI's issue[#560](https://github.com/containernetworking/cni/issues/560)).
--- a/plugins/meta/iplink/README.md
+++ b/plugins/meta/iplink/README.md
@@ -2,10 +2,9 @@
 
 ## Overview
 
-This plugin changes some interface attributes, like 'ip link' command.
-It does not create any network interfaces and therefore does not bring connectivity by itself.
-This iplink plugin applies changes to interfaces created by previously applied plugins as meta plugin.
-It is only useful when used in addition to other plugins.
+This plugin provides a method to change interface attributes, like 'ip link' command.
+It can be used in a chain to apply several specific changes to interface previously configured
+by other plugin.
 
 ## Example configuration
 
--- a/plugins/meta/iplink/main.go
+++ b/plugins/meta/iplink/main.go
@@ -1,0 +1,223 @@
+// Copyright 2018 CNI authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// This is a "meta-plugin". It reads in its own netconf, it does not create
+// any network interface but just changes the network sysctl.
+
+package main
+
+import (
+	"encoding/json"
+	"fmt"
+	"net"
+
+	"github.com/containernetworking/cni/pkg/skel"
+	"github.com/containernetworking/cni/pkg/types"
+	"github.com/containernetworking/cni/pkg/types/current"
+	"github.com/containernetworking/cni/pkg/version"
+	"github.com/containernetworking/plugins/pkg/ns"
+	"github.com/vishvananda/netlink"
+)
+
+/*
+ * Current support ip link command:
+ + ip link set promisc <bool>
+ + ip link set address <mac addr>
+ + ip link set mtu <numeric>
+
+--- syntax
+{
+  "name": "myiplink",
+  "type": "iplink",
+  "iplink": {
+          "promisc": "true",
+          "macaddress": "new mac addr",
+          "mtu": "1454",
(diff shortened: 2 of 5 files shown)


s1061123:
Will update diff to resolve CI...

plugins/meta/iplink/README.md
```diff
@@ -0,0 +1,54 @@
+# iplink plugin
+
+## Overview
+
+This plugin changes some interface attributes, like 'ip link' command.
```
jellonek: This plugin provides a method to change interface attributes, linke `ip link` command.

plugins/meta/iplink/README.md
```diff
@@ -0,0 +1,54 @@
...
+It does not create any network interfaces and therefore does not bring connectivity by itself.
+This iplink plugin applies changes to interfaces created by previously applied plugins as meta plugin.
```
jellonek: It can be used in a chain to apply several specific changes to interface previously configured by other plugin.

plugins/meta/iplink/main.go
```diff
@@ -0,0 +1,223 @@
...
+  "type": "iplink",
+  "iplink": {
+          "promisc": "true",
+          "macaddress": "new mac addr",
+          "mtu": "1454",
+  }
+}
+*/
+
+// TuningConf represents the network tuning configuration.
```
jellonek: Wrong name.

plugins/meta/iplink/main.go
```diff
@@ -0,0 +1,223 @@
...
+
+func changeMacAddr(ifName string, newMacAddr string) error {
+	addr, err := net.ParseMAC(newMacAddr)
+	if err != nil {
+		return fmt.Errorf("Invalid args %v for MAC addr: %v", newMacAddr, err)
+	}
+
+	link, err := netlink.LinkByName(ifName)
+	if err != nil {
+		return fmt.Errorf("Failed to get %s: %v", ifName, err)
```
jellonek: Please use `%q` instead of `%s`.

plugins/meta/iplink/main.go
```diff
@@ -0,0 +1,223 @@
...
+	}
+
+	err = netlink.LinkSetDown(link)
+	if err != nil {
+		return fmt.Errorf("Failed to set %s down: %v", ifName, err)
```
jellonek: Same as above.

plugins/meta/iplink/main.go
```diff
@@ -0,0 +1,223 @@
...
+	}
+	err = netlink.LinkSetHardwareAddr(link, addr)
+	if err != nil {
+		return fmt.Errorf("Failed to set %s address to %s: %v", ifName, newMacAddr, err)
```
jellonek: Same as above.

plugins/meta/iplink/main.go
```diff
@@ -0,0 +1,223 @@
...
+			i.Mac = newMacAddr
+		}
+	}
+	return nil
+}
+
+func changePromisc(ifName string, val bool) error {
+	link, err := netlink.LinkByName(ifName)
+	if err != nil {
+		return fmt.Errorf("Failed to get %s: %v", ifName, err)
```
jellonek: Same as above.

plugins/meta/iplink/main.go
```diff
@@ -0,0 +1,223 @@
...
+
+type MACEnvArgs struct {
+	types.CommonArgs
+	MAC types.UnmarshallableString `json:"mac,omitempty"`
+}
+
+func changeMacAddr(ifName string, newMacAddr string) error {
+	addr, err := net.ParseMAC(newMacAddr)
+	if err != nil {
+		return fmt.Errorf("Invalid args %v for MAC addr: %v", newMacAddr, err)
```
jellonek: Please start message with lower case if it's returned to other func and not displayed directly to user.

plugins/meta/iplink/main.go
```diff
@@ -0,0 +1,223 @@
...
+	if val {
+		return netlink.SetPromiscOn(link)
+	}
+	return netlink.SetPromiscOff(link)
+}
+
+func changeMtu(ifName string, mtu int) error {
+	link, err := netlink.LinkByName(ifName)
+	if err != nil {
+		return fmt.Errorf("Failed to get %s: %v", ifName, err)
```
jellonek: Same as above.

plugins/meta/iplink/main.go
```diff
@@ -0,0 +1,223 @@
...
+	return &conf, nil
+}
+
+func cmdAdd(args *skel.CmdArgs) error {
+	conf, err := parseConf(args.StdinData, args.Args)
+	if err != nil {
+		return err
+	}
+
+	err = ns.WithNetNSPath(args.Netns, func(_ ns.NetNS) error {
```
jellonek: You can skip `_ ` leaving plaing `func(ns.NetNS)`.

plugins/meta/iplink/main.go
```diff
@@ -0,0 +1,223 @@
...
+		}
+
+		if conf.Promisc != false {
+			if err = changePromisc(args.IfName, true); err != nil {
+				return err
+			}
+		}
+
+		if conf.Mtu != -1 {
+			if err = changeMtu(args.IfName, conf.Mtu); err != nil {
```
jellonek: MTU modified in this place is not updated in returned `Result`. It should be updated in similar way as hwaddr is updated.

plugins/meta/iplink/main.go
```diff
@@ -0,0 +1,223 @@
...
+		return err
+	}
+
+	return types.PrintResult(conf.PrevResult, conf.CNIVersion)
+}
+
+func cmdDel(args *skel.CmdArgs) error {
+	// TODO: the settings are not reverted to the previous values. Reverting the
+	// settings is not useful when the whole container goes away but it could be
+	// useful in scenarios where plugins are added and removed at runtime.
```
jellonek: Different plugins can have a different expectations about found situation. Some of them can possibly scream about an error if they will see in netns other configuration than one which they leaved at their end. I don't know any of such plugins which would break because of that at the moment e.g. when mtu is different from configured by it - but there is such possibility, contract is a bit violated.

plugins/meta/iplink/main.go
s1061123: Currently there is no 'mtu' in CNI specification, https://github.com/containernetworking/cni/blob/master/SPEC.md#result, hence no need to update.

Adding 'mtu' into result should be taken care of another issues/PR, I suppose.

plugins/meta/iplink/main.go
s1061123: It should be depend on each case, I guess and CNI spec does not take care about it for now.
Currently CNI only provide a way to combination (i.e. main plugin/ipam/meta) and no strict rules to combine plugins. So user can configure flannel + host-local or bridge(with mtu param) and there is no violation from CNI spec point of view.

I believe it should be discussed in CNI community if you' would like to clarify, not this PR.

plugins/meta/iplink/README.md
s1061123: Fixed

plugins/meta/iplink/README.md
s1061123: Fixed

plugins/meta/iplink/main.go
s1061123: Fixed

plugins/meta/iplink/main.go
s1061123: Fixed

plugins/meta/iplink/main.go
s1061123: Fixed

plugins/meta/iplink/main.go
s1061123: Fixed

plugins/meta/iplink/main.go
s1061123: Fixed

plugins/meta/iplink/main.go
s1061123: Fixed

plugins/meta/iplink/main.go
s1061123: Fixed

plugins/meta/iplink/main.go
s1061123: Fixed

plugins/meta/iplink/main.go
jellonek: Ah, you are right.

squeed:
@s1061123 we chatted about this at the maintainer's meeting, and were wondering if you'd like to fold this in to the tuning plugin? It might make sense, given that both tweak an existing interface. Thoughts?

s1061123:
@squeed Thank you for your comments! I am okey to merge it into tuning plugin. I suppose this feature is iplink feature, not tuning, that's why I add new CNI plugin, iplink. 

Let me add new PR for that.

Well, 'tuning' may include a lot of feature (well, bandwidth feature is also a kind of 'tuning' in some sense, right?) and that is another reason...

PR #165, opened 2018-06-18
Support CNI_ARGS in static IPAM plugin

s1061123:
This change is to add CNI_ARGS support in static IPAM plugin.
When IP/SUBNET/GATEWAY are given in CNI_ARGS, static IPAM adds
these info in addition to config files.

To configure ip address only from CNI_ARGS, 'address' field in config
is changed to optional from required.

### DIFF
--- a/plugins/ipam/static/README.md
+++ b/plugins/ipam/static/README.md
@@ -38,16 +38,17 @@ static IPAM is very simple IPAM plugin that assigns IPv4 and IPv6 addresses stat
 ## Network configuration reference
 
 * `type` (string, required): "static"
-* `addresses` (array, optional): an array of arrays of ip address objects:
+* `addresses` (array, optional): an array of ip address objects:
 	* `address` (string, required): CIDR notation IP address.
 	* `gateway` (string, optional): IP inside of "subnet" to designate as the gateway.
 * `routes` (string, optional): list of routes add to the container namespace. Each route is a dictionary with "dst" and optional "gw" fields. If "gw" is omitted, value of "gateway" will be used.
 * `dns` (string, optional): the dictionary with "nameservers", "domain" and "search".
 
 ## Supported arguments
+
 The following [CNI_ARGS](https://github.com/containernetworking/cni/blob/master/SPEC.md#parameters) are supported:
 
-* `IP`: request a specific IP address from a subnet
+* `IP`: request a specific IP address
 * `SUBNET`: request a specific subnet
 * `GATEWAY`: request a specific gateway address
 
--- a/plugins/ipam/static/README.md
+++ b/plugins/ipam/static/README.md
@@ -48,7 +48,7 @@ static IPAM is very simple IPAM plugin that assigns IPv4 and IPv6 addresses stat
 
 The following [CNI_ARGS](https://github.com/containernetworking/cni/blob/master/SPEC.md#parameters) are supported:
 
-* `IP`: request a specific CIDR notation IP address
+* `IP`: request a specific CIDR notation IP addresses, comma separated
 * `GATEWAY`: request a specific gateway address
 
     (example: CNI_ARGS="IP=10.10.0.1/24;GATEWAY=10.10.0.254")
--- a/plugins/ipam/static/README.md
+++ b/plugins/ipam/static/README.md
@@ -48,8 +48,7 @@ static IPAM is very simple IPAM plugin that assigns IPv4 and IPv6 addresses stat
 
 The following [CNI_ARGS](https://github.com/containernetworking/cni/blob/master/SPEC.md#parameters) are supported:
 
-* `IP`: request a specific IP address
-* `SUBNET`: request a specific subnet
+* `IP`: request a specific CIDR notation IP address
 * `GATEWAY`: request a specific gateway address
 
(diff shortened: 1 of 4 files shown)


plugins/ipam/static/README.md
```diff
@@ -38,8 +38,17 @@ static IPAM is very simple IPAM plugin that assigns IPv4 and IPv6 addresses stat
...
 * `type` (string, required): "static"
-* `addresses` (array, required): an array of arrays of ip address objects:
+* `addresses` (array, optional): an array of arrays of ip address objects:
 	* `address` (string, required): CIDR notation IP address.
 	* `gateway` (string, optional): IP inside of "subnet" to designate as the gateway.
 * `routes` (string, optional): list of routes add to the container namespace. Each route is a dictionary with "dst" and optional "gw" fields. If "gw" is omitted, value of "gateway" will be used.
 * `dns` (string, optional): the dictionary with "nameservers", "domain" and "search".
+
+## Supported arguments
+The following [CNI_ARGS](https://github.com/containernetworking/cni/blob/master/SPEC.md#parameters) are supported:
```
jellonek: Missing new line before this one. Please keep that in sync with rest of document.

plugins/ipam/static/README.md
```diff
@@ -38,8 +38,17 @@ static IPAM is very simple IPAM plugin that assigns IPv4 and IPv6 addresses stat
...
+
+* `IP`: request a specific IP address from a subnet
```
jellonek: Please remove ` from a subnet`.

plugins/ipam/static/README.md
```diff
@@ -38,8 +38,17 @@ static IPAM is very simple IPAM plugin that assigns IPv4 and IPv6 addresses stat
 ## Network configuration reference
 
 * `type` (string, required): "static"
-* `addresses` (array, required): an array of arrays of ip address objects:
+* `addresses` (array, optional): an array of arrays of ip address objects:
```
jellonek: That's incorrect even in master - it's an array of ip addresses, not an array of arrays.

plugins/ipam/static/README.md
s1061123: Fixed.

plugins/ipam/static/README.md
s1061123: Fixed.

plugins/ipam/static/README.md
s1061123: Fixed.

squeed:
Seems fine, except it adds a new runtime convention. others might care about this.

s1061123:
@squeed thank you for your comments. Let me clarify that your comment indicate to remove runtime convention, correct? I will remove runtime and update PR.

s1061123:
@squeed I've taken care of your comments. Could you please take a look into it?

dcbw:
@s1061123 thanks for the update, discussed this one during maintainer's meeting.  Instead of having two variables for IP and SUBNET, could we instead combine them into just IP that would have a CIDR format eg "1.2.3.4/24"?  You can give that to net.ParseCIDR() and you'll get both the IP address itself and the subnet mask out.

s1061123:
@dcbw and other CNI committer, I need to modify a bit to support multiple ip address ("10.2.2.42/24" and "2001:db8::5/64").

s1061123:
@dcbw , I updated the PR.

plugins/ipam/static/main.go
```diff
@@ -109,6 +117,50 @@ func LoadIPAMConfig(bytes []byte, envArgs string) (*IPAMConfig, string, error) {
...
+		if e.GATEWAY != "" {
+			for _, item := range strings.Split(string(e.GATEWAY), ",") {
+				gwip := net.ParseIP(strings.TrimSpace(item))
+				if gwip == nil {
+					return nil, "", fmt.Errorf("invalid gateway address: %s", item)
+				}
+
+				for i := range n.IPAM.Addresses {
+					if n.IPAM.Addresses[i].Address.Contains(gwip) {
+						n.IPAM.Addresses[i].Gateway = gwip
```
dcbw: I would parse GATEWAY before IP, and hten assign the gateways to each Address while parsing the IPs.  We should also make sure that the GATEWAY array is the same length as the IP array.

plugins/ipam/static/main.go
```diff
@@ -109,6 +117,50 @@ func LoadIPAMConfig(bytes []byte, envArgs string) (*IPAMConfig, string, error) {
...
+			return nil, "", err
+		}
+
+		if e.IP != "" {
+			for _, item := range strings.Split(string(e.IP), ",") {
+				ipstr := strings.TrimSpace(item)
+
+				ip, subnet, err := net.ParseCIDR(ipstr)
+				if err != nil {
+					return nil, "", fmt.Errorf("invalid CIDR %s: %s", e.IP, err)
```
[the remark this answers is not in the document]
s1061123: Sure. I will change it.

plugins/ipam/static/main.go
s1061123: In some case, there is no gateway corresponding to certain IP address, especially multiple interface case. So I'd like to parse IP first then match gateway if there exists. Is that okey to you?

dcbw: @s1061123 yeah, you're right.

PR #166, opened 2018-06-18
plugins/host-local: ensure subnet is a network address

### LINKED ISSUE #161: IPV6 subnets specified with a full address generate unexpected range
host-local IPAM plugin

If I specify a subnet CIDR using a full address, then the range seems to start at that address rather than the default addresses for the range. 

i.e. the code seems to use the address and not the network to allocate the default range.

e.g 

```
{
          "subnet": "2a02:1348:178:7112:24:19ff:fee1:c44a/64"
}
```

Changing the 'should generate sane defaults for ipv6' spec in `allocator/range_test.go` so the subnet string is an address with a mask, not a clean prefix causes the spec to fail.

NeilW:
Allocation code assumes the specified subnet is a clean network address
prefix, so check that is the case and throw an error otherwise

Fixes #161


squeed:
Thanks this is great!

PR #169, opened 2018-07-02
Fix tc-tbf burst value in bytes

hustcat:
According to [iproute2](https://github.com/hustcat/iproute2/blob/master/tc/q_tbf.c#L93), we should calculate the buffer in bytes for `rate` and `burst` both. But the current `burs`t is in bit unit.

### DIFF
--- a/plugins/meta/bandwidth/bandwidth_linux_test.go
+++ b/plugins/meta/bandwidth/bandwidth_linux_test.go
@@ -80,7 +80,7 @@ var _ = Describe("bandwidth test", func() {
 	"ingressRate": 8,
 	"ingressBurst": 8,
 	"egressRate": 16,
-	"egressBurst": 9,
+	"egressBurst": 8,
 	"prevResult": {
 		"interfaces": [
 			{
@@ -135,7 +135,7 @@ var _ = Describe("bandwidth test", func() {
 
 				Expect(qdiscs[0]).To(BeAssignableToTypeOf(&netlink.Tbf{}))
 				Expect(qdiscs[0].(*netlink.Tbf).Rate).To(Equal(uint64(2)))
-				Expect(qdiscs[0].(*netlink.Tbf).Limit).To(Equal(uint32(9)))
+				Expect(qdiscs[0].(*netlink.Tbf).Limit).To(Equal(uint32(1)))
 
 				hostVethLink, err := netlink.LinkByName(hostIfname)
 				Expect(err).NotTo(HaveOccurred())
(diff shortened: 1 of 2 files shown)


hustcat:
@rosenhouse Updated, PTAL

squeed:
Hmm. I tested the original bandwidth plugin and it seemed to create the correct values, at least when I viewed the queues. But I don't know enough about this stuff.

hustcat:
@squeed  This is my test with original bandwidth tool before this PR:

* Create veth pair

```
# ip netns add netns1
# ip link add veth_host type veth peer name veth_con
# ip link set veth_con netns netns1
# ip link show veth_host
2719: veth_host@if2718: <BROADCAST,MULTICAST> mtu 1500 qdisc noop state DOWN mode DEFAULT qlen 1000
    link/ether b2:1d:56:6a:ff:3c brd ff:ff:ff:ff:ff:ff link-netnsid 5
```

* Set `rate` to `8000bit` and `burt` to `80bit` by bandwidth

```
# cat bandwidth.conf 
{
"cniVersion": "0.3.1",
"name": "slowdown",
"type": "bandwidth",
"ingressRate": 8000,
"ingressBurst": 80,
"egressRate": 8000,
"egressBurst": 80,
"prevResult": {"cniVersion":"0.3.1", "interfaces": [{"name": "veth_host","mac": "b2:1d:56:6a:ff:3c", "sandbox": ""}]}
}

# CNI_PATH=`pwd`/bin CNI_COMMAND=ADD CNI_NETNS=/var/run/netns/netns1 CNI_CONTAINERID=netns1 CNI_IFNAME=veth_con CNI_ARGS="" bin/bandwidth < bandwidth.conf
{
    "cniVersion": "0.3.1",
    "interfaces": [
        {
            "name": "veth_host",
            "mac": "b2:1d:56:6a:ff:3c"
        },
        {
            "name": "5026",
            "mac": "7a:49:f8:e0:ff:9e"
        }
    ],
    "dns": {}
}

# ./tc -r qdisc show dev veth_host
qdisc tbf 1:[00010000] root refcnt 2 rate 8Kbit burst 80b [001312d0] lat 25.0ms limit 105b 
qdisc ingress ffff:[ffff0000] parent ffff:fff1 ---------------- 
```

* Set `rate` to `8000bit` and `burt` to `80bit` by `tc`

```
# tc qdisc delete dev veth_host root handle 1:
# tc qdisc delete dev veth_host ingress


# tc qdisc add dev veth_host handle 1: root tbf rate 8kbit burst 10b latency 25ms

# ./tc  -r qdisc show dev veth_host                                                 
qdisc tbf 1:[00010000] root refcnt 2 rate 8Kbit burst 10b [0002625a] lat 25.0ms limit 35b
```

As we can see, the values of `burst` and `limit` are not same between the `bandwidth` and `tc` ouput.

hustcat:
Maybe we should make `ingressBurst` and `egressBurst` use byte as unit, but not bit. See [here](https://github.com/containernetworking/plugins/blob/master/plugins/meta/bandwidth/ifb_creator.go#L116).

squeed:
Ahh, I see. Thanks for the explanation. I definitely want to keep the config API the same (bits), so your fix seems right.

PR #170, opened 2018-07-02
ipam/host-local: Update for spec v0.7

squeed:
This updates the host-local ipam allocator for v0.7:
- It now supports GET
- It will allow allocating multiple IPs per network (with different
    interfaces)

### DIFF
--- a/plugins/ipam/host-local/backend/disk/backend.go
+++ b/plugins/ipam/host-local/backend/disk/backend.go
@@ -15,14 +15,15 @@
 package disk
 
 import (
+	"fmt"
 	"io/ioutil"
 	"net"
 	"os"
 	"path/filepath"
+	"runtime"
 	"strings"
 
 	"github.com/containernetworking/plugins/plugins/ipam/host-local/backend"
-	"runtime"
 )
 
 const lastIPFilePrefix = "last_reserved_ip."
@@ -39,6 +40,7 @@ type Store struct {
 // Store implements the Store interface
 var _ backend.Store = &Store{}
 
+// New creates a new store
 func New(network, dataDir string) (*Store, error) {
 	if dataDir == "" {
 		dataDir = defaultDataDir
@@ -55,7 +57,9 @@ func New(network, dataDir string) (*Store, error) {
 	return &Store{lk, dir}, nil
 }
 
-func (s *Store) Reserve(id string, ip net.IP, rangeID string) (bool, error) {
+// Reserve attempts to claim an ip in the name of a container with the given id.
+// The store should be locked before calling this.
+func (s *Store) Reserve(id backend.Key, ip net.IP, rangeID string) (bool, error) {
 	fname := GetEscapedPath(s.dataDir, ip.String())
 
 	f, err := os.OpenFile(fname, os.O_RDWR|os.O_EXCL|os.O_CREATE, 0644)
@@ -65,7 +69,7 @@ func (s *Store) Reserve(id string, ip net.IP, rangeID string) (bool, error) {
 	if err != nil {
 		return false, err
 	}
-	if _, err := f.WriteString(strings.TrimSpace(id)); err != nil {
+	if _, err := f.Write(WriteKey(&id)); err != nil {
 		f.Close()
 		os.Remove(f.Name())
 		return false, err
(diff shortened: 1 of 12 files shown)


plugins/ipam/host-local/backend/disk/backend.go
```diff
@@ -55,7 +57,9 @@ func New(network, dataDir string) (*Store, error) {
 	return &Store{lk, dir}, nil
 }
 
-func (s *Store) Reserve(id string, ip net.IP, rangeID string) (bool, error) {
+// Reserve attempts to claim an ip in the name of a container with the given id.
```
bboreham: I'd avoid "name" here, since containers do have names, and we are using ID+interface as the key.

How about "claim an ip for a given container interface" ?

plugins/ipam/host-local/backend/disk/backend.go
```diff
@@ -93,28 +97,73 @@ func (s *Store) LastReservedIP(rangeID string) (net.IP, error) {
...
+			if err != nil {
+				fmt.Fprintf(os.Stderr, "WARN host-local: failed to remove ip reservation file %s: %v", path, err)
+			}
+		}
+		return nil
+	})
+	return err
+}
+
+// GetByID returns all IPs reserved by a given container ID
```
bboreham: "container ID" is a specific thing, and we are using ID+interface (with some fuzzing to cope with older stores).  Maybe "a given key" ?

plugins/ipam/host-local/backend/disk/backend.go
```diff
@@ -93,28 +97,73 @@ func (s *Store) LastReservedIP(rangeID string) (net.IP, error) {
...
+func (s *Store) GetByID(id backend.Key) ([]net.IP, error) {
+	out := []net.IP{}
+	err := s.walkReservations(func(path string, ip net.IP, b []byte) error {
+		matches, err := Matches(&id, b)
 		if err != nil {
+			fmt.Fprintf(os.Stderr, "WARN: host-local: Failed to parse ip reservation file %s: %v", path, err)
```
bboreham: Surely "failed to parse" is something the lower-level function would know; all we know here is something went wrong.
(And in fact it looks like we treat "failed to parse" as "binary file")

plugins/ipam/host-local/backend/disk/key.go
```diff
@@ -0,0 +1,114 @@
...
+		return false, err
+	}
+
+	switch kind {
+	case KindReservationV0:
+		return k.ID == k1.ID, nil
+	case KindReservationV1:
+		return k.Equals(k1), nil
+	}
+	return false, fmt.Errorf("unreachable")
```
bboreham: Maybe just `panic()` ?

plugins/ipam/host-local/backend/disk/backend.go
```diff
@@ -124,3 +173,10 @@ func GetEscapedPath(dataDir string, fname string) string {
 	}
 	return filepath.Join(dataDir, fname)
 }
+
+func UnescapePath(fname string) net.IP {
```
bboreham: Asymmetry between `GetEscapedPath` and `UnescapePath `.
Maybe `EscapedPathToIP` ?

plugins/ipam/host-local/backend/disk/backend.go
jellonek: Also instead of `%s` please use `%q`.

plugins/ipam/host-local/backend/disk/backend.go
```diff
@@ -93,28 +97,73 @@ func (s *Store) LastReservedIP(rangeID string) (net.IP, error) {
...
 // release as much as possible
-func (s *Store) ReleaseByID(id string) error {
-	err := filepath.Walk(s.dataDir, func(path string, info os.FileInfo, err error) error {
-		if err != nil || info.IsDir() {
+//
+func (s *Store) ReleaseByID(id backend.Key) error {
+	err := s.walkReservations(func(path string, _ net.IP, b []byte) error {
+		match, err := Matches(&id, b)
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "WARN: host-local: Failed to parse ip reservation file %s: %v", path, err)
```
jellonek: Please use `%q` in place of `%s`.

plugins/ipam/host-local/backend/disk/backend.go
```diff
@@ -93,28 +97,73 @@ func (s *Store) LastReservedIP(rangeID string) (net.IP, error) {
...
 			return nil
 		}
-		data, err := ioutil.ReadFile(path)
+		if match {
+			err := os.Remove(path)
+			if err != nil {
+				fmt.Fprintf(os.Stderr, "WARN host-local: failed to remove ip reservation file %s: %v", path, err)
```
jellonek: Same as above.

plugins/ipam/host-local/backend/disk/backend.go
```diff
@@ -93,28 +97,73 @@ func (s *Store) LastReservedIP(rangeID string) (net.IP, error) {
...
+			return nil
+		}
+		ip := UnescapePath(info.Name())
+		if len(ip) == 0 { // not a reservation file
+			return nil
+		}
+
+		b, err := ioutil.ReadFile(path)
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "WARN: host-local: Failed to read ip reservation file %s: %v", path, err)
```
jellonek: Same as above.

dcbw:
Closed in preference to https://github.com/containernetworking/plugins/pull/203

PR #55, opened 2017-08-15
[RFC] add GET support to plugins

dcbw:

PR #172, opened 2018-07-04
Add GET for bridge plugin

bboreham:
Copied from https://github.com/containernetworking/plugins/pull/55/files by @dcbw

### DIFF
--- a/plugins/main/bridge/bridge.go
+++ b/plugins/main/bridge/bridge.go
@@ -507,6 +507,77 @@ func main() {
 }
 
 func cmdGet(args *skel.CmdArgs) error {
-	// TODO: implement
-	return fmt.Errorf("not implemented")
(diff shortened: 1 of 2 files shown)


plugins/main/bridge/bridge.go
```diff
@@ -507,6 +507,77 @@ func main() {
...
+	}
+
+	br, err := bridgeByName(n.BrName)
+	if err != nil {
+		return err
+	}
+
+	var containerVeth netlink.Link
+	hostVethIndex := -1
+	if err := netns.Do(func(_ ns.NetNS) error {
```
jellonek: IMO it's better to use form `func(ns.NetNS)` without underscore, but I can live with that ;)

bboreham:
The CNI maintainers are re-thinking GET, so I'll close for now.

PR #175, opened 2018-07-11
fix_loopback_readme

liucimin:
The loopback cni just link up the loopback interface.
It dont create the loopback interface.And the loopback interface is create by docker none network.


bboreham:
Good spot!  Thanks for this, it's true it doesn't create the interface.  However "link up" has connotations that aren't right either.  How about "bring up"?   Or something longer - "set state to 'up'" ?

liucimin:
@bboreham  
Thx for your suggestion.
I have modified the description of the loopback cni.

PR #176, opened 2018-07-20
add extraRoutes config for flannel ipam

kadisi:
Signed-off-by: zhangjie <iamkadisi@163.com>

**What this PR does / why we need it**:
if my runtime is containerd, and cni plugin is flannel(host-gw model), and kubernetes。 By default, the flannel plugin only uses flanneld  /run/flannel/subnet.env  to set one route for the container. This will cause the container to communicate with the container only. but container can not access non-pod cluster network.. The purpose of this pr is to make the container access other network.
**Special notes for your reviewer**:
flannel cni config example：
```
{
    "name": "mynet",
    "ipMasq": false,
    "type": "flannel",
    "delegate": {
        "extraRoutes": [
            {
                "dst": "0.0.0.0/0"
            }
        ]
    }
}
```
this will set container `0.0.0.0/0` route , such as 
```
Destination     Gateway         Genmask         Flags Metric Ref    Use Iface
0.0.0.0         172.30.2.1      0.0.0.0         UG    0      0        0 eth0
172.30.0.0      172.30.2.1      255.255.0.0     UG    0      0        0 eth0
172.30.2.0      0.0.0.0         255.255.255.0   U     0      0        0 eth0
```

**Release note**:
```
add extraRoutes config for flannal ipam
```


plugins/meta/flannel/flannel.go
```diff
@@ -36,8 +36,16 @@ import (
 )
 
 const (
-	defaultSubnetFile = "/run/flannel/subnet.env"
-	defaultDataDir    = "/var/lib/cni/flannel"
+	defaultSubnetFile         = "/run/flannel/subnet.env"
+	defaultDataDir            = "/var/lib/cni/flannel"
+	delegateKeyType           = "type"
```
squeed: I don't think it's necessary to define constants for all the struct keys.

kadisi: ok, thank you , and is there any other problem except this ?

kadisi:
@squeed  please review if you have time

squeed:
Now that I've had a bit of a think, I think extraRoutes might make sense to be a top-level parameter, not a parameter to delegate.

@tomdee @matthewdupre what do you think?

kadisi:
@squeed  @jellonek is that like this?
```
{
    "name": "mynet",
    "type": "flannel",
    "extraRoutes": [
        {
            "dst": "0.0.0.0/0"
        }
    ]
}
```
or 
```
{
    "name": "mynet",
    "type": "flannel",
    "makeDefaultGWThroughIt": true
}
```
?

jellonek:
Original proposal is imo more flexible, while last one is easiest to use in originally described case.
Lets wait for opinions of maintainers (i'm only contributor) what should be better.

kadisi:
@bboreham  do you mean it is no necessary to add extra Routes?

kadisi:
@squeed @bboreham

bboreham:
> @bboreham do you mean it is no necessary to add extra Routes?

I meant the code recognizes the absence of any routes as a signal to add one of its own.  The default config looks like this:

        {
            "name": "weave",
            "type": "weave-net",
        },

and by using this config you get a default route and gateway through the interface.

kadisi:
@bboreham PTAL

bboreham:
Sorry I don't follow what this is doing now.  If you think it's right, please explain step by step how the code achieves your aims.

PR #177, opened 2018-07-31
Tuninig iplink

s1061123:
This change adds 'ip link' command related feature into tuning
meta cni plugin. Currently MTU, MAC and promiscas mode are
supported.

This code comes from #159 and @squeed comments.


plugins/meta/tuning/tuning.go
```diff
@@ -90,6 +164,26 @@ func cmdAdd(args *skel.CmdArgs) error {
...
+			}
+		}
+
+		if tuningConf.Promisc != false {
+			if err = changePromisc(args.IfName, true); err != nil {
+				return err
+			}
+		}
+
+		if tuningConf.Mtu != -1 {
```
squeed: Since 0 is the golang default value, and a 0 MTU is invalid, you can just use that.

squeed:
A few small things, but otherwise looks great.

plugins/meta/tuning/tuning.go
```diff
@@ -90,6 +164,26 @@ func cmdAdd(args *skel.CmdArgs) error {
 				return err
 			}
 		}
+
+		var err error
+
+		if tuningConf.Mac != "" {
+			if err = changeMacAddr(args.IfName, tuningConf.Mac); err == nil {
+				err = updateResultsMacAddr(*tuningConf, args.IfName, tuningConf.Mac)
```
bboreham: I don't think this `err` is ever passed back.

bboreham:
Looks good overall; couple of nits.  Thanks!

plugins/meta/tuning/tuning.go
dcbw: @s1061123 since updateResultsMacAddr() never returns any thing except nil, I think you should just remove the return value from updateResultsMacAddr().

plugins/meta/tuning/tuning.go
s1061123: Fixed. Thanks!

plugins/meta/tuning/tuning.go
s1061123: I removed 'return' in updateResultsMacAddr(). Thank you all!

plugins/meta/tuning/tuning.go
```diff
@@ -90,6 +164,26 @@ func cmdAdd(args *skel.CmdArgs) error {
...
+			if err = changePromisc(args.IfName, true); err != nil {
+				return err
+			}
+		}
+
+		if tuningConf.Mtu != -1 {
+			if err = changeMtu(args.IfName, tuningConf.Mtu); err != nil {
+				return err
+			}
+		}
```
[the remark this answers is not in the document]
s1061123: In the 'else' case of changeMtu(), we only return nil to indicate we succeed to change mtu, so it is same as line187 "return nil", hence it should be okey.

plugins/meta/tuning/tuning.go
```diff
@@ -90,6 +163,26 @@ func cmdAdd(args *skel.CmdArgs) error {
 				return err
 			}
 		}
+
+		var err error
+
+		if tuningConf.Mac != "" {
+			if err = changeMacAddr(args.IfName, tuningConf.Mac); err == nil {
+				updateResultsMacAddr(*tuningConf, args.IfName, tuningConf.Mac)
+			}
```
bboreham: need `else return err`

plugins/meta/tuning/tuning.go
```diff
@@ -90,6 +163,26 @@ func cmdAdd(args *skel.CmdArgs) error {
 				return err
 			}
 		}
+
+		var err error
```
bboreham: I can't see what this gives, over `err :=` on the three subsequent uses

s1061123: Removed.

plugins/meta/tuning/tuning.go
s1061123: I understand that and fixed.

dcbw:
LGTM now.

dcbw:
@bboreham @squeed @jellonek PTAL thanks!

jellonek:
LGTM

PR #186, opened 2018-08-09
scripts: support building releases without a TTY

### LINKED ISSUE #183: Unable to run scripts/release.sh in Jenkins (not a TTY)
I'm trying to run `scripts/release.sh` in Jenkins, but this fails because the script has `docker run -t` and Jenkins executes its builds without a TTY. Would you accept a PR to remove the `-t`, or to only set the flag if `tty -s` returns a 0 exit code?

ncdc:
This allows systems such as Jenkins, which do not provide a TTY, to run
scripts/release.sh

Fixes #183 

Signed-off-by: Andy Goldstein <andy.goldstein@gmail.com>

### DIFF
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -15,7 +15,7 @@ rm -Rf ${SRC_DIR}/${RELEASE_DIR}
 mkdir -p ${SRC_DIR}/${RELEASE_DIR}
 mkdir -p ${OUTPUT_DIR}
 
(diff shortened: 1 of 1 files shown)


ncdc:
@squeed @dcbw PTAL!

dcbw:
LGTM, I have no idea why it was -it originally.

jellonek:
LGTM

PR #195, opened 2018-08-26
L2 bridge Implementation

SchSeba:
This PR Change the bridge plugin to allow empty IPAM.

When the IPAM dictionary is empty the bridge plugin will connect a pure L2 interface(without any IP address on the interface) from the pod to a bridge on the host.

This PR will allow containers and everything running on top of them to have a pure L2 interface connected to a bridge. This can be useful for use cases of sniffing, running dhclient, give the pod a trunk interface, etc..

### DIFF
--- a/plugins/main/bridge/README.md
+++ b/plugins/main/bridge/README.md
@@ -28,6 +28,17 @@ If the bridge is missing, the plugin will create one on first use and, if gatewa
 }
 ```
 
+## Example L2-only configuration
+```
+{
+    "cniVersion": "0.3.1",
+	"name": "mynet",
+	"type": "bridge",
+	"bridge": "mynet0",
+	"ipam": {}
+}
+```
+
 ## Network configuration reference
 
 * `name` (string, required): the name of the network.
@@ -39,5 +50,5 @@ If the bridge is missing, the plugin will create one on first use and, if gatewa
 * `ipMasq` (boolean, optional): set up IP Masquerade on the host for traffic originating from this network and destined outside of it. Defaults to false.
 * `mtu` (integer, optional): explicitly set MTU to the specified value. Defaults to the value chosen by the kernel.
 * `hairpinMode` (boolean, optional): set hairpin mode for interfaces on the bridge. Defaults to false.
-* `ipam` (dictionary, required): IPAM configuration to be used for this network.
+* `ipam` (dictionary, required): IPAM configuration to be used for this network. For L2-only network, create empty dictionary.
 * `promiscMode` (boolean, optional): set promiscuous mode on the bridge. Defaults to false.
--- a/plugins/main/bridge/bridge.go
+++ b/plugins/main/bridge/bridge.go
@@ -334,6 +334,8 @@ func cmdAdd(args *skel.CmdArgs) error {
 		return err
 	}
 
+	isLayer3 := n.IPAM.Type != ""
+
 	if n.IsDefaultGW {
 		n.IsGW = true
 	}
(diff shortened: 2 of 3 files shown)


plugins/main/bridge/README.md
```diff
@@ -28,6 +28,16 @@ If the bridge is missing, the plugin will create one on first use and, if gatewa
 }
 ```
 
+## Example for l2 configuration
```
dankenigsberg: I'd call this "L2-only configuration"

plugins/main/bridge/README.md
```diff
@@ -39,5 +49,5 @@ If the bridge is missing, the plugin will create one on first use and, if gatewa
 * `ipMasq` (boolean, optional): set up IP Masquerade on the host for traffic originating from this network and destined outside of it. Defaults to false.
 * `mtu` (integer, optional): explicitly set MTU to the specified value. Defaults to the value chosen by the kernel.
 * `hairpinMode` (boolean, optional): set hairpin mode for interfaces on the bridge. Defaults to false.
-* `ipam` (dictionary, required): IPAM configuration to be used for this network.
+* `ipam` (dictionary, required): IPAM configuration to be used for this network(for pure l2 network create empty dictionary).
```
dankenigsberg: Nicer English, imo:
```
 IPAM configuration to be used for this network. For L2-only network, create empty dictionary.
```

plugins/main/bridge/bridge.go
```diff
@@ -358,103 +360,109 @@ func cmdAdd(args *skel.CmdArgs) error {
 		return err
 	}
 
-	// run the IPAM plugin and get back the config to apply
-	r, err := ipam.ExecAdd(n.IPAM.Type, args.StdinData)
-	if err != nil {
-		return err
-	}
+	// Assume l2 interface only
```
dankenigsberg: Please use uppercase L2 for consistency and legibility.

SchSeba: Done

plugins/main/bridge/README.md
SchSeba: Done

plugins/main/bridge/README.md
SchSeba: Done

dcbw:
@SchSeba thanks for the update!  Can you add some testcases though?  Let's make sure we don't break anything.

plugins/main/bridge/bridge.go
```diff
@@ -485,41 +493,44 @@ func cmdDel(args *skel.CmdArgs) error {
...
-	if err != nil {
-		return err
-	}
+		// There is a netns so try to clean up. Delete can be called multiple times
+		// so don't return an error if the device is already removed.
+		// If the device isn't there then don't try to clean up IP masq either.
+		var ipnets []*net.IPNet
+		err = ns.WithNetNSPath(args.Netns, func(_ ns.NetNS) error {
+			var err error
+			ipnets, err = ip.DelLinkByNameAddr(args.IfName)
```
squeed: It looks like the container interface is only deleted in Layer3 mode? That doesn't seem right.

plugins/main/bridge/bridge.go
```diff
@@ -485,41 +493,44 @@ func cmdDel(args *skel.CmdArgs) error {
...
-	var ipnets []*net.IPNet
-	err = ns.WithNetNSPath(args.Netns, func(_ ns.NetNS) error {
-		var err error
-		ipnets, err = ip.DelLinkByNameAddr(args.IfName)
-		if err != nil && err == ip.ErrLinkNotFound {
-			return nil
+	isLayer3 := n.IPAM.Type != ""
+
+	if isLayer3 {
+		if err := ipam.ExecDel(n.IPAM.Type, args.StdinData); err != nil {
```
bboreham: This line needs to stay before `if args.Netns == "" {` which is on line 496 as of my review.

Reason is that we want to clean up ipam even when the namespace has disappeared.

plugins/main/bridge/bridge.go
```diff
@@ -358,103 +360,109 @@ func cmdAdd(args *skel.CmdArgs) error {
...
+				ipam.ExecDel(n.IPAM.Type, args.StdinData)
+				os.Setenv("CNI_COMMAND", "ADD")
+			}
+		}()
 
-	// Configure the container hardware address and IP address(es)
-	if err := netns.Do(func(_ ns.NetNS) error {
-		contVeth, err := net.InterfaceByName(args.IfName)
+		// Convert whatever the IPAM result was into the current Result type
+		result, err = current.NewResultFromResult(r)
```
squeed: You're setting result.interfaces twice. This variable should be renamed to `ipamResult`.

SchSeba: I first assume the interface is L2 so I create a L2 result 

```go
// Assume L2 interface only
result := &current.Result{CNIVersion: cniVersion}
result.Interfaces = []*current.Interface{brInterface, hostInterface, containerInterface}
```

But if the IPAM is configured then 

```go
// Convert whatever the IPAM result was into the current Result type
result, err = current.NewResultFromResult(r)
```

So I can use one variable only for the return 

```go
return types.PrintResult(result, cniVersion)
```

And don't need to check again if the interface is a L2 to decide what variable to return 

```go
if  isLayer3 {
 return types.PrintResult(ipamResult, cniVersion)
}

return types.PrintResult(result, cniVersion)
```

plugins/main/bridge/bridge.go
SchSeba: Done

plugins/main/bridge/bridge.go
bboreham: @squeed this makes sense - do you still want it changed?

squeed: I guess I'm still a bit uncomfortable with the structure of the code. The fact that the `result` defined in line 364 is *maybe* overwritten strikes me as a source of later bugs. In other words, someone else might manipulate `result`, not understanding that their changes are maybe going to be overwritten.

Could you only define `result` once and update it as needed?

squeed:
@SchSeba thanks! that's exactly what I was thinking about.

One last thing: Could you add a test case, please?

squeed:
Also, you'll need to rebase on master to fix a testcase flake.

dcbw:
@SchSeba could you rebase to fix the test flake and update testcases per @squeed's comment?  Thanks!

SchSeba:
@squeed Can you please check if the tests are good or you want me to check something else?

plugins/main/bridge/README.md
```diff
@@ -28,6 +28,17 @@ If the bridge is missing, the plugin will create one on first use and, if gatewa
 }
 ```
 
+## L2-only configuration
```
matthewdupre: I think "Example L2-only configuration" would be a better title?

squeed:
Looks good to me. Can you squash down to a single commit?

plugins/main/bridge/README.md
SchSeba: Done

SchSeba:
@squeed now it looks better or you want me to remove also the merge commits?

dcbw:
@SchSeba yes, please rebase against master and squash to remove the merge commits.  In general, instead of merging master into your topic branch, please rebase your topic branch on top of master.

SchSeba:
@dcbw  Done.

PR #196, opened 2018-08-28
RFC: Runtime hint: macvlan: Provide hint regarding the runtime in use

amshinde:
All  CNI plugins assume that the runtime uses network namespaces solely for isolation.
Certain OCI runtimes rely on virtual machines for container isolation.
In case of runtimes that make use of VMs, this information can be passed to the plugin.The plugin can then choose an interface setup that is optimal for that runtime.

This change adds support for VM based runtimes for macvlan plugin by passing this information as metadata `args` to the plugin. In case of VM based runtimes, the plugin creates a macvtap interface which can be passed directly to the VM.

 If this approach is acceptable, I would like to add this support for other plugins as well.

Signed-off-by: Archana Shinde <archana.m.shinde@intel.com>

### DIFF
--- a/plugins/main/macvlan/macvlan.go
+++ b/plugins/main/macvlan/macvlan.go
@@ -44,6 +44,11 @@ type NetConf struct {
 	MTU    int    `json:"mtu"`
 }
 
+type ConfigureVMArgs struct {
+	types.CommonArgs
+	ConfigureNetworkForVM types.UnmarshallableBool `json:"configurenetworkforvm,omitempty"`
+}
+
 func init() {
 	// this ensures that main runs only on main thread (thread group leader).
 	// since namespace ops (unshare, setns) are done for a single thread, we
@@ -77,7 +82,7 @@ func modeFromString(s string) (netlink.MacvlanMode, error) {
 	}
 }
 
-func createMacvlan(conf *NetConf, ifName string, netns ns.NetNS) (*current.Interface, error) {
+func createMacvlan(conf *NetConf, ifName string, netns ns.NetNS, configureNetworkForVM bool) (*current.Interface, error) {
 	macvlan := &current.Interface{}
 
 	mode, err := modeFromString(conf.Mode)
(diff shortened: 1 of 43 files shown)


amshinde:
cc @mcastelino

dcbw:
@amshinde it looks like the change really just creates a macvtap instead of a macvlan.  Or am I missing more steps that it takes?

If that's the case, should the config option then just be called "useMacvtap" or something like that?  macvtaps aren't exclusive to VM use-cases, so the config option should probably be generic.

amshinde:
@dcbw The config "useMacvtap" will be very specific to the the macvlan plugin. What I am trying to do here is pass a generic arg to the plugin to let it know that it needs to configure an interface that can be used by a VM. 
I have made changes to macvlan here, but I plan to use the same generic arg for other plugins like ipvlan for them to create VM friendly interfaces as well.

amshinde:
ping @dcbw
Having interfaces that work better with VM based containers will be a useful feature for us, both in terms of performance as well as leveraging features that hypervisor based runtimes provide, while still using the same CNI plugins for namespace based and hypervisor based runtimes.

In the future, I plan to replace the argument introduced here `ConfigureNetworkForVM` with a generic classification string that the Runtime configuration would provide with the introduction of [Runtime Class](https://kubernetes.io/docs/concepts/containers/runtime-class/) in k8s.

plugins/main/macvlan/macvlan.go
```diff
@@ -105,19 +107,45 @@ func createMacvlan(conf *NetConf, ifName string, netns ns.NetNS, configureNetwor
...
+		//
+		// Here the kernel will ensure that there is no race condition
+
+		r := rand.New(rand.NewSource(time.Now().UnixNano()))
+
+		for i := 0; i < linkRetries; i++ {
+			index := hostLinkOffset + (r.Int() & linkRange)
+			mv = &netlink.Macvtap{
+				Macvlan: netlink.Macvlan{
+					LinkAttrs: netlink.LinkAttrs{
```
dcbw: Why do we need to specify an interface index on creation?  We know the ifname so we can look up the index later if we need to, no?

dcbw:
@amshinde what kind of changes were you thinking of for ipvlan?  The ipvtap thing that landed recently in the kernel?

In any case, ConfigureNetworkForVM is a pretty specific thing, but the behavior it would create is not really VM specific.  macvtap and ipvtap etc are useful for other things, not necessarily VMs.  So that's why it makes at least some sense to have them be generic, and not tied to VM-specific configuration options.

Furthermore, the fact that Kubernetes doesn't support multiple networks isn't something the CNI works around, since CNI already easily supports multiple networks (see rkt and CRIO/podman for example). The fact that you cannot say that a VM "container" gets a different network configuration than a namespace container is Kubernetes' problem, not CNIs.  Yes, many of us on the maintainers team work on k8s too and are well aware of this issue, but CNI typically does not work around the limitations of specific runtimes.

That all said, I feel like there has to be a better solution to the issue than making a boolean for "this is VM".  Does that flag work for all the VM-type use-cases from lightweight qemu to full libvirt/VMWare etc?

amshinde:
> @amshinde what kind of changes were you thinking of for ipvlan? The ipvtap thing that landed recently in the kernel?

Yes thats right. The plugin would provide an ipvtap for a VM.

> That all said, I feel like there has to be a better solution to the issue than making a boolean for "this is VM". Does that flag work for all the VM-type use-cases from lightweight qemu to full libvirt/VMWare etc?

@dcbw In case of the k8s, that flag would be provided by the the k8s runtime, since it is aware of the container runtime that will be used for starting the container. I dont see any issues for this not working with libvirt/other hypervisors. 

This issue should provide some context on how the CRI runtime would send this flag:
https://github.com/kubernetes-sigs/cri-o/issues/1733
(That flag would be part of the Runtime Class configuration)

rhatdan:
@amshinde Is this still being worked on?

amshinde:
@rhatdan No, I dont have the cycles to work on this currently.

PR #203, opened 2018-09-18
host-local: add ifname to file tracking IP address used 

### LINKED ISSUE #164: host-local: update to support multiple allocations per network
The CNI spec, in 0.7.0, has changed to allow multiple allocations per network. Host-local doesn't expect this, and needs to be updated.

mccv1r0:
host-local: add ifname to file tracking IP address used 

Obtain ifname from CmdArgs and pass to backend 
In the file tracking the IP address in use, add ifname to the line after ContainerID
Update host-local tests to use ifname along with ContainerID in storage file

Fixes #164

### DIFF
--- a/plugins/ipam/host-local/backend/disk/backend.go
+++ b/plugins/ipam/host-local/backend/disk/backend.go
@@ -100,6 +100,8 @@ func (s *Store) Release(ip net.IP) error {
 // N.B. This function eats errors to be tolerant and
 // release as much as possible
 func (s *Store) ReleaseByID(id string, ifname string) error {
+
+	found := false
 	err := filepath.Walk(s.dataDir, func(path string, info os.FileInfo, err error) error {
 		if err != nil || info.IsDir() {
 			return nil
@@ -112,9 +114,32 @@ func (s *Store) ReleaseByID(id string, ifname string) error {
 			if err := os.Remove(path); err != nil {
 				return nil
 			}
+			found = true
 		}
 		return nil
 	})
+
+	if (!found) && (err == nil) {
+		err = filepath.Walk(s.dataDir, func(path string, info os.FileInfo, err error) error {
+			if err != nil || info.IsDir() {
+				return nil
+			}
+			data, err := ioutil.ReadFile(path)
+			if err != nil {
+				return nil
+			}
+			isCurrent := strings.Contains(string(data), "\n")
+
+			if !isCurrent {
+				if strings.TrimSpace(string(data)) == (strings.TrimSpace(id)) {
+					if err := os.Remove(path); err != nil {
+						return nil
+					}
+				}
+			}
+			return nil
+		})
+	}
 	return err
 }
 
--- a/plugins/ipam/host-local/backend/disk/backend.go
+++ b/plugins/ipam/host-local/backend/disk/backend.go
@@ -26,6 +26,7 @@ import (
 )
 
 const lastIPFilePrefix = "last_reserved_ip."
+const LineBreak = "\r\n"
 
 var defaultDataDir = "/var/lib/cni/networks"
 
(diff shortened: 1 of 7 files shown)


plugins/ipam/host-local/backend/disk/backend.go
```diff
@@ -99,7 +99,9 @@ func (s *Store) Release(ip net.IP) error {
 
 // N.B. This function eats errors to be tolerant and
 // release as much as possible
-func (s *Store) ReleaseByID(id string) error {
+func (s *Store) ReleaseByID(id string, ifname string) error {
+
```
dcbw: spurious newline added...

plugins/ipam/host-local/backend/disk/backend.go
```diff
@@ -65,7 +65,7 @@ func (s *Store) Reserve(id string, ip net.IP, rangeID string) (bool, error) {
 	if err != nil {
 		return false, err
 	}
-	if _, err := f.WriteString(strings.TrimSpace(id)); err != nil {
+	if _, err := f.WriteString(strings.TrimSpace(id) + "\n" + ifname); err != nil {
```
dcbw: @mccv1r0 your concerns WRT Windows are probably valid.  We could use \r\n if you want to be more compatible there.

plugins/ipam/host-local/backend/disk/backend.go
```diff
@@ -108,13 +110,36 @@ func (s *Store) ReleaseByID(id string) error {
...
+		if strings.TrimSpace(string(data)) == (strings.TrimSpace(id) + "\n" + ifname) {
 			if err := os.Remove(path); err != nil {
 				return nil
 			}
+			found = true
 		}
 		return nil
 	})
+
+	if (!found) && (err == nil) {
```
dcbw: You actually don't need the () around !found since it's a bool already.

Also, maybe a comment here explaining that this block is for backwards compatibility when reading files written by a previous version of the plugin that didn't store interface name?

plugins/ipam/host-local/backend/disk/backend.go
```diff
@@ -108,13 +110,36 @@ func (s *Store) ReleaseByID(id string) error {
...
+		err = filepath.Walk(s.dataDir, func(path string, info os.FileInfo, err error) error {
```
dcbw: Another approach would be to pull the directory Walk/data compare/file remove code out to a new function, and then just call it twice; once for id+\n+iface, and if that wasn't found a second time for just 'id'.

plugins/ipam/host-local/host_local_test.go
```diff
@@ -281,7 +281,7 @@ var _ = Describe("host-local Operations", func() {
 		ipFilePath := filepath.Join(tmpDir, "mynet", result.IPs[0].Address.IP.String())
 		contents, err := ioutil.ReadFile(ipFilePath)
 		Expect(err).NotTo(HaveOccurred())
-		Expect(string(contents)).To(Equal("dummy"))
+		Expect(string(contents)).To(Equal("dummy" + "\n" + ifname))
 
 		// Release the IP
 		err = testutils.CmdDelWithArgs(args, func() error {
```
dcbw: I'd add two more tests to this file that (a) allocates two IPs for the same container on different interfaces, then deletes one and makes sure the right file got deleted, and (b) writes an old-style (eg containerid only) IP file to the data dir and runs DEL and makes sure the old-style file is deleted.

plugins/ipam/host-local/backend/disk/backend.go
rosenhouse: Please ensure you've run `go fmt`.  It should remove those unnecessary parens (and help prevent other stylistic inconsistencies).

plugins/ipam/host-local/backend/disk/backend.go
squeed: Nobody seemed to like my suggestion of using ASCII BEL :-)

dcbw:
Testcase failures are unrelated and will be fixed by https://github.com/containernetworking/plugins/pull/207

PR #207, opened 2018-09-27
Fix vendor regression in dhcp4server

### LINKED ISSUE #206: dhcp4server vendor regression
plugins/vendor/github.com/d2g/dhcp4server lost a fix in leasepool/memorypool/memorypool.go that causes dhcp_test.go to fail.

mccv1r0:
vendor: update dhcp4server 

Bring plugins/vendor/github.com/d2g/dhcp4server up to date.

Fixes #206

### DIFF
--- a/plugins/ipam/dhcp/dhcp_test.go
+++ b/plugins/ipam/dhcp/dhcp_test.go
@@ -107,13 +107,8 @@ func dhcpServerStart(netns ns.NetNS, leaseIP, serverIP net.IP, stopCh <-chan boo
 const (
 	hostVethName string = "dhcp0"
 	contVethName string = "eth0"
-	pidfilePath  string = "/var/run/cni/dhcp-client.pid"
 	cniDirPrefix string = "/var/run/cni"
 )
-
-var socketPath string
-var tmpDir string
-var err error
 
 var _ = BeforeSuite(func() {
 	err := os.MkdirAll(cniDirPrefix, 0700)
@@ -125,6 +120,9 @@ var _ = Describe("DHCP Operations", func() {
 	var dhcpServerStopCh chan bool
 	var dhcpServerDone *sync.WaitGroup
 	var clientCmd *exec.Cmd
+	var socketPath string
+	var tmpDir string
+	var err error
 
 	BeforeEach(func() {
 		dhcpServerStopCh = make(chan bool)
--- a/plugins/ipam/dhcp/dhcp_test.go
+++ b/plugins/ipam/dhcp/dhcp_test.go
@@ -41,7 +41,7 @@ import (
 )
 
 func getTmpDir() (string, error) {
-	tmpDir, err := ioutil.TempDir(tmpPrefix, "dhcp")
+	tmpDir, err := ioutil.TempDir(cniDirPrefix, "dhcp")
 	if err == nil {
 		tmpDir = filepath.ToSlash(tmpDir)
 	}
@@ -108,7 +108,7 @@ const (
 	hostVethName string = "dhcp0"
 	contVethName string = "eth0"
 	pidfilePath  string = "/var/run/cni/dhcp-client.pid"
-	tmpPrefix           = "/run/cni"
+	cniDirPrefix string = "/var/run/cni"
 )
 
 var socketPath string
@@ -116,15 +116,8 @@ var tmpDir string
 var err error
 
 var _ = BeforeSuite(func() {
-	os.Remove(socketPath)
-	os.Remove(pidfilePath)
-	err := os.MkdirAll(tmpPrefix, 0700)
+	err := os.MkdirAll(cniDirPrefix, 0700)
 	Expect(err).NotTo(HaveOccurred())
-})
-
-var _ = AfterSuite(func() {
-	os.Remove(socketPath)
-	os.Remove(pidfilePath)
 })
 
 var _ = Describe("DHCP Operations", func() {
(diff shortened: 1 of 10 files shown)


squeed:
Hmm, the test failures don't actually look like flakes - they're some kind of DHCP errors.

mccv1r0:
The test fails because socketPath (/run/cni/dhcp.sock) isn't found.

```
• Failure in Spec Setup (BeforeEach) [15.021 seconds]
DHCP Operations [BeforeEach] configures and deconfigures a link with ADD/DEL 
/home/travis/gopath/src/github.com/containernetworking/plugins/gopath/src/github.com/containernetworking/plugins/plugins/ipam/dhcp/dhcp_test.go:214
  Timed out after 15.000s.
  Expected
      <bool>: false
  to be true
```
```
		// Wait up to 15 seconds for the client socket
		Eventually(func() bool {
			_, err := os.Stat(socketPath)
			return err == nil
		}, time.Second*15, time.Second/4).Should(BeTrue())
```
Locally, I've had it fail right in the middle of the test.  e..g cmdAdd found /run/cni/dhcp.sock and shortly after cmdDel couldn't find /run/cni/dhcp.sock.

Each run results in something different, all related to /run/cni/dhcp.sock not being there.

Commenting out os.Remove(socketPath)...

```
var _ = AfterSuite(func() {
	//os.Remove(socketPath)
	os.Remove(pidfilePath)
```

... and in AfterEach() results in test always passing

dcbw:
The DHCP failures are pretty odd, we're tracking them down.  I can't get them to fail locally myself, but we'll figure it out.

rosenhouse:
Hmm, is it really necessary to disable all test parallelization just to fix this problem?  Could we instead make `socketPath` be configurable at runtime (args or env var) so that each test could get a unique socket?

dcbw:
> Hmm, is it really necessary to disable all test parallelization just to fix this problem? Could we instead make `socketPath` be configurable at runtime (args or env var) so that each test could get a unique socket?

@rosenhouse we thought about that too.  It would require pretty invasive changes since we'd need to plumb the socket path through both the server-side code (with a CLI flag) and the CNI JSON config as well.  All to make the testcase work; it wouldn't really be used in normal operation.  But yes, that is a potential fix.

What was really odd to Michael and I was that Ginkgo parallelized tests *in the same _test.go file*, as opposed to just files in the same directory or different packages.  Anyway, that's how it is I guess...

plugins/ipam/dhcp/dhcp_test.go
```diff
@@ -99,6 +110,10 @@ const (
 	pidfilePath  string = "/var/run/cni/dhcp-client.pid"
 )
 
+var socketPath string
+var tmpDir string
+var err error
+
 var _ = BeforeSuite(func() {
 	os.Remove(socketPath)
 	os.Remove(pidfilePath)
```
dcbw: I don't think we need the Remove() for socketPath and pidfilePath anymore, since we're calling RemoveAll on the directory they are in.  Same for AfterSuite()

plugins/ipam/dhcp/dhcp_test.go
```diff
@@ -107,6 +122,7 @@ var _ = BeforeSuite(func() {
 var _ = AfterSuite(func() {
 	os.Remove(socketPath)
 	os.Remove(pidfilePath)
+	defer os.RemoveAll(tmpDir)
```
dcbw: And actually, we don't even need this right?  We should put the RemoveAll(tmpDir) inside the AfterEach(), since it'll be unique to each testcase.  This RemoveAll() call in AfterSuite would only remove the last testcase tmpdir, which should already have been done by AfterEach()

plugins/ipam/dhcp/dhcp_test.go
```diff
@@ -118,6 +134,10 @@ var _ = Describe("DHCP Operations", func() {
 	BeforeEach(func() {
 		dhcpServerStopCh = make(chan bool)
 
+		tmpDir, err = getTmpDir()
```
dcbw: We should balance this in AfterEach() with a os.RemoveAll(tmpDir) to clean things up for each test.

mccv1r0: I fixed this in last push which ensured that /run/cni existed.    (I was able to reproduce test "hang" when /run didn't contain cni directory)

plugins/ipam/dhcp/dhcp_test.go
mccv1r0: pidfilePath is (currently) in /var/run/cni, not /run/cni.  Should it be moved to /run/cni?

plugins/ipam/dhcp/dhcp_test.go
```diff
@@ -97,16 +108,16 @@ const (
 	hostVethName string = "dhcp0"
 	contVethName string = "eth0"
 	pidfilePath  string = "/var/run/cni/dhcp-client.pid"
```
dcbw: I don't think we need pidfilePath as a global constant anymore either.

mccv1r0: Agreed.  Removed.

plugins/ipam/dhcp/dhcp_test.go
```diff
@@ -97,16 +108,16 @@ const (
...
+	cniDirPrefix string = "/var/run/cni"
 )
 
-var _ = BeforeSuite(func() {
-	os.Remove(socketPath)
-	os.Remove(pidfilePath)
-})
+var socketPath string
+var tmpDir string
+var err error
```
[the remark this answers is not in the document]
mccv1r0: Moved to the var() section of Describe().  Tested with both -p and without.

plugins/ipam/dhcp/main.go
```diff
@@ -28,18 +29,24 @@ import (
 	"github.com/containernetworking/cni/pkg/version"
 )
 
-const socketPath = "/run/cni/dhcp.sock"
+const defaultSocketPath = "/run/cni/defaultdhcp.sock"
```
squeed: We should keep the same default path.

PR #211, opened 2018-10-10
portmap: don't panic if listing a chain returns fewer lines than expe…

plwhite:
…cted


plwhite:
Embarrassing - opened PR by mistake, sorry.

PR #212, opened 2018-10-10
Create new Source Based Routing plugin

plwhite:
This creates a new plugin (sbr) which sets up source based routing, for use
as a chained plugin for multi-network environments.

*Note to maintainers - this is now updated with tests etc. so should be ready to merge. I think I have addressed all comments so far.*

### DIFF
--- a/plugins/meta/sbr/README.md
+++ b/plugins/meta/sbr/README.md
@@ -1,0 +1,127 @@
+# Source based routing plugin
+
+## Introduction
+
+This plugin performs Source Based Routing (SBR). The most common and standard way to
+perform routing is to base it purely on the destination. However, in some
+applications which are using network separation for traffic management and
+security, there is no way to tell *a priori* which interface should be used,
+but the application is capable of making the decision.
+
+As an example, a Telco application might have two networks, a management
+network and a SIP (telephony) network for traffic, with rules which state that:
+
+- SIP traffic (only) must be routed over the SIP network;
+
+- all other traffic (but no SIP traffic) must be routed over the management
+  network.
+
+There is no way of configuring this based on destination IP, since there is no
+way of telling whether a destination IP on the internet is (say) an address
+used for downloading updated software packages or a remote SIP endpoint.
+
+Hence Source Based Routing is used.
+
+- The application explicitly listens on the correct interface for incoming
+  traffic.
+
+- When the application wishes to send to an address via the SIP network, it
+  explicitly binds to the IP of the device on that network.
+
+- Routes for the SIP interface are configured in a separate routing table, and
+  a rule is configured to use that table based on the source IP address.
+
+Note that in most cases there is a management device (the first one) and that
+the SBR plugin is called once for each device after the first, leaving the
+default routing table applied to the management device. However, this not
+mandatory, and source based routing may be configured on any or all of the
+devices as appropriate.
+
+## Usage
+
+This plugin runs as a chained plugin, and requires the following information
+passed in from the previous plugin (which has just set up the network device):
+
+- What is the network interface in question?
+
+- What is the IP address (or addresses) of that network interface?
+
+- What is the default gateway for that interface (if any)?
+
+It then reads all routes to the network interface in use.
+
+Here is an example of what the plugin would do. (The `ip` based commands are
+implemented in go, but easier to describe via the command line.) Suppose that
+it reads that:
+
+- The interface is `net1`.
+
+- The IP address on that interface is `192.168.1.209`.
+
+- The default gateway on that interface is `192.168.1.1`.
+
+- There is one route configured on that network interface, which is
+  `192.168.1.0/24`.
+
+Then the actions it takes are the following.
+
+- It creates a new routing table, and sets a rule to use it for the IP address in question.
+
+        ip rule add from 192.168.1.209/32 table 100
+
+- It adds a route to the default gateway for the relevant table.
+
+        ip route add default via 192.168.1.1 dev net1 table 100
+
+- It moves every existing route on the device to the new table.
+
+        ip route del 192.168.1.0/24 dev net1 src 192.168.1.209
+        ip route add 192.168.1.0/24 dev net1 src 192.168.1.209 table 100
+
+On deletion it:
+
+- deletes the rule (`ip rule del from 192.168.1.209/32 table 100`), which it
+  finds by deleting rules relating to IPs on the device which is about to be
+  deleted.
+
+- does nothing with routes (since the kernel automatically removes routes when
+  the device with which they are associated is deleted).
+
+## Future enhancements and known limitations
+
+The following are possible future enhancements.
+
+- The table number is currently selected by starting at 100, then incrementing
+  the value until an unused table number is found. It might be nice to have an
+  option to pass the table number as an input.
+
+- There is no log severity, and there is no logging to file (pending changes to
+  CNI logging generally).
+
+- This plugin sets up Source Based Routing, as described above. In future,
+  there may be a need for a VRF plugin (that uses
+  [VRF routing](https://www.kernel.org/doc/Documentation/networking/vrf.txt)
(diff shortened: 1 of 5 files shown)


plugins/meta/sbr/README.md
```diff
@@ -0,0 +1,128 @@
...
+- It adds a route to the default gateway for the relevant table.
+
+        ip route add default via 192.168.1.1 table 100
+
+- It moves every existing route on the device to the new table.
+
+        ip route del 192.168.1.0/24 dev net1 src 192.168.1.209
+        ip route add 192.168.1.0/24 dev net1 src 192.168.1.209 table 100
+
+On deletion it does nothing at all (since it assumes the pod is about to be
```
dcbw: With Kubernetes as a runtime you can assume this, but CNI isn't only used with Kubernetes and so I don't think we can assume this in the general case.  Would it be hard to add cleanup on DEL?

plwhite: Good point. I think it's probably fairly easy to tidy up the table and delete the rule that sends to it. I'll tweak the PR in due course, and update the docs to reflect.

plugins/meta/sbr/main.go
```diff
@@ -0,0 +1,291 @@
...
+		if found {
+			table++
+		} else {
+			break
+		}
+	}
+
+	log.Printf("First unreferenced table: %d", table)
+
+    link, err := netlink.LinkByName(iface)
```
squeed: whoops, please run `gofmt`.

plugins/meta/sbr/main.go
```diff
@@ -0,0 +1,291 @@
...
+		if intIdx >= 0 && intIdx < len(conf.PrevResult.Interfaces) && conf.PrevResult.Interfaces[intIdx].Name != args.IfName {
+			log.Printf("Incorrect interface for IP address %s", ip.Address.IP)
+			continue
+		}
+
+		log.Printf("Found IP address %s", ip.Address.IP.String())
+		ips = append(ips, ip)
+	}
+
+	nsname := os.Getenv("CNI_NETNS")
```
dcbw: args.NetNS

plugins/meta/sbr/main.go
```diff
@@ -0,0 +1,291 @@
...
+	log.Printf("Network namespace to use: %s", nsname)
+	namespace, err := ns.GetNS(nsname)
```
dcbw: Can simplify this with ns.WithNetNSPath() rather than getting the namespace object and then Do() later.

plwhite:
I think this is now a decent candidate for a merge. I should squash commits before doing so, but deliberately haven't done that yet so you can see what's new.

plugins/meta/sbr/README.md
```diff
@@ -0,0 +1,131 @@
...
+~~~json
+{
+  "name": "flannel-sbr",
+  "cniVersion": "0.3.0",
+  "plugins":
+    [
+        { "type": "flannel" },
+        {
+          "type": "sbr",
+          "logfile": "/var/log/sbr.log"
```
squeed: Generally, plugins just log to stderr.

plugins/meta/sbr/main.go
```diff
@@ -0,0 +1,360 @@
...
+
+	routes, err := netlink.RouteList(nil, netlink.FAMILY_ALL)
+	if err != nil {
+		return fmt.Errorf("Failed to list all routes: %v", err)
+	}
+
+	// Pick a table ID to use. We pick the first table ID from firstTableID
+	// on that has no existing rules mapping to it and no existing routes in
+	// it.
+	table := firstTableID
```
squeed: If two instances of this plugin run at the same time, there's a race condition. Do we care? If so, you should probably add a (sigh) lockfile.

squeed: More specifically, the spec does say that multiple ADDs can happen at the same time.

plugins/meta/sbr/README.md
plwhite: True, but I'd like to appeal on this one. By default the plugin just logs to stderr, but can be configured to log to file. I actually found the option of setting that debug option was very useful when testing.

Hence I can remove the logging to file if you want, but it felt like a good thing to have available when I needed it. What do you think?

plugins/meta/sbr/main.go
plwhite: Good spot, and depressing that I missed this. It's because I am too Kubernetes focussed, where this cannot happen in practice.

I'll add a file lock, probably using the code from the backend plugin.

plugins/meta/sbr/README.md
dcbw: @plwhite we don't have any general objection to logging elsewhere than stderr, but we believe plugins should always log to stderr. Any other logging should be additional to stderr. But we also want to figure out a "standard" way to tell plugins what level to log at and what the additional method to use, since we don't think that should be different for every plugin.

(eg, you would just want to tell Kubernetes or CRIO or Mesos to "turn on CNI debugging" and it would work for all plugins in a chain)

dcbw: @plwhite so for now, if it's OK, we'd like to remove the logging and add it back later when we have a better plan for logging in general.

plwhite: Having thought this over, I agree that every plugin logging in its own way is insane, and strictly it's way more sensible that K8s should be trapping stderr and saving it somewhere.
So fair enough, I have removed it in my latest commit.

plwhite:
@dcbw Do you reckon this is ready for merge? It obviously needs commits squashed down to one, but I won't do that until we are ready.

dcbw:
@plwhite could you 'git rm plugins/meta/sbr/cnitest-c18c58aa-d341-2759-1fcd-2bbc1d1f2d01' and update? looks like that leftover test output snuck in or something.

plugins/meta/sbr/README.md
```diff
@@ -0,0 +1,126 @@
...
+
+## Future enhancements and known limitations
+
+The following are possible future enhancements.
+
+- The table number is currently selected by starting at 100, then incrementing
+  the value until an unused table number is found. It might be nice to have an
+  option to pass the table number as an input.
+
+- The logging is entirely on or entirely off - there is no log severity.
```
matthewdupre: I can't see a way to turn the logging off?  Maybe I'm missing something.  Would be nice to be able to change.

plugins/meta/sbr/main.go
```diff
@@ -0,0 +1,378 @@
...
+	PrevResult    *current.Result         `json:"-"`
+
+	// Add plugin-specific flags here
+}
+
+// Wrapper that does a lock before and unlock after operations to serialise
+// this plugin.
+func withLockAndNetNS(nspath string, toRun func(_ ns.NetNS) error) error {
+	// We lock on the network namespace to ensure that no other instance
+	// clashes with this one. Note that
```
matthewdupre: Unfinished sentence?

plugins/meta/sbr/README.md
dcbw: Or if there isn't a way to turn it off, I think we're fine with fixing the README file and merging, right Matt?

plwhite: There is no way to turn it off. There used to be logging to file, and when I removed that capability, I missed changing the README. Will fix this now.

plugins/meta/sbr/main.go
plwhite: Good spot, sorry. Will fix.

plwhite:
I think I fixed everything required - happy to squash all the commits ready for merge now if you reckon it's ready to go?

squeed:
Seems good to me. If later users file an issue asking for logging to be a bit less chatty, I would agree with them - but not enough to stop this PR now.

LGTM pending squash.

plwhite:
I've squashed commits, ready for merge.

PR #217, opened 2018-10-15
Add container ifName to the dhcp clientID, making the clientID value

### LINKED ISSUE #214: dhcp: update to support multiple allocations per network
The CNI spec, in 0.7.0, has changed to allow multiple allocations per network.  The dhcp plugin doesn't expect this, and needs to be updated.     aka the dhcp version of  https://github.com/containernetworking/plugins/issues/164(url)

mccv1r0:
now equal to: ContainerID / Host Network / ifname inside container

Use clientID as a key for the lease

Fixes #214


plugins/ipam/dhcp/daemon.go
```diff
@@ -91,40 +97,43 @@ func (d *DHCP) Release(args *skel.CmdArgs, reply *struct{}) error {
...
-	l, ok := d.leases[contID+netName]
+	l, ok := d.leases[clientID]
 	if !ok {
 		return nil
 	}
 	return l
 }
 
-func (d *DHCP) setLease(contID, netName string, l *DHCPLease) {
+//func (d *DHCP) setLease(contID, netName string, ifName string, l *DHCPLease) {
```
s1061123: how about to remove it?

plugins/ipam/dhcp/daemon.go
```diff
@@ -50,6 +50,12 @@ func newDHCP() *DHCP {
 	}
 }
 
+func generateClientID(containerID string, netName string, ifName string) string {
+	clientID := containerID + "/" + netName + "/" + ifName
```
s1061123: suppose we just make it one line as 
`return containerID + "/" + netName + "/" + ifName`

plugins/ipam/dhcp/daemon.go
mccv1r0: Missed that one, thanks

plugins/ipam/dhcp/daemon.go
mccv1r0: done

dcbw:
On further inspection we figured out that the DHCP client code doesn't actually pass the Client ID in the packet, so even though we've fixed the internal code we have more to do.  Related: https://github.com/d2g/dhcp4client/pull/24

dcbw:
@mccv1r0 thanks for the update; one small nit otherwise LGTM

plugins/ipam/dhcp/dhcp_test.go
```diff
@@ -331,3 +339,267 @@ var _ = Describe("DHCP Operations", func() {
...
+	var dhcpServerDone *sync.WaitGroup
+	var clientCmd *exec.Cmd
+	var socketPath string
+	var tmpDir string
+	var err error
+
+	var br *netlink.Bridge
+
+	BeforeEach(func() {
+		dhcpServerStopCh = make(chan bool)
```
bboreham: The next 100+ lines look very similar to lines 49 onwards in dhcp2_test.go - if they are different please add some comments indicating what is going on, otherwise pull out the common code to a shared setup function.

plugins/ipam/dhcp/client.go
```diff
@@ -0,0 +1,243 @@
...
+	}
+
+	declinePacket.PadToMinSize()
+
+	return declinePacket, c.SendPacket(declinePacket)
+}
+
+/*
+//Create Discover Packet
+func DhcpDiscoverPacket(c *dhcp4client.Client, options dhcp4.Options) dhcp4.Packet {
```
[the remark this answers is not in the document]
mccv1r0: Missed those, thanks.

plugins/ipam/dhcp/dhcp_test.go
mccv1r0: I created a shared setup function as suggested

bboreham:
I was wondering what happens if someone was running the previous version, and they upgrade to this version: will the IDs sent to DHCP change?  But @squeed worked out that the daemon doesn't persist this information so all the leases are forgotten on every restart so this isn't a new problem.

Issue #178, opened 2018-08-01
macvlan: invalid mode in README


onlyjob:
macvlan's [README](https://github.com/containernetworking/plugins/blame/master/plugins/main/macvlan/README.md#L27) contains invalid `passthrough` mode that yields the following error:

    unknown macvlan mode: "passthrough"

The correct name for this mode is `passthru`.

Please update REAME by replacing `passthrough` with `passthru`.

onlyjob:
Fixed by 5fc8209, thanks.

PR #218, opened 2018-10-17
fix the typo of macvlan and also modify documents to meet the current plugins.

hwchiu:
Refer to issue #178


hwchiu:
Should I need to squash to one commit ?

Issue #184, opened 2018-08-07
host-device: set link up after moving host device to pod namespace


zshi-redhat:
Currently host-device plugin only moves the host-device to pod namespace.

But sometimes, when moving the host device to pod, user wants to use that device as an interface for running real network traffic, it would be good to set the device link up in the pod namespace and perhaps even further to add the ability for host-device to run ipam configuration.

one possible use case: pod running in a virtual machine wants to use one of the virtual interfaces in the VM(in this case, the virtual interfaces can be seen as host-device for the pod).

dpward:
While IPAM can now be used with the `host-device` plugin, this doesn't help if the interface is being used at layer 2.

This issue seems unique to the `host-device` plugin. The [interface-creating](https://www.cni.dev/plugins/current/#main-interface-creating) plugins will bring the new interface up automatically. When an interface gets moved into another namespace, though, the kernel automatically brings it down.

There are a few approaches here. Which would be best? (It might actually be more than one.)
- [ ] Change the `host-device` plugin to bring the interface up by default.
- [ ] Add an interface attribute to the `host-device` plugin to set the interface state (up or down).
- [ ] Add an interface attribute to the `tuning` plugin to set the interface state (up or down).

zshi-redhat:
> While IPAM can now be used with the `host-device` plugin, this doesn't help if the interface is being used at layer 2.
> 
We probably want to have a similar code piece like [sriov-cni](https://github.com/k8snetworkplumbingwg/sriov-cni/blob/master/pkg/sriov/sriov.go#L195-L198) to explicitly set the interface link up.

jayunit100:
generally curious about what is meant by "use that device as an interface for running real network traffic" :)

PR #220, opened 2018-10-21
Support the IPAM in the host-device

hwchiu:
Refer to the issue #184 
### What's the change
- Support the IPAM options in the host-device
- Skip the IPAM function if it doesn't exist in the config (compatibility)
- Change the variable style from ALL_CAPS to CamelCase

### Testing
- Also add the test case to test the IPAM with static module
- Check the IP address in the container should be same as what we set in the config.

### DIFF
--- a/plugins/main/host-device/host-device.go
+++ b/plugins/main/host-device/host-device.go
@@ -17,6 +17,7 @@ package main
 import (
 	"bytes"
 	"encoding/json"
+	"errors"
 	"fmt"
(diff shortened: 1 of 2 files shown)


squeed:
This looks.. great! Thanks for a nice PR.

Could you please squash down to one commit?

hwchiu:
> This looks.. great! Thanks for a nice PR.
> 
> Could you please squash down to one commit?

Got it

PR #231, opened 2018-11-18
Added vlan tag to the bridge cni plugin.

SchSeba:
Allow the user to assign VLAN tag.

Updated netlink library to introduce the VlanFiltering variable into the bridge struct


squeed:
Neat. Linux bridges never cease to surprise me :-). Some minor requests:
- Can you split this in to two commits, one for the actual code and one for the `vendor/` bump?
- Can you add some documentation as to what this does? Especially, be clear that this sets the vlan on the *container interface* (I didn't immediately understand this)

And some larger questions:
- Is there a use-case for setting a vlan and *not* enabling `vlanFiltering`? Otherwise, I'd rather just delete that knob.
- Should we handle the case where an existing bridge doesn't have `vlanFiltering` enabled? Is that a setting that can be changed on-the-fly? `ensureBridge` should be as idempotent as reasonably possible.

squeed:
So, we talked about this at the CNI maintainers meeting, and this came up: If a container is started in a vlan, how does it get connectivity? Presumably there's no equivalent vlan-100 "uplink" interface on the "host" side of the bridge.

What are you envisioning? Do you have a separate step that creates the "br0.100" interface? Or do you expect non-default vlans to be private / isolated?

It's generally expected, but by no means required, that a CNI plugin results in a container being "online." Of course, these are loose definitions, but they are helpful to guide decisions.

Should this plugin also create the "br0.100" interface if configured with a vlan?

phoracek:
@squeed forgive me, I don't know much about setting of VLAN directly on a bridge, but my understanding was this:

If I connect two containers on the same host to bridge+vlan, they would be able to reach each other.

If I create a bridges on two nodes and connect them via a trunk interface. Then containers connected to this bridge+vlan can reach each other node-to-node.

Is that right or is more needed?

dankenigsberg:
> So, we talked about this at the CNI maintainers meeting, and this came up: If a container is started in a vlan, how does it get connectivity? Presumably there's no equivalent vlan-100 "uplink" interface on the "host" side of the bridge (i.e. "br0.100").

If properly set, the bridge has one trunk uplink (e.g. eth0) and multiple ports, one per Pod. Traffic arriving to eth0 tagged with  vlanX is repeated (under usual bridge rules) to all ports belonging to this vlan, but with the tag stripped. The reverse happen to egress traffic.

So if I get what you're asking, the bridge port itself plays the role of  br0.100 vlan device. They tag egress and strip ingress in the same way.

(/me slightly rephrasing @phoracek)

dankenigsberg:
I don't know how it usually works on github, but @SchSeba maybe you can start you commit message by explaining what bridge vlan filtering is (include https://developers.redhat.com/blog/2017/09/14/vlan-filter-support-on-bridge/ ) and why it is useful here.

SchSeba:
@squeed 
> Should this plugin also create the "br0.100" interface if configured with a vlan?

The code to allow the vlan to be tagged on the bridge interface. there is not need to create a br0.100 interface.

``` go
if n.Vlan != 0 {
		err = netlink.BridgeVlanAdd(br, uint16(n.Vlan), false, true, true, false)
		if err != nil {
			return nil, nil, fmt.Errorf("failed to setup vlan tag on the bridge interface %q: %v", br.Name, err)
		}
	}
```

SchSeba:
@squeed there is any update on this PR?

squeed:
Then I still don't understand it. Are you manually trunking in another physical interface? If I want to send a packet from the host to a container on vlan 100, how do I do it?

plugins/main/bridge/bridge.go
```diff
@@ -293,12 +302,23 @@ func calcGatewayIP(ipn *net.IPNet) net.IP {
...
+	}
 	// create bridge if necessary
-	br, err := ensureBridge(n.BrName, n.MTU, n.PromiscMode)
+	br, err := ensureBridge(n.BrName, n.MTU, n.PromiscMode, vlanFiltering)
 	if err != nil {
 		return nil, nil, fmt.Errorf("failed to create bridge %q: %v", n.BrName, err)
 	}
 
+	if n.Vlan != 0 {
+		err = netlink.BridgeVlanAdd(br, uint16(n.Vlan), false, true, true, false)
```
rosenhouse: I don't see this vlan filter entry being removed on `DEL`.  Are there any downsides to leaving it in place even after all veths tagged with the vid have been removed?

plugins/main/bridge/bridge_test.go
```diff
@@ -171,6 +180,7 @@ var counter uint
 // arguments for a test case.
 func (tc testCase) createCmdArgs(targetNS ns.NetNS, dataDir string) *skel.CmdArgs {
 	conf := tc.netConfJSON(dataDir)
+	fmt.Println(conf)
```
rosenhouse: did you intend to leave this in the `testCase`?

plugins/main/bridge/bridge.go
SchSeba: Yeah I don't remove it from the bridge I can't find any good reason to remove it just link we are not removing the bridge if there is no veths connected to it.

plugins/main/bridge/bridge_test.go
SchSeba: Nope my bad sorry

SchSeba:
Hi @squeed there is something that prevent this PR from be merged?

plugins/main/bridge/README.md
```diff
@@ -52,3 +52,8 @@ If the bridge is missing, the plugin will create one on first use and, if gatewa
 * `hairpinMode` (boolean, optional): set hairpin mode for interfaces on the bridge. Defaults to false.
 * `ipam` (dictionary, required): IPAM configuration to be used for this network. For L2-only network, create empty dictionary.
 * `promiscMode` (boolean, optional): set promiscuous mode on the bridge. Defaults to false.
+* `vlan` (int, optional): assign VLAN tag. Defaults to none.
+
+*Note:* The VLAN parameter config the VLAN tag on the host end of the veth and also enable the vlan_filtering feature on the bridge interface.
```
dcbw: "config" -> "configures"
"enable" -> "enables"

plugins/main/bridge/README.md
```diff
@@ -52,3 +52,8 @@ If the bridge is missing, the plugin will create one on first use and, if gatewa
...
+
+*Note:* To configure up link for L2 network you need to allow the vlan on the up link interface by using the follow command ``` bridge vlan add vid VLAN_ID dev DEV```
```
dcbw: "up link" -> "uplink"
"follow" -> "following"

And I assume DEV here is like eth0 or whatever the hosts physical NIC would usually be?

dcbw:
> So if the want so send a packet you need to allow the vlan on the bridge interface and configure ip address in the same subnet on that interface.

This isn't quite clear enough to me... what would be required on the host side to send a packet to a container on a given VLAN? Would you need to create another veth and attach it to the bridge using the same VLAN ID as the container?

plugins/main/bridge/README.md
dankenigsberg: > And I assume DEV here is like eth0 or whatever the hosts physical NIC would usually be?

Correct. A better text would be

To configure up link for L2 network you need to allow the vlan on the up link interface `DEV` by using the follow command

or maybe use the blog example of `bridge vlan add dev bond0 vid 2 master`

squeed:
I tested this with two configurations, same bridge:

```json
{
        "cniVersion": "0.3.1",
        "name": "mybridge",
        "type": "bridge",
        "ipMasq": true,
        "ipam": {
                "type": "host-local",
                "subnet": "10.42.0.0/24"
        },
        "isDefaultGateway": true,
        "hairpinMode": true
}
```

and
```json
{
        "cniVersion": "0.3.1",
        "name": "myvlanbridge",
        "type": "bridge",
        "ipMasq": true,
        "ipam": {
                "type": "host-local",
                "subnet": "10.52.0.0/24"
        },
        "isDefaultGateway": true,
        "hairpinMode": true,
        "vlan": 99
}
```

and got the following error:
```console
$ sudo CNI_PATH=./bin ../cni/bin/cnitool add myvlanbridge /var/run/netns/vlan100 
failed to set bridge addr: "cni0" already has an IP address different from 10.52.0.1/24
```

squeed:
Just to follow up on my previous comment: I would expect setting the `vlan` property on the bridge to work, even when users expect layer3 connectivity.

If I understand your use-case correctly, you only really care about layer2 connectivity, so that's not an issue. However, I would really like the plugins to continue to meet user expectations, and it's not at all clear to me as an end-user why adding vlan tag would require me to drop layer3.

So I'd like you to come up with a way for containers to have both a vlan tag and layer3 connectivity. If this means you have to create `bridgename.100` on the host or not, I leave up to you.

SchSeba:
Hi @squeed sorry it took me a lot of time to replay.

I update this PR and change a little bit the way we configure a L3 vlan interface.

For a l2:
Configure the vlan tag on the veth host side.

For a L3:
Configure the vlan tag on the veth host side.
Create a veth peer on the host side connect one side to the bridge and configure the vlan tag, The other side is configured with the gw ip.

plugins/main/bridge/bridge.go
```diff
@@ -188,6 +188,48 @@ func ensureBridgeAddr(br *netlink.Bridge, family int, ipn *net.IPNet, forceAddre
 	return nil
 }
 
+func ensureVlanInterfaceAddr(br *netlink.Bridge, vlan, family int, ipn *net.IPNet, forceAddress bool) error {
```
squeed: This function seems to be a full copy of `ensureBridgeAddr()` -- can share the logic? Maybe just move `ensureVlanInterface` around line 547?

plugins/main/bridge/bridge.go
```diff
@@ -198,6 +240,16 @@ func deleteBridgeAddr(br *netlink.Bridge, ipn *net.IPNet) error {
 	return nil
 }
 
+func deleteVlanAddr(vlan netlink.Link, ipn *net.IPNet) error {
```
squeed: Likewise, this just seems like a duplicate -- can we get rid of it?

squeed:
Thanks for the update - I'll try it out soon.

Some code seems copied-and-pasted - could you please de-duplicate that, and instead refactor the existing function as necessary?

Question: does it make sense / is it easy to avoid creating the host-side veth? As in, could you trunk the additional vlans to the host, and add a separate vlan interface? If creating the host-side veth makes more sense, then that's fine.

SchSeba:
@squeed Thanks for the comments I will try to fix the code related to your comment.

> Question: does it make sense / is it easy to avoid creating the host-side veth? As in, could you trunk the additional vlans to the host, and add a separate vlan interface? If creating the host-side veth makes more sense, then that's fine.

This is why it took me so much time to update the PR, I was investigating the vlan and there is no way to configure the vlan filter on a vlan interface and if we configure the vlan filter on the bridge interface we can have problems because all the vlan interfaces will get traffic from all the pod(unrelated to the vlan) .

So I prefer to change the logic to create a veth per vlan.

SchSeba:
@squeed also another problem with the vlan interface approach that I find trying to solve this issue is that the bridge interface must have ip or the traffic will not get into the vlan interface.

example:
```
br0 -> no ip
br0.100 -> 10.0.0.1

pod -> 10.0.0.2 (vlan 100) 
```
the traffic will not hit the bridge vlan interface until the br0 interface as any ip address.

This can be a problem if the user only want a vlan bridge.

squeed:
Ah, I see. Thanks for the explanation.

SchSeba:
@squeed this travis fail looks unrelated to my PR

https://travis-ci.org/containernetworking/plugins/jobs/489972970

squeed:
Agreed, it looks like windows CI is broken.

SchSeba:
Hi @squeed can you try to rerun the tests?

jellonek:
@squeed ping ;)

dankenigsberg:
@jellonek may I ask if you have tested it yourself? Do you find this useful, too?

jellonek:
No, I didn't tested that manually. I'm waiting for results from CI, and yes - I see that as something really useful.

squeed:
With no existing bridges, I ran the plugin with `vlan=99`.

The result doesn't match the interfaces created.

result:
```json
{
    "cniVersion": "0.3.1",
    "interfaces": [
        {
            "name": "cni0",
            "mac": "12:ce:02:f1:c6:ee"
        },
        {
            "name": "vethaceb2b79",
            "mac": "12:ce:02:f1:c6:ee"
        },
        {
            "name": "eth0",
            "mac": "9a:74:f7:69:f4:3c",
            "sandbox": "/var/run/netns/testing1"
        }
    ],
    "ips": [
        {
            "version": "4",
            "interface": 2,
            "address": "10.52.0.4/24",
            "gateway": "10.52.0.1"
        }
    ],
    "routes": [
        {
            "dst": "0.0.0.0/0",
            "gw": "10.52.0.1"
        }
    ],
    "dns": {}
}
```

```console
ip link show 
...
12: cni0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue state UP mode DEFAULT group default qlen 1000
    link/ether 12:ce:02:f1:c6:ee brd ff:ff:ff:ff:ff:ff
13: vethaceb2b79@if5: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue master cni0 state UP mode DEFAULT group default 
    link/ether 12:ce:02:f1:c6:ee brd ff:ff:ff:ff:ff:ff link-netns testing1
14: veth707eb98c@cni0.99: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue master cni0 state UP mode DEFAULT group default 
    link/ether 1e:27:94:38:d7:d5 brd ff:ff:ff:ff:ff:ff
15: cni0.99@veth707eb98c: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue state UP mode DEFAULT group default 
    link/ether fe:12:20:c8:cd:3e brd ff:ff:ff:ff:ff:ff
```

squeed:
So, I would expect the `result` to also include `cni0.99` and the corresponding IP.

SchSeba:
Hi @squeed thanks for the comment.

I run the regular bridge (without vlan) and this is the output
```json
{
    "cniVersion": "0.3.1",
    "interfaces": [
        {
            "name": "cni0",
            "mac": "76:f1:74:85:83:d9"
        },
        {
            "name": "veth6b61c0c4",
            "mac": "4e:d8:77:53:1e:88"
        },
        {
            "name": "eth0",
            "mac": "8a:6a:28:58:d8:e0",
            "sandbox": "/var/run/netns/testing"
        }
    ],
    "ips": [
        {
            "version": "4",
            "interface": 2,
            "address": "10.42.0.5/24",
            "gateway": "10.42.0.1"
        }
    ],
    "routes": [
        {
            "dst": "0.0.0.0/0",
            "gw": "10.42.0.1"
        }
    ],
    "dns": {}
}
```


```
ip addr

49: cni0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue state UP group default qlen 1000
    link/ether 76:f1:74:85:83:d9 brd ff:ff:ff:ff:ff:ff
    inet 10.42.0.1/24 brd 10.42.0.255 scope global cni0
       valid_lft forever preferred_lft forever
    inet6 fe80::74f1:74ff:fe85:83d9/64 scope link 
       valid_lft forever preferred_lft forever
50: veth6b61c0c4@if3: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue master cni0 state UP group default 
    link/ether 4e:d8:77:53:1e:88 brd ff:ff:ff:ff:ff:ff link-netnsid 1
    inet6 fe80::4cd8:77ff:fe53:1e88/64 scope link 
       valid_lft forever preferred_lft forever
```

Related to 

> So, I would expect the `result` to also include `cni0.99` and the corresponding IP.

I have some questions:

1. The ip address of cni0 is not listed with the preset behavior of the bridge plugin where I need to add the corresponding IP, I also need to add the for the current behavior too?
2. about the cni0.99 interface I don't added to the interfaces in purpose, the cni0 interface that return is a bridge interface type and the cni0.99 is a veth interface (the container veth is not connected to the cni0.99 interface)

SchSeba:
Hi @squeed I update the PR to add also the cni0.99 interface to the result struct. also the tests are green now can you please take a look?

squeed:
Hi @SchSeba,
You are right, and my understanding of the spec was wrong (to be fair, the other maintainers were of the same opinion). I think you're right, and we should not include the addresses of the host-side interface. I'm very sorry for the confusion.

SchSeba:
Hi @squeed thanks for the comment just to be sure you want me to remove the cni0.99 interface from the result.interfaces list?

SchSeba:
@squeed ping?

dcbw:
> Hi @squeed thanks for the comment just to be sure you want me to remove the cni0.99 interface from the result.interfaces list?

@SchSeba not quite; leave the cni0.99 interface in the Interfaces array. But when there is an L3 address for cni0 or cni0.99 we don't need to include that in the IPs array because the existing bridge plugin does not include the cni0 address there either.

So I think what you have now is what Casey is asking for.

SchSeba:
Thanks for the comment @dcbw so I think this PR is ready to get merge :)

dcbw:
@SchSeba Casey is on PTO today but will do a final check on Monday.

dankenigsberg:
@SchSeba @squeed happy Wednesday!

SchSeba:
Hi @dcbw @squeed any update on this PR?

SchSeba:
Hi @rosenhouse maybe you can help me get this PR merge?

squeed:
Code LGTM. Can you squash in to two commits (vendor and code), and I'll merge?

SchSeba:
Hi @squeed please take a look on my last commit I was requested to add some lines to the test to fix them after the rebase.
If that's fine from your side I will squash it.

SchSeba:
Hi @squeed any update?

squeed:
@SchSeba Gotcha. looks good from my side. Squash this and I'll merge it :-)

SchSeba:
@squeed done and also the tests looks good :)

plugins/main/bridge/bridge.go
```diff
@@ -293,8 +330,12 @@ func calcGatewayIP(ipn *net.IPNet) net.IP {
 }
 
 func setupBridge(n *NetConf) (*netlink.Bridge, *current.Interface, error) {
+	vlanFiltering := false
+	if n.Vlan != 0 {
+		vlanFiltering = true
+	}
```
rootfs, line 336: what if vlan filtering is enabled in the begining (for those with Vlan IDs), and a new vlan, without vlan ID, is to be created. Wouldn't the new vlan disable vlan filtering?

CJCShadowsan:
I know this sounds like an obvious question, but...

...How does this work?

I've tried creating a bridge interface but can't get pod-to-host networking working over VLANs.

dankenigsberg:
> I know this sounds like an obvious question, but...
> 
> ...How does this work?
> 
> I've tried creating a bridge interface but can't get pod-to-host networking working over VLANs.

I believe you must first connect the host to a bridge port with the relevant Vlan ID enabled. You should be able to do so by running this CNI in the host networking namespace and set an IP address on the free veth.

CJCShadowsan:
> > I know this sounds like an obvious question, but...
> > ...How does this work?
> > I've tried creating a bridge interface but can't get pod-to-host networking working over VLANs.
> 
> I believe you must first connect the host to a bridge port with the relevant Vlan ID enabled. You should be able to do so by running this CNI in the host networking namespace and set an IP address on the free veth.

Ok - so my setup:

Host:
- 1x bond device
- 4x vlan devices (on 4 VLANS) attached to bond0 (bond0.101, bond0.102, etc)

Other hosts:
- 1x bond device
- VLAN device bond0.101
- IPMI device on VLAN 102

Via multus:

```
apiVersion: "k8s.cni.cncf.io/v1"
kind: NetworkAttachmentDefinition
metadata:
  name: br1-vl101
spec:
  config: '{
      "cniVersion": "0.4.0",
      "type": "bridge",
      "bridge": "br1",
      "vlan": 101,
      "ipam": {
        "type": "host-local",
        "subnet": "10.8.0.0/24",
        "rangeStart": "10.8.0.220",
        "rangeEnd": "10.8.0.220"
      }
    }'
```

When I attach this to a POD (incidentally a KubeVirt instance), the address comes up and I see the device show up on a bridge vlan command on the physical host... But can't ping a 10.8.0.5 address which is on the same VLAN, on a physical host (not the host this Pod is running on).

Yet... with the real bond0.101 device I can ping 10.8.0.5.

What am I missing? I was hoping there would be some sort of example to make this work considering this merge request enables a VLAN tag on a bridge plugin but i'm just coming up agonisingly short.

dankenigsberg:
Kudos for using KubeVirt ;-) You're most welcome to share your use case on [kubevirt-dev](https://groups.google.com/g/kubevirt-dev)

Can you ping between two pods on the same node over their `br1-vl101` network? You should!
I don't see how you connect `br1` to the external world. I assume it is connected to `bond0`. IIRC Linux vlan devices such as your `bond0.101` grab ethernet frames with the `101` tag before they reach to a bridge connected to `bond0`. To let the both host and bridge see bidirectional vlan 101 traffic, I suggest again that you create a veth pair, with one side attached to a bridge port tagged with 101 and the other with ip address, both on the host network namespace.

If my guess is off, you'd need a bit of `tcpdump` sleuthing to find out where your packets are lost.

CJCShadowsan:
I'm betting this is where i'm falling down here. So there's _more_ that needs to be done in order to make this work? Lol!

So - dummies guide - If we just use the physical bond, bond0 (which carries all VLANS):

```
brctl addbr br1 bond0
ip link set br1 type bridge vlan_filtering 1
ip link set bond0 type bridge vlan_filtering 1
bridge vlan add dev bond0 vid 101 master
```

```
apiVersion: "k8s.cni.cncf.io/v1"
kind: NetworkAttachmentDefinition
metadata:
  name: br1-vl101
spec:
  config: '{
      "cniVersion": "0.4.0",
      "type": "bridge",
      "bridge": "br1",
      "master": "bond0",
      "vlan": 101,
      "ipam": {
        "type": "host-local",
        "subnet": "10.8.0.0/24",
        "rangeStart": "10.8.0.220",
        "rangeEnd": "10.8.0.220"
      }
    }'
```

This creates a veth with VLAN 101, bond0 bridge carries 101... And br1 gets created that links to bond0 if it's created. Is that correct?

gaby:
Can someone provide an example of how is this suppose to work? We can use docker ipvlan and get vlan tagged traffic. When using this plugin we can't ping/reach anything using vlans.

Issue #233, opened 2018-11-20
IPv6 Gateway not allowed to be outside subnet


DennisGlindhart:
When using the following configuration where the gateway is outside of the subnet the following error is received:

**invalid range set 0: gateway 2001:db8:5::1 not in network 2001:db8:5:333::/64**

```
    "ipam": {
        "type": "host-local",
        "subnet": "2001:db8:5:333::/64",
        "gateway": "2001:db8:5::1",
    }
```

It's not allowed in IPv4, but in IPv6 this is perfectly ok to have the gateway in a different subnet as long as the gateway is on the same L2 link. (Gateway Link-Local is looked up via Neighbor Discovery).

This is AFAIK fairly normal for ISP's to assign i.e. /48 or /56 subnet and only provide a single gateway address.

A workaround is to change the subnet size and use range to ensure that IP's are in the intended range.

```
    "ipam": {
        "type": "host-local",
        "subnet": "2001:db8:5:333::/48",
        "gateway": "2001:db8:5::1",
        "rangeStart": "2001:db8:5:333::",
        "rangeEnd": "2001:db8:5:333:ffff:ffff:ffff:ffff",
    }
```

but this results in an address in the container/endpoint with the /48 prefix instead of /64 which is preferable for IPv6 to work properly.
```
# ip a
2: eth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue state UNKNOWN group default 
    link/ether xx:xx:xx:xx:xx:xx brd ff:ff:ff:ff:ff:ff link-netnsid 0
    inet6 2001:db8:5:333::5/48 scope global 
       valid_lft forever preferred_lft forever
    inet6 fe80::x:x/64 scope link 
       valid_lft forever preferred_lft forever
```

squeed:
Oh duh, good catch.

squeed:
Fixed in #236, closing.

PR #236, opened 2018-11-26
remove gateway check 

BSWANG:
TO #233

### DIFF
--- a/plugins/ipam/host-local/backend/allocator/range.go
+++ b/plugins/ipam/host-local/backend/allocator/range.go
@@ -53,10 +53,6 @@ func (r *Range) Canonicalize() error {
 		if err := canonicalizeIP(&r.Gateway); err != nil {
 			return err
 		}
-		subnet := (net.IPNet)(r.Subnet)
-		if !subnet.Contains(r.Gateway) {
-			return fmt.Errorf("gateway %s not in network %s", r.Gateway.String(), subnet.String())
-		}
 	}
 
 	// RangeStart: If specified, make sure it's sane (inside the subnet),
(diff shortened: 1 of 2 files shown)


plugins/ipam/host-local/backend/allocator/range.go
```diff
@@ -49,7 +49,7 @@ func (r *Range) Canonicalize() error {
 	// If the gateway is nil, claim .1
 	if r.Gateway == nil {
 		r.Gateway = ip.NextIP(r.Subnet.IP)
-	} else {
+	} else if r.Gateway.To4() != nil {
```
s1061123: How about to add canonicalizeIP() for IPv6? It seems to just skip code in case of IPv6.

BSWANG: Ok, added check after Canonicalize.

plugins/ipam/host-local/backend/allocator/range.go
```diff
@@ -54,7 +54,7 @@ func (r *Range) Canonicalize() error {
 			return err
 		}
 		subnet := (net.IPNet)(r.Subnet)
-		if !subnet.Contains(r.Gateway) {
+		if r.Gateway.To4() != nil && !subnet.Contains(r.Gateway) {
```
matthewdupre: I think it might be better to remove the check entirely - there are a number of plugins that have gateways outside the subnet in IPv4 as well (e.g. Calico uses a /32 on the interface inside the container with "default via gateway" and "gateway/32 onlink").

It depends on which misconfigurations we're most likely to catch here.  If it's just typos in the gateway, then @dcbw points out that the incorrect gateway could be in subnet and still break things.

WDYT?

BSWANG: Yes, IMO, the `ipam` plugin doesn't need to check `gateway` in subnet.
I will remove the check.

PR #238, opened 2018-11-28
Support ethtool offloads setting for tuning plugin.

hustcat:
Add ethtool offloads setting for `tuning` plugin.


dcbw:
@hustcat would it be possible to add GSO/TSO/etc support to github.com/safchain/ethtool and then re-vendor it and use that instead?

dcbw:
@hustcat any thoughts on my comment https://github.com/containernetworking/plugins/pull/238#issuecomment-442517985 ?

dcbw:
We'd like the ethtool changes to go upstream into safchain/ethtool and then re-vendor that and use it, rather than implementing this all in CNI itself. Otherwise, if there is another ethtool Go library that does implement the functionality, we could switch to that one instead. If that sounds good, please rework and re-open when that's been done. Thanks!

Issue #244, opened 2018-12-08
Create a docker image for easy deployment of binaries


micw:
There should be a docker images so that the binaries can easily be deployed using a kubernetes job or daemonset.

micw:
I have created the Dockerfile. It was quite easy since the build script is already there. I'd appreciate if you could create an official docker hub repo for it.
If you accept the PR, I'll also use the docker image to finally fix https://github.com/intel/multus-cni/issues/192

bboreham:
How do you envisage the binaries inside a container will be executed by the runtime?

Perhaps you could comment on https://github.com/containernetworking/cni/issues/284

micw:
@bboreham That's an interesting discussion. But I get the impression that it went the wrong way (adding more complexity, making cni plugins more smart about how to find and call other plug-ins).
The latest comment (https://github.com/containernetworking/cni/issues/284#issuecomment-240950145) shows a very good idea how this could be done. One container that manage the references plug-ins could be a building brick to achive this.
When this is done, multus could profit a lot of it. Rather than calling other CNI Plug-Ins it would just need to tell the lib which to call.

But it will require a big design change in both, cni and multus - so it will take time to do this. Meanwhile I'd just pack the binaries to a container and install it into /opt/cni/bin.

bboreham:
>  just pack the binaries to a container and install it into /opt/cni/bin.

I don’t get what you mean here. Please explain step by step. 

Also how does this fit with “deployed using a kubernetes job or daemonset“?

micw:
@bboreham Sorry for delay, I'll explain:

The most common way, cni plug-ins are currently installed to kubernetes clusters is the following:
Someone who wrote a CNI packs the plug-ins into a container. On cni startup (in most cases on an initContainer), the host's /opt/cni/bin is mounted into a container and a script within the container copies the plug-ins to this directory. Then this plug-in (and maybe other, nested cni plug-ins) can not access the binaries by mounting the host's /opt/cni/bin directory.

Since there's no official docker image that contains the default cni plug-ins, each cni plug-in packs a copy of the default plug-ins it needs. Flannel for example ships a copy of the "hostlocal" plug-in as well as some others - but not all.

If we had an official docker image, this could be used to install the default plug-ins.

I have created an inofficial one and added it to the multus-cni. This is a good example how the docker image can be used: https://github.com/intel/multus-cni/pull/207/files

bboreham:
[...]

PR #245, opened 2018-12-08
Add a Dockerfile that contains the built CNI plug-ins

micw:
Fixes https://github.com/containernetworking/plugins/issues/244

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,7 +7,7 @@ WORKDIR /script
 RUN /script/test_install_cni_plugins.sh
 
 # In this container we build the plug-ins
-FROM golang:1.11 as build
+FROM golang:1.15 as build
 
 ENV BRANCH_OR_TAG=master
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,4 +20,4 @@ FROM busybox
 COPY --from=build /go/src/app/bin /opt/cni/bin
 VOLUME /host/opt/cni/bin
 ADD docker-installer/install_cni_plugins.sh /script/install_cni_plugins.sh
-CMD ['/script/install_cni_plugins.sh','/opt/cni/bin','/host/opt/cni/bin']
+CMD ["/script/install_cni_plugins.sh","/opt/cni/bin","/host/opt/cni/bin"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,10 +1,23 @@
-FROM golang:1.11.2 as build
+# In this container we test the install script
+FROM busybox as test
+
+ADD docker-installer/install_cni_plugins.sh /script/install_cni_plugins.sh
+ADD docker-installer/test_install_cni_plugins.sh /script/test_install_cni_plugins.sh
+WORKDIR /script
+RUN /script/test_install_cni_plugins.sh
+
+# In this container we build the plug-ins
+FROM golang:1.11 as build
+
+ENV BRANCH_OR_TAG=master
 
 WORKDIR /go/src/app
-COPY . .
+ADD . .
 RUN ./build_linux.sh
 
+# This is the final, minimal container
 FROM busybox
 COPY --from=build /go/src/app/bin /opt/cni/bin
-WORKDIR /opt/cni/bin
-
+VOLUME /host/opt/cni/bin
+ADD docker-installer/install_cni_plugins.sh /script/install_cni_plugins.sh
+CMD ['/script/install_cni_plugins.sh','/opt/cni/bin','/host/opt/cni/bin']
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,10 @@
+FROM golang:1.11.2 as build
+
+WORKDIR /go/src/app
+COPY . .
+RUN ./build_linux.sh
+
+FROM busybox
+COPY --from=build /go/src/app/bin /opt/cni/bin
+WORKDIR /opt/cni/bin
+
--- a/docker-installer/install_cni_plugins.sh
+++ b/docker-installer/install_cni_plugins.sh
@@ -1,0 +1,59 @@
+#!/bin/sh
+
+set -e
+
+if [ $# -ne 2 ]; then
+  echo "USAGE: $0 source-dir dest-dir"
+  exit 1
+fi
+
+SRC=$1
+DST=$2
+
+install_cni_plugin() {
+	NAME=$1
+	MD5=$( md5sum $SRC/$NAME | awk '{ print $1 }' )
+	if [ -e $DST/$NAME ]; then
+		if [ ! -e $DST/$NAME.by_cni_installer_image ]; then
+			# The file already exists but there's no marker that
+			# it was installed by this installer -> keep untouched
+			echo "* '$NAME' ignored (exists but not installed by me)"
+			return
+		fi
+		OTHER_MD5=$( md5sum $DST/$NAME | awk '{ print $1 }' )
+		INSTALLED_MD5=$( cat $DST/$NAME.by_cni_installer_image )
+
+		if [ "$OTHER_MD5" != "$INSTALLED_MD5" ]; then
+			# The file was previously installed by this installer
+			# but later changed -> keep untouched
+			echo "* '$NAME' ignored (previously installed by me but changed by someone else)"
+			return
+		fi
+
+		if [ "$OTHER_MD5" == "$MD5" ]; then
+			# The file was previously installed by this installer
+			# and is up-to-date
+			echo "* '$NAME' is up-to-date"
+			return
+		fi
+
+		# The file was previously installed by this installer
+		# but needs an update
+		cp -a $SRC/$NAME $DST/$NAME
+		echo $MD5 > $DST/$NAME.by_cni_installer_image
+		echo "* '$NAME' updated"
+
+	else
+		cp -a $SRC/$NAME $DST/$NAME
+		echo $MD5 > $DST/$NAME.by_cni_installer_image
+		echo "* '$NAME' installed"
+	fi
+}
+
+echo "Installing CNI plug-ins to $DST"
+echo
+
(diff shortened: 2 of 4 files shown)


Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM golang:1.11.2 as build
```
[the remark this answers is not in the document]
micw: I'm using a multi-stage docker build here. The 1st container is used to build the binaries. It does not need to be the latest version - pinning it will prevent us from surprises by sudden version upgrades. But if it's important to have it built on the latest go version, of course I can change it to "latest".

The actual (delivered) docker image is build in the second stage - it consists of the latest busybox image with the binaries created in the build stage. Nothing else from stage 1 is in this image (ensuring smallest possible image size).

micw:
> I’d be curious to know what maintaining this would involve.
Usually no extra effort except setting up an automated build on docker hub (a few clicks). Docker hub can automatically build "latest" on commit to master and version-tags based on commits to branches or new tags.

> Would we need to trigger a new build on a regular cadence?
This is done automatically on commits. It is also possible to auto-trigger if certain images on docker hub are updated (e.g. new golang version or new busybox container).

Dockerfile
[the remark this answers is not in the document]
rosenhouse: > pinning it will prevent us from surprises by sudden version upgrades

Good point.  Still, if there's a security vulnerability in the golang compiler such that it produces vulnerable binaries, I'd prefer we pick up the fix automatically.  Could we omit patch version number here?  i.e. tag it as `golang:1.11`

We bump the go minor version semi-regularly (e.g. https://github.com/containernetworking/plugins/pull/223) so we could bump this version at the same time.

rosenhouse:
Thanks, this makes a lot of sense to me.  I like the idea of automatic, isolated, CI-driven reproducible builds as an alternative to our current binary release process.

Still, publishing binaries on dockerhub would be a significant change, and I'd want consensus from @containernetworking/cni-maintainers before committing to this.  With Kubecon and upcoming holidays, this may take a bit.

Thanks again.

matthewdupre:
I'd much prefer to have the binaries be built and published somewhere automatically.  I don't have strong feelings about docker hub vs quay etc.  I'm also not too fussed whether we use docker to do the build, or have some other CI system (Travis, Semaphore etc) do it.

micw:
@matthewdupre Having releases built by travis-ci is easy and does not conflict with the docker builds - these are just another option for deployment. Have a look at https://github.com/micw/elasticsearch-kopf/blob/master/.travis.yml - that's a minimal config that pushes every tag build on travis to github releases.

If you have working travis releases, travis could also create a docker image from it and push it to docker hub. But it's not as easy to setup as letting docker hub do the multistage build.

bboreham:
I think this PR is insufficiently useful to merge. 

Some documentation about the expected usage would help. Especially what to do if the underlying machine has the same files installed by package. 

See the discussion on #244.

micw:
@bboreham I do agree about the documentation. What would be a good place for it? I could also add an install-script that does the copy stuff.

> Especially what to do if the underlying machine has the same files installed by package.

Since there's no easy way to figure out what the latest or "right" version of the binaries, it's up to the user do decide how it should be installed. So if you decide to install calico on your machine where you previously installed the default plug-ins via package, they simply got replaced.

Here's how canal/calico does it currently:

1. They build a few reference plug-ins in their makefile: 
https://github.com/projectcalico/cni-plugin/blob/master/Makefile#L274

2. They package it with their CNI and an installer script:
https://github.com/projectcalico/cni-plugin/blob/master/k8s-install/scripts/install-cni.sh#L72

3. They launch the container and run the script:
https://github.com/projectcalico/calico/blob/master/v2.6/getting-started/kubernetes/installation/hosted/calico.yaml#L203

(Canal uses the same script: https://github.com/projectcalico/canal/blob/master/k8s-install/canal.yaml#L139)


The advantage of an official docker image containing the plug-ins would be that you would not get a custom-build version per CNI but an official one build by you.

bboreham:
Just so we're on the same page, I used the same technique, in the same timeframe: https://github.com/weaveworks-experiments/weave-kube/commit/b90bfd54fcbd12bad13198e4f5b6d4c5d5ed5b0d . It's not the broad idea I need from you, it's the precise details of how an end-user would apply your idea.

I still don't see how the files would get from the CNI-built docker image to the Calico deployment.  Calico would say `from cni-plugins` in their Dockerfile ?  Calico would `docker cp` the files ?

micw:
@bboreham I'd say, calico (or multus-cni in my case) would call your docker image as an initContainer if they want to install the default plug-ins rather than calling their orn docker image with their own version of the default plug-ins.

bboreham:
So your proposed CNI image needs to contain a script which copies the binaries?

In this PR I can only see a Dockerfile; no install script.

micw:
@bboreham see https://github.com/containernetworking/plugins/pull/245#issuecomment-447480246

micw:
@squeed Thank you for your feedback.

> If we are going to ship the binaries as a docker image, they need to be the released binaries built with scripts/release.sh or something similar

This is true, there should be a single source of binaries, so the Docker image should be created from the official releases. Either directly in your travis.yaml or via extra build job.

> Does it make sense for the docker image to always copy to /host/cni (if it exists)? Then you could use it as an init container.

The target directory may vary, /host/cni/bin is the most common. There is another use case for this image where the binaries would never be copied: https://github.com/containernetworking/cni/issues/284 . The discussion there is in a very early state but it might turn out that they want to call the binaries directly within a container.

micw:
@squeed I have reworked my docker image and added an installer script. That script created a marker-file for each cni-plugin it installs/updates and ensures that existing plugins (without marker file) are untouched.

If you are fine with this, I'd appreciate if that could be merged and an official image could be created from it. I'm currently building an inofficial image from that code so that we can use it in multus-cni (https://github.com/micw/cni-plugin-installer https://github.com/intel/multus-cni/pull/207)

micw:
@bboreham
> So your proposed CNI image needs to contain a script which copies the binaries?
> 
> In this PR I can only see a Dockerfile; no install script.

I have added this script now. It also ensures that the plug-ins are not overwritten if they are already in place.

halfcrazy:
any update?

c0ffee:
hmm I'm struggeling with this on my k3s installation, any plans on merging this?

Dockerfile
```diff
@@ -0,0 +1,23 @@
+# In this container we test the install script
+FROM busybox as test
+
+ADD docker-installer/install_cni_plugins.sh /script/install_cni_plugins.sh
+ADD docker-installer/test_install_cni_plugins.sh /script/test_install_cni_plugins.sh
+WORKDIR /script
+RUN /script/test_install_cni_plugins.sh
+
+# In this container we build the plug-ins
+FROM golang:1.11 as build
```
dcbw, line 10: @micw could this get updated to 1.15?

bboreham:
I have no objections to merging now it seems that the basic requirements on usability are covered.
However I imagine what you are really looking for is the built images, which requires a little more work.
Especially if we script it to cover all architectures.

docker-installer/install_cni_plugins.sh
```diff
@@ -0,0 +1,59 @@
...
+		if [ "$OTHER_MD5" == "$MD5" ]; then
+			# The file was previously installed by this installer
+			# and is up-to-date
+			echo "* '$NAME' is up-to-date"
+			return
+		fi
+
+		# The file was previously installed by this installer
+		# but needs an update
+		cp -a $SRC/$NAME $DST/$NAME
```
dcbw, line 42: @micw do we really want `-a` here? that adds `-R` and I don't think we want recursive copy of a single file right?

squeed, line 42: We should make these writes atomic, so can you copy to a temporary file (in the same directory) then `mv` them? That way there's no chance of reading a partially-written executable.

I actually had to fix this bug in a different container I maintain, so it can definitely happen :-/

micw: I'm impressed that my PR woke up after 2 years of deep sleep ;-)
If I recall correctly, the plug-ins are executables, so either -a is required or a chmod/chown afterwards.

micw: Another option would be to name them $NAME.by_cni_installer (rather than creating that marker file) and set a symlink. That makes handling of installation/updates easier:
* if $NAME not exists -> install it
* if $NAME exists and is the correct symlink -> probably update the binary
* if $NAME exists and is not a symlink (or has wrong target) -> not our file, keep it

dcbw: @micw `-a` is an alias for `-dR --preserve=all` but we dont' want the `-R`. So you could change this to:

`cp -d --preserve=all $SRC/$NAME $DST/$NAME`

and solve the issue?

bboreham:
Thinking about how we make a release, the procedure to build this image should be integrated with the release script.
Further, the executable files inside this container should be identical to those we release as plain binaries.

That suggests that we don't want a build stage as part of the container: we should run the non-container build then copy the binaries into a container.  (Or alternatively stop building the non-containerised binaries, but I think this will be less acceptable)

docker-installer/install_cni_plugins.sh
squeed: You still have the same problem, though - if you write $NAME.by_cni_installer in place, you could still get a partial read.

zg27 (approved):
#

e0ne:
@micw hi, are you going to continue work on this PR?

e0ne:
Here is an example how we build image for https://github.com/Mellanox/network-operator/: https://github.com/Mellanox/k8s-images/blob/master/container-networking-plugins/Dockerfile

maxpain:
Any updates?

PR #246, opened 2018-12-10
Move Windows tests to Travis

benmoss:
I more or less tried combining the existing `.appveyor.yml` and the `test.sh` script. I removed line from the travis.yml installing gcc-multilib gcc-mingw-w64 since that won't work on Windows, and doesn't appear to be necessary on Linux either.

I had to change `build_windows.sh` to create a temp directory for the GOPATH tricks that we do, since building a GOPATH in the current directory ends up with a path longer 255 characters which Windows doesn't like.

Another semi-unfortunate thing is having to manually specify to test the `pkg/hns` package. I couldn't think of another way, since other other packages in that directory don't build on Windows.


benmoss:
Appveyor failing is expected since I removed its config file 😄

test_windows.sh
```diff
@@ -0,0 +1,14 @@
...
+# Run CNI plugin tests.
+#
+set -e
+
+source ./build_windows.sh
+
+echo "Running tests"
+
+PLUGINS=$(cat plugins/windows_only.txt | tr '\n' ' ')
+GINKGO_FLAGS="-p -r --randomizeAllSpecs --randomizeSuites --failOnPending --progress pkg/hns $PLUGINS"
```
rosenhouse: I'm not thrilled with this explicit `pkg/hns` but I'm also not ready to say we need a new abstraction here either.  If we see other windows-only packages that aren't getting run in travis because of omission from this list, then we can revisit.

rosenhouse:
Yeah we can disable AppVeyor after this is merged.

squeed:
Removed the appveyor hook too.

benmoss:
Looks like there's some flakiness in Travis itself which is causing a failure on Windows/Go 1.10.x right now: https://travis-ci.community/t/go-command-not-found/340

Issue #192, opened 2018-08-17
ipam/dhcp: please allow setting default gateway


onlyjob:
Similar to `ipam/host-local` there should be option to set/override default gateway in the `ipam/dhcp`. This is especially important as currently `ipam/dhcp` does not use gateway info from DHCP server.

Lack of gateway in DHCP prevents communication with networks beyond LAN.

saravanakumar-periyasamy:
check this comment - https://github.com/containernetworking/cni/issues/310#issuecomment-363146469

saravanakumar-periyasamy:
may be it can ignore the routes from DHCP and use the override if its explicitly set.

dcbw:
#255 has been merged, so I think this can be closed.

PR #255, opened 2019-01-29
some DHCP servers expect to request for explicit router options

saravanakumar-periyasamy:
Without setting this, the default route is missing in the containers. This is related to #192


squeed:
Very good catch, nice!
LGTM.

dcbw:
lgtm

PR #259, opened 2019-02-13
Quote PATH in Travis settings

bboreham:
Hypothesis that absence of quoting causes the failure at https://travis-ci.org/containernetworking/plugins/jobs/489972970


bboreham:
Apparently the Travis infrastructure has already added some quoting so we don't need to.

PR #264, opened 2019-02-19
Add CHECK support for linux plugins

### LINKED ISSUE #263: Add Check Support
[CNI Spec 0.4.0-dev ](  https://github.com/containernetworking/cni/blob/master/SPEC.md) supports CHECK API.   Add support to plugins

mccv1r0:
Add check support for: 

bridge, ipvlan, macvlan, p2p, vlan and host-device main plugins
host-local and static ipam plugins
tuning, bandwidth and portmap meta plugins

Utility functions created for common PrevResult checking

Fixes #263

### DIFF
--- a/plugins/ipam/host-local/main.go
+++ b/plugins/ipam/host-local/main.go
@@ -15,6 +15,7 @@
 package main
 
 import (
+	"encoding/json"
 	"fmt"
 	"net"
 	"strings"
@@ -30,12 +31,38 @@ import (
 
 func main() {
 	// TODO: implement plugin version
-	skel.PluginMain(cmdAdd, cmdGet, cmdDel, version.All, "TODO")
+	skel.PluginMain(cmdAdd, cmdCheck, cmdDel, version.All, "TODO")
 }
 
-func cmdGet(args *skel.CmdArgs) error {
-	// TODO: implement
-	return fmt.Errorf("not implemented")
+func loadNetConf(bytes []byte) (*types.NetConf, string, error) {
+	n := &types.NetConf{}
+	if err := json.Unmarshal(bytes, n); err != nil {
+		return nil, "", fmt.Errorf("failed to load netconf: %v", err)
+	}
+	return n, n.CNIVersion, nil
+}
+
+func cmdCheck(args *skel.CmdArgs) error {
+
+	ipamConf, _, err := allocator.LoadIPAMConfig(args.StdinData, args.Args)
+	if err != nil {
+		return err
+	}
+
+	// Look to see if there is at least one IP address allocated to the container
+	// in the data dir, irrespective of what that address actually is
+	store, err := disk.New(ipamConf.Name, ipamConf.DataDir)
+	if err != nil {
+		return err
+	}
+	defer store.Close()
+
+	containerIpFound := store.FindByID(args.ContainerID, args.IfName)
+	if containerIpFound == false {
+		return fmt.Errorf("host-local: Failed to find address added by container %v", args.ContainerID)
+	}
+
+	return nil
 }
 
 func cmdAdd(args *skel.CmdArgs) error {
--- a/plugins/ipam/static/main.go
+++ b/plugins/ipam/static/main.go
@@ -59,12 +59,59 @@ type Address struct {
 
 func main() {
 	// TODO: implement plugin version
-	skel.PluginMain(cmdAdd, cmdGet, cmdDel, version.All, "TODO")
+	skel.PluginMain(cmdAdd, cmdCheck, cmdDel, version.All, "TODO")
 }
 
-func cmdGet(args *skel.CmdArgs) error {
-	// TODO: implement
-	return fmt.Errorf("not implemented")
+func loadNetConf(bytes []byte) (*types.NetConf, string, error) {
+	n := &types.NetConf{}
+	if err := json.Unmarshal(bytes, n); err != nil {
+		return nil, "", fmt.Errorf("failed to load netconf: %v", err)
+	}
+	return n, n.CNIVersion, nil
+}
+
+func cmdCheck(args *skel.CmdArgs) error {
+	ipamConf, _, err := LoadIPAMConfig(args.StdinData, args.Args)
+	if err != nil {
+		return err
+	}
+
+	// Get PrevResult from stdin... store in RawPrevResult
+	n, _, err := loadNetConf(args.StdinData)
+	if err != nil {
+		return err
+	}
+
+	// Parse previous result.
+	if n.RawPrevResult == nil {
+		return fmt.Errorf("Required prevResult missing")
+	}
+
+	if err := version.ParsePrevResult(n); err != nil {
+		return err
+	}
+
+	result, err := current.NewResultFromResult(n.PrevResult)
+	if err != nil {
+		return err
+	}
+
+	// Each configured IP should be found in result.IPs
+	for _, rangeset := range ipamConf.Addresses {
+		for _, ips := range result.IPs {
+			// Ensure values are what we expect
+			if rangeset.Address.IP.Equal(ips.Address.IP) {
+				if rangeset.Gateway == nil {
+					break
+				} else if rangeset.Gateway.Equal(ips.Gateway) {
+					break
+				}
+				return fmt.Errorf("static: Failed to match addr %v on interface %v", ips.Address.IP, args.IfName)
+			}
+		}
+	}
+
+	return nil
 }
 
 // canonicalizeIP makes sure a provided ip is in standard form
--- a/plugins/main/bridge/bridge.go
+++ b/plugins/main/bridge/bridge.go
@@ -537,10 +537,269 @@ func cmdDel(args *skel.CmdArgs) error {
 
 func main() {
 	// TODO: implement plugin version
-	skel.PluginMain(cmdAdd, cmdGet, cmdDel, version.All, "TODO")
+	skel.PluginMain(cmdAdd, cmdCheck, cmdDel, version.All, "TODO")
 }
 
-func cmdGet(args *skel.CmdArgs) error {
-	// TODO: implement
-	return fmt.Errorf("not implemented")
+type cniBridgeIf struct {
+	Name        string
+	ifIndex     int
+	peerIndex   int
+	masterIndex int
+	found       bool
+}
+
+func validateInterface(intf current.Interface, expectInSb bool) (cniBridgeIf, netlink.Link, error) {
+
+	ifFound := cniBridgeIf{found: false}
+	if intf.Name == "" {
+		return ifFound, nil, fmt.Errorf("Interface name missing ")
+	}
+
+	link, err := netlink.LinkByName(intf.Name)
+	if err != nil {
+		return ifFound, nil, fmt.Errorf("Interface name %s not found", intf.Name)
+	}
+
+	if expectInSb {
+		if intf.Sandbox == "" {
+			return ifFound, nil, fmt.Errorf("Interface %s is expected to be in a sandbox", intf.Name)
+		}
+	} else {
+		if intf.Sandbox != "" {
+			return ifFound, nil, fmt.Errorf("Interface %s should not be in sandbox", intf.Name)
+		}
+	}
+
+	return ifFound, link, err
+}
+
+func validateCniBrInterface(intf current.Interface, n *NetConf) (cniBridgeIf, error) {
+
+	brFound, link, err := validateInterface(intf, false)
+	if err != nil {
+		return brFound, err
+	}
+
+	_, isBridge := link.(*netlink.Bridge)
+	if !isBridge {
+		return brFound, fmt.Errorf("Interface %s does not have link type of bridge", intf.Name)
+	}
+
+	if intf.Mac != "" {
+		if intf.Mac != link.Attrs().HardwareAddr.String() {
+			return brFound, fmt.Errorf("Bridge interface %s Mac doesn't match: %s", intf.Name, intf.Mac)
+		}
+	}
+
+	linkPromisc := link.Attrs().Promisc != 0
+	if linkPromisc != n.PromiscMode {
+		return brFound, fmt.Errorf("Bridge interface %s configured Promisc Mode %v doesn't match current state: %v ",
+			intf.Name, n.PromiscMode, linkPromisc)
+	}
+
+	brFound.found = true
+	brFound.Name = link.Attrs().Name
+	brFound.ifIndex = link.Attrs().Index
+	brFound.masterIndex = link.Attrs().MasterIndex
+
+	return brFound, nil
+}
+
+func validateCniVethInterface(intf *current.Interface, brIf cniBridgeIf, contIf cniBridgeIf) (cniBridgeIf, error) {
+
+	vethFound, link, err := validateInterface(*intf, false)
+	if err != nil {
+		return vethFound, err
+	}
+
+	_, isVeth := link.(*netlink.Veth)
+	if !isVeth {
+		// just skip it, it's not what CNI created
+		return vethFound, nil
+	}
+
+	_, vethFound.peerIndex, err = ip.GetVethPeerIfindex(link.Attrs().Name)
+	if err != nil {
+		return vethFound, fmt.Errorf("Unable to obtain veth peer index for veth %s", link.Attrs().Name)
+	}
+	vethFound.ifIndex = link.Attrs().Index
+	vethFound.masterIndex = link.Attrs().MasterIndex
+
+	if vethFound.ifIndex != contIf.peerIndex {
+		return vethFound, nil
+	}
+
+	if contIf.ifIndex != vethFound.peerIndex {
+		return vethFound, nil
+	}
+
+	if vethFound.masterIndex != brIf.ifIndex {
+		return vethFound, nil
+	}
+
+	if intf.Mac != "" {
+		if intf.Mac != link.Attrs().HardwareAddr.String() {
+			return vethFound, fmt.Errorf("Interface %s Mac doesn't match: %s not found", intf.Name, intf.Mac)
+		}
+	}
+
+	vethFound.found = true
+	vethFound.Name = link.Attrs().Name
+
+	return vethFound, nil
+}
+
+func validateCniContainerInterface(intf current.Interface) (cniBridgeIf, error) {
+
+	vethFound, link, err := validateInterface(intf, true)
+	if err != nil {
+		return vethFound, err
+	}
+
+	_, isVeth := link.(*netlink.Veth)
+	if !isVeth {
+		return vethFound, fmt.Errorf("Error: Container interface %s not of type veth", link.Attrs().Name)
+	}
+	_, vethFound.peerIndex, err = ip.GetVethPeerIfindex(link.Attrs().Name)
+	if err != nil {
+		return vethFound, fmt.Errorf("Unable to obtain veth peer index for veth %s", link.Attrs().Name)
+	}
+	vethFound.ifIndex = link.Attrs().Index
+
+	if intf.Mac != "" {
+		if intf.Mac != link.Attrs().HardwareAddr.String() {
+			return vethFound, fmt.Errorf("Interface %s Mac %s doesn't match container Mac: %s", intf.Name, intf.Mac, link.Attrs().HardwareAddr)
+		}
+	}
+
+	vethFound.found = true
+	vethFound.Name = link.Attrs().Name
+
+	return vethFound, nil
+}
+
+func cmdCheck(args *skel.CmdArgs) error {
+
+	n, _, err := loadNetConf(args.StdinData)
+	if err != nil {
+		return err
+	}
+	netns, err := ns.GetNS(args.Netns)
+	if err != nil {
+		return fmt.Errorf("failed to open netns %q: %v", args.Netns, err)
+	}
+	defer netns.Close()
+
+	// run the IPAM plugin and get back the config to apply
+	err = ipam.ExecCheck(n.IPAM.Type, args.StdinData)
+	if err != nil {
+		return err
+	}
+
+	// Parse previous result.
+	if n.NetConf.RawPrevResult == nil {
+		return fmt.Errorf("Required prevResult missing")
+	}
+
+	if err := version.ParsePrevResult(&n.NetConf); err != nil {
+		return err
+	}
+
+	result, err := current.NewResultFromResult(n.PrevResult)
+	if err != nil {
+		return err
+	}
+
+	var errLink error
+	var contCNI, vethCNI cniBridgeIf
+	var brMap, contMap current.Interface
+
+	// Find interfaces for names whe know, CNI Bridge and container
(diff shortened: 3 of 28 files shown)


dcbw:
@mccv1r0 I mentioned this in a private email, but we'll need to split out the Linux CHECK code that depends on netlink into *_linux.go files with "//+build linux" at the top, and a stubbed out implementation in a "*_windows.go" file with "//+build windows". That should fix the windows build errors.

plugins/ipam/host-local/main.go
```diff
@@ -26,16 +27,71 @@ import (
...
+		return err
+	}
+
+	// Each address in results.IPs must match what is in the container
+	for _, ips := range result.IPs {
+		ourAddr := netlink.Addr{IPNet: &ips.Address}
+		match := false
+
+		requestedIPs := map[string]net.IP{} //net.IP cannot be a key
+		matchAddr, _ := netlink.ParseAddr(ourAddr.String())
```
dcbw: Since host-local has to build on Windows, and netlink doesn't work on Windows, can we switch to net.ParseCIDR()?

plugins/ipam/static/main.go
```diff
@@ -59,12 +60,68 @@ type Address struct {
...
+	if err != nil {
+		return err
+	}
+
+	// Each address in results.IPs must match what is in the container
+	for _, ips := range result.IPs {
+		match := false
+
+		ourAddr := netlink.Addr{IPNet: &ips.Address}
+		matchAddr, _ := netlink.ParseAddr(ourAddr.String())
```
dcbw: Same as above for host-local; let's swithc to net.ParseCIDR() so we don't have to use netlink.

plugins/ipam/host-local/main.go
mccv1r0: Yes.  Done

plugins/ipam/static/main.go
mccv1r0: Better, here I can just use ips.Address.IP and avoid the need to parse at all.

plugins/ipam/host-local/main.go
```diff
@@ -30,12 +31,66 @@ import (
...
+		return err
+	}
+
+	result, err := current.NewResultFromResult(n.PrevResult)
+	if err != nil {
+		return err
+	}
+
+	// Each address in results.IPs must match what is in the container
+	for _, ips := range result.IPs {
```
dcbw: @mccv1r0 So we decided the following about the IPAM plugins in the last maintainer's meeting. The Spec says:

```
The plugin should return an error if a resource included in the CNI Result type (interface, address or route):
   - was created by the plugin, and
   - is listed in prevResult, and
   - does not exist, or is in an invalid state.
```

So for host-local, that would mean for each file we find in the network name's data directory that contains the ContainerID + interface name ("was created by the plugin") we look for that IP in prevResult ("is listed in prevResult"). If that IP is in prevResult, only then do we "validate" it.

But host-local intentionally doesn't handle anything about the interface, so it shouldn't check for the existence of the IP. So that means that host-local really doesn't have to CHECK anything about the IP addresses themselves and should probably just return "success".

*However*, it certainly should return an error if one of the two following conditions exists:

1) the data directory is not accessible (because the plugin was asked to allocate an IP on ADD there clearly must be a data directory or the ADD would have failed; so there must still be a data dir)
2) the ContainerID cannot be found in the data directory (the plugin would have allocated an IP on ADD so there clearly must be >= 1 file for the container in the data dir)

Does that all make sense?

plugins/ipam/static/main.go
```diff
@@ -59,12 +59,65 @@ type Address struct {
...
+		return err
+	}
+
+	// Each address in results.IPs must match what is in the container
+	for _, ips := range result.IPs {
+		match := false
+
+		// Loop through all ranges until we find the address found in container
+		for _, rangeset := range ipamConf.Addresses {
+			if rangeset.Address.IP.Equal(ips.Address.IP) {
```
dcbw: Similar issue for static as for host-local.  Given the above, we have:

1) loop through static IP we are asked to give to the container from the configuration
2) find that IP in prevResult
3) if the IP exists in prevResult, validate that the prefix/gateway are what we expect

And since it has no data directory or anything, that's it.

So for host-local, that would mean for each file we find in the network name's data directory that contains the ContainerID + interface name ("was created by the plugin") we look for that IP in prevResult ("is listed in prevResult"). If that IP is in prevResult, only then do we "validate" it.

plugins/main/host-device/host-device.go
```diff
@@ -276,10 +284,120 @@ func getLink(devname, hwaddr, kernelpath string) (netlink.Link, error) {
...
+		return err
+	}
+
+	result, err := current.NewResultFromResult(cfg.PrevResult)
+	if err != nil {
+		return err
+	}
+
+	var contMap current.Interface
+	// Find interfaces for name whe know, that of host-device inside container
```
dcbw: "whe" -> "we"

plugins/main/host-device/host-device.go
```diff
@@ -276,10 +284,120 @@ func getLink(devname, hwaddr, kernelpath string) (netlink.Link, error) {
...
+		}
+	}
+
+	// The namespace must be the same as what was configured
+	if args.Netns != contMap.Sandbox {
+		return fmt.Errorf("Sandbox in prevResult %s doesn't match configured netns: %s",
+			contMap.Sandbox, args.Netns)
+	}
+
+	// TODO - Is there another check that can be done?
```
dcbw: I think we can just remove this commented out block.

plugins/main/host-device/host-device.go
```diff
@@ -276,10 +284,120 @@ func getLink(devname, hwaddr, kernelpath string) (netlink.Link, error) {
...
+	}
+	link, err = netlink.LinkByName(intf.Name)
+	if err != nil {
+		return fmt.Errorf("Container Interface name in prevResult: %s not found", intf.Name)
+	}
+	if intf.Sandbox == "" {
+		return fmt.Errorf("Error: Container interface %s should not be in host namespace", link.Attrs().Name)
+	}
+
+	if link.Type() != "dummy" {
```
dcbw: This was probably left in from testing, right? I don't think this will work for eg SRIOV or regular eth devices :)

plugins/main/ipvlan/ipvlan.go
```diff
@@ -146,6 +156,10 @@ func createIpvlan(conf *NetConf, ifName string, netns ns.NetNS) (*current.Interf
 		ipvlan.Mac = contIpvlan.Attrs().HardwareAddr.String()
 		ipvlan.Sandbox = netns.Path()
 
+		// TODO: Bug?  ipvlan's ParentIndex isn't set
```
dcbw: I think here you want MasterIndex, not ParentIndex. Bridge ports, macvlans, ipvlans, and team interfaces will have MasterIndex == to their master interface (IFLA_MASTER), but others like veth will have their ParentIndex set to their peer (IFLA_LINK).

plugins/main/ipvlan/ipvlan.go
```diff
@@ -175,9 +189,20 @@ func cmdAdd(args *skel.CmdArgs) error {
 	var result *current.Result
 	// Configure iface from PrevResult if we have IPs and an IPAM
 	// block has not been configured
-	if n.IPAM.Type == "" && n.PrevResult != nil && len(n.PrevResult.IPs) > 0 {
-		result = n.PrevResult
-	} else {
+	haveResult := false
+	if n.IPAM.Type == "" && n.PrevResult != nil {
+		if err := version.ParsePrevResult(&n.NetConf); err != nil {
```
dcbw: Can't we just use the n.PrevResult here? That already go parsed in loadConf() so we should just be able to call NewResultFromResult() and check len(IPs)

plugins/main/ipvlan/ipvlan.go
```diff
@@ -246,10 +271,139 @@ func cmdDel(args *skel.CmdArgs) error {
...
+			return err
+		}
+	}
+
+	// Parse previous result.
+	if n.NetConf.RawPrevResult == nil {
+		return fmt.Errorf("Required prevResult missing")
+	}
+
+	if err := version.ParsePrevResult(&n.NetConf); err != nil {
```
dcbw: I know above there's the shortcut to loadConf() before it does the prevResult processing. I wonder if we could in fact split up loadConf() so that the Master validation happens in a separate function that cmdAdd() would call. While the rest of loadConf() including prevResult handling can be shared with cmdAdd/Del/Check? then we don't have to do the parsing in each of those functoins but can just use n.PrevResult.

plugins/main/ipvlan/ipvlan.go
```diff
@@ -246,10 +271,139 @@ func cmdDel(args *skel.CmdArgs) error {
...
+	if err != nil {
+		return err
+	}
+
+	var contMap current.Interface
+	// Find interfaces for names whe know, ipvlan inside container
+	for _, intf := range result.Interfaces {
+		if args.IfName == intf.Name {
+			if args.Netns == intf.Sandbox {
+				//contMap.Name = intf.Name
```
dcbw: Remove the commented-out stuff...

plugins/main/ipvlan/ipvlan.go
```diff
@@ -246,10 +271,139 @@ func cmdDel(args *skel.CmdArgs) error {
...
+		}
+	}
+
+	// The namespace must be the same as what was configured
+	if args.Netns != contMap.Sandbox {
+		return fmt.Errorf("Sandbox in prevResult %s doesn't match configured netns: %s",
+			contMap.Sandbox, args.Netns)
+	}
+
+	// TODO - Is there another check that can be done?
```
dcbw: As above I don't think we need this check since you've got the args.Netns != contMap.Sandbox check above which is actually the same thing.

plugins/main/macvlan/macvlan.go
```diff
@@ -135,6 +151,13 @@ func createMacvlan(conf *NetConf, ifName string, netns ns.NetNS) (*current.Inter
 		macvlan.Mac = contMacvlan.Attrs().HardwareAddr.String()
 		macvlan.Sandbox = netns.Path()
 
+		// TODO: Bug?  macvlan's ParentIndex isn't set
```
dcbw: As for ipvlan I think you want MasterIndex not ParentIndex.

plugins/main/ptp/ptp.go
```diff
@@ -286,10 +286,115 @@ func cmdDel(args *skel.CmdArgs) error {
...
-	skel.PluginMain(cmdAdd, cmdGet, cmdDel, version.All, "TODO")
+	skel.PluginMain(cmdAdd, cmdCheck, cmdDel, version.All, "TODO")
 }
 
-func cmdGet(args *skel.CmdArgs) error {
-	// TODO: implement
-	return fmt.Errorf("not implemented")
+func cmdCheck(args *skel.CmdArgs) error {
+	conf := NetConf{}
+	if err := json.Unmarshal(args.StdinData, &conf); err != nil {
```
dcbw: Could you make a loadConf() function to handle the unmarshal and any prevResult parsing needed?

plugins/main/vlan/vlan.go
```diff
@@ -193,10 +194,137 @@ func cmdDel(args *skel.CmdArgs) error {
...
+	if intf.Sandbox == "" {
+		return fmt.Errorf("ptp: Error: Container interface %s should not be in host namespace", link.Attrs().Name)
+	}
+
+	vlan, isVlan := link.(*netlink.Vlan)
+	if !isVlan {
+		return fmt.Errorf("Error: Container interface %s not of type vlan", link.Attrs().Name)
+	}
+
+	// TODO This works when unit testing via cnitool; fails with ./test.sh
```
dcbw: As with ipvlan and macvlan, I think you want MasterIndex

plugins/main/ipvlan/ipvlan.go
```diff
@@ -246,10 +271,139 @@ func cmdDel(args *skel.CmdArgs) error {
...
+		return fmt.Errorf("Container interface name missing in prevResult: %v", intf.Name)
+	}
+	link, err = netlink.LinkByName(intf.Name)
+	if err != nil {
+		return fmt.Errorf("Container Interface name in prevResult: %s not found", intf.Name)
+	}
+	if intf.Sandbox == "" {
+		return fmt.Errorf("Error: Container interface %s should not be in host namespace", link.Attrs().Name)
+	}
+
```
dcbw: Should also check that masterIndex == link.MasterIndex since you're already passing that in.

plugins/main/host-device/host-device.go
mccv1r0: Actually that's the type that is used in the link struct.

plugins/main/macvlan/macvlan.go
```diff
@@ -256,10 +279,140 @@ func cmdDel(args *skel.CmdArgs) error {
...
+	if intf.Name == "" {
+		return fmt.Errorf("Container interface name missing in prevResult: %v", intf.Name)
+	}
+	link, err = netlink.LinkByName(intf.Name)
+	if err != nil {
+		return fmt.Errorf("Container Interface name in prevResult: %s not found", intf.Name)
+	}
+	if intf.Sandbox == "" {
+		return fmt.Errorf("Error: Container interface %s should not be in host namespace", link.Attrs().Name)
+	}
```
dcbw: As with ipvlan should check MasterIndex == parentIndex too.

plugins/main/ptp/ptp.go
```diff
@@ -286,10 +286,115 @@ func cmdDel(args *skel.CmdArgs) error {
...
+		}
+	}
+
+	// The namespace must be the same as what was configured
+	if args.Netns != contMap.Sandbox {
+		return fmt.Errorf("Sandbox in prevResult %s doesn't match configured netns: %s",
+			contMap.Sandbox, args.Netns)
+	}
+
+	// TODO - Is there another check that can be done?
```
dcbw: Let's remove this too as the args.Netns/contMap.Sandbox check does the same thing.

plugins/main/ptp/ptp.go
```diff
@@ -286,10 +286,115 @@ func cmdDel(args *skel.CmdArgs) error {
...
+	}
+	link, err = netlink.LinkByName(intf.Name)
+	if err != nil {
+		return fmt.Errorf("ptp: Container Interface name in prevResult: %s not found", intf.Name)
+	}
+	if intf.Sandbox == "" {
+		return fmt.Errorf("ptp: Error: Container interface %s should not be in host namespace", link.Attrs().Name)
+	}
+
+	_, isVeth := link.(*netlink.Veth)
```
dcbw: Maybe we should check ParentIndex == peer index in hostns too? Kinda like we check that macvlan/ipvlan's MasterIndex is the physical interface index.

plugins/main/vlan/vlan.go
```diff
@@ -193,10 +194,137 @@ func cmdDel(args *skel.CmdArgs) error {
...
+		}
+	}
+
+	// The namespace must be the same as what was configured
+	if args.Netns != contMap.Sandbox {
+		return fmt.Errorf("Sandbox in prevResult %s doesn't match configured netns: %s",
+			contMap.Sandbox, args.Netns)
+	}
+
+	// TODO - Is there another check that can be done?
```
dcbw: As with other plugins let's remove this comented out section.

plugins/main/host-device/host-device.go
mccv1r0: Should I remove?

plugins/main/bridge/bridge.go
```diff
@@ -537,10 +537,282 @@ func cmdDel(args *skel.CmdArgs) error {
...
+
+	// The namespace must be the same as what was configured
+	if args.Netns != contMap.Sandbox {
+		return fmt.Errorf("Sandbox in prevResult %s doesn't match configured netns: %s",
+			contMap.Sandbox, args.Netns)
+	}
+
+	// TODO - Is there another check that can be done?
+	//handle := netns.Get()
+	//if strings.Compare(handle.UniqueId(), contMap.Sandbox) != 0 {
```
dcbw: Remove this commented out block too.

plugins/main/macvlan/macvlan.go
```diff
@@ -256,10 +279,140 @@ func cmdDel(args *skel.CmdArgs) error {
...
+
+	// The namespace must be the same as what was configured
+	if args.Netns != contMap.Sandbox {
+		return fmt.Errorf("Sandbox in prevResult %s doesn't match configured netns: %s",
+			contMap.Sandbox, args.Netns)
+	}
+
+	// TODO - Is there another check that can be done?
+	//handle := netns.Get()
+	//if strings.Compare(handle.UniqueId(), contMap.Sandbox) != 0 {
```
dcbw: Remove commented out block.

plugins/meta/tuning/tuning.go
```diff
@@ -200,10 +203,80 @@ func cmdDel(args *skel.CmdArgs) error {
...
+		for key, conf_value := range tuningConf.SysCtl {
+			fileName := filepath.Join("/proc/sys", strings.Replace(key, ".", "/", -1))
+			fileName = filepath.Clean(fileName)
+
+			contents, err := ioutil.ReadFile(fileName)
+			if err != nil {
+				return err
+			}
+			cur_value := strings.TrimSuffix(string(contents), "\n")
+			if strings.Compare(conf_value, cur_value) != 0 {
```
dcbw: Just use == instead of strings.Compare()

pkg/utils/utils.go
```diff
@@ -39,3 +45,97 @@ func FormatChainName(name string, id string) string {
...
+	for _, route := range resultRoutes {
+		find := &netlink.Route{Dst: &route.Dst, Gw: route.GW}
+		routeFilter := netlink.RT_FILTER_DST | netlink.RT_FILTER_GW
+		var family int
+
+		switch {
+		case route.Dst.IP.To4() != nil:
+			family = netlink.FAMILY_V4
+			// Default route needs Dst set to nil
+			if strings.Compare(route.Dst.String(), "0.0.0.0/0") == 0 {
```
dcbw: Just use == instead of strings.Compare()

plugins/main/bridge/bridge.go
```diff
@@ -537,10 +537,282 @@ func cmdDel(args *skel.CmdArgs) error {
...
+	return brFound, nil
+}
+
+func validateCniVethInterface(intf *current.Interface, brIf cniBridgeIf, contIf cniBridgeIf) (cniBridgeIf, error) {
+
+	var link netlink.Link
+	var err error
+	vethFound := cniBridgeIf{found: false}
+
+	if intf.Name == "" {
```
bboreham: This seems to repeat similar lines from the previous and next function - could they be extracted into a helper function to reduce verbosity?

plugins/main/host-device/host-device.go
```diff
@@ -124,9 +126,15 @@ func cmdAdd(args *skel.CmdArgs) error {
 		if err != nil {
 			return err
 		}
+
+		result.DNS = cfg.DNS // &mike& do we need?
+
+		return types.PrintResult(result, cfg.CNIVersion)
+	} else {
```
bboreham: having an `else` as the last line of a function is not idiomatic Go

plugins/main/ptp/ptp.go
mccv1r0: As discussed via private email, the logic isn't common between add/check/del

plugins/main/ipvlan/ipvlan.go
mccv1r0: logic isn't the same for add/check/del

pkg/utils/utils.go
```diff
@@ -39,3 +45,97 @@ func FormatChainName(name string, id string) string {
...
+		switch {
+		case ips.Version == "4":
+			family = netlink.FAMILY_V4
+		case ips.Version == "6":
+			family = netlink.FAMILY_V6
+		default:
+			return fmt.Errorf("Invalid IP Version %v for interface %v", ips.Version, ifName)
+		}
+
+		gwy, err := netlink.RouteListFiltered(family, findGwy, routeFilter)
```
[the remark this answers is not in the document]
mccv1r0: I tested all main plugins (except windows) and all had a gateway.   But this might be a result of [Issue 273](https://github.com/containernetworking/plugins/issues/273)?  I'm pushing v2 with other comments addressed.  We can come back to this.

plugins/ipam/host-local/main.go
```diff
@@ -30,12 +31,42 @@ import (
...
+	 * Look to see if there is at least one IP address allocated to the container
+	 * in the data dir, irrespective of what that address actually is
+	 */
+	store, err := disk.New(ipamConf.Name, ipamConf.DataDir)
+	if err != nil {
+		return err
+	}
+	defer store.Close()
+
+	//
```
dcbw: Can remove the empty comment here and just below?

mccv1r0: done

pkg/utils/utils.go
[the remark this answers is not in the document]
f1-outsourcing: in #273 I asked if how/if it is possbile to add such a route to the bridge plugin.

`ip route add 192.168.10.153/32 dev eth0`

plugins/main/host-device/host-device.go
```diff
@@ -124,6 +126,10 @@ func cmdAdd(args *skel.CmdArgs) error {
 		if err != nil {
 			return err
 		}
+
+		result.DNS = cfg.DNS // &mike& do we need?
```
dcbw: @mccv1r0 can you remove this comment?

plugins/ipam/host-local/main.go
```diff
@@ -30,12 +31,40 @@ import (
...
+}
+
+func cmdCheck(args *skel.CmdArgs) error {
+
+	ipamConf, _, err := allocator.LoadIPAMConfig(args.StdinData, args.Args)
+	if err != nil {
+		return err
+	}
+
+	/*
```
dcbw: @mccv1r0 Can you convert this comment to "//" format?

plugins/main/macvlan/macvlan.go
```diff
@@ -135,6 +150,13 @@ func createMacvlan(conf *NetConf, ifName string, netns ns.NetNS) (*current.Inter
 		macvlan.Mac = contMacvlan.Attrs().HardwareAddr.String()
 		macvlan.Sandbox = netns.Path()
 
+		// TODO: Bug?  macvlan's ParentIndex isn't set
+		/*
+			if contMacvlan.Attrs().ParentIndex != m.Attrs().Index {
+			   return fmt.Errorf("failed to set macvlan %q Parent Index is: %v should be %v", ifName, contMacvlan.Attrs().ParentIndex, m.Attrs().Index)
+			}
+		*/
```
dcbw: @mccv1r0 Can you remove this commented out code?

PR #265, opened 2019-02-20
Change http to https for security links

huynq0911:
For security, we should change http into https links.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -43,7 +43,7 @@ This is a rough outline of how to prepare a contribution:
 - Make sure your commit messages are in the proper format (see below).
 - Push your changes to a topic branch in your fork of the repository.
 - If you changed code:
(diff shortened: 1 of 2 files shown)


dcbw:
sure...

plugins/meta/bandwidth/README.md
```diff
@@ -59,6 +59,6 @@ Both egressRate and egressBurst must be set in order to limit egress bandwidth.
 
 ## tc tbf documentation
 
-- [tldp traffic control](http://tldp.org/HOWTO/Traffic-Control-HOWTO/components.html)
+- [tldp traffic control](https://tldp.org/HOWTO/Traffic-Control-HOWTO/components.html)
```
matthewdupre: Amusingly this site has an incorrect certificate.  Perhaps we shouldn't change it?

huynq0911: hmm, it can be accepted as long as the URL is not harmful.

huynq0911: I have updated. Please check it!

PR #268, opened 2019-02-22
Fix -hostprefix option

### LINKED ISSUE #267: dhcp: '-hostprefix' option broken
`dhcp daemon -hostprefix /host` should create socket under `/host`, however, current dhcp ignores `-hostprefix` argument and it creates `/run/cni/dhcp.sock`

s1061123:
Fix #267

### DIFF
--- a/plugins/ipam/dhcp/README.md
+++ b/plugins/ipam/dhcp/README.md
@@ -18,7 +18,7 @@ $ ./dhcp daemon
 
 If given `-pidfile <path>` arguments after 'daemon', the dhcp plugin will write
 its PID to the given file.
-If given `-hostprefix <prefix>` arguments after 'daemon', the dhcp plugin will use this prefix for netns as `<prefix>/<original netns>`. It could be used in case of running dhcp daemon as container.
+If given `-hostprefix <prefix>` arguments after 'daemon', the dhcp plugin will use this prefix for DHCP socket as `<prefix>/run/cni/dhcp.sock`. It could be used in case of running dhcp daemon as container.
 
 Alternatively, you can use systemd socket activation protocol.
 Be sure that the .socket file uses /run/cni/dhcp.sock as the socket path.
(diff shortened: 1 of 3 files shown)


bboreham:
If we accept this change the documentation for this option would be incorrect.
https://github.com/containernetworking/plugins/blob/f879dad626893b26c50cc228a3788959f6f295fb/plugins/ipam/dhcp/main.go#L41
https://github.com/containernetworking/plugins/blob/696b1f9ab1022c34e1514e0e29bf84bb9179899f/plugins/ipam/dhcp/README.md#L21

Assuming you have good justification for the change (#267 merely asserts it should be this way), I suggest you change the documentation to match the new behaviour in this PR and add more background in the README.

s1061123:
@bboreham thank you for your comments.
I will update diffs to incorporate your comments. Please take a look into it?

s1061123:
@bboreham second try. could you please take a look into it?

plugins/ipam/dhcp/README.md
```diff
@@ -18,7 +18,7 @@ $ ./dhcp daemon
 
 If given `-pidfile <path>` arguments after 'daemon', the dhcp plugin will write
 its PID to the given file.
-If given `-hostprefix <prefix>` arguments after 'daemon', the dhcp plugin will use this prefix for netns as `<prefix>/<original netns>`. It could be used in case of running dhcp daemon as container.
+If given `-hostprefix <prefix>` arguments after 'daemon', the dhcp plugin will use this prefix for DHCP socket as `<prefix>/run/cni/dhcp.sock`. It could be used in case of running dhcp daemon as container.
```
bboreham: From my reading of the code the parameter is now used for _both_ netns and socket.
The context is that the daemon wants to send DHCP messages from the interface of other containers, so it wants to reach into their namespace.
Maybe you could say something like "use this prefix for references to the host filesystem, e.g. to access netns and the unix socket" ?

s1061123: Changed. Thank you for your valuable comment!

squeed:
lgtm

PR #276, opened 2019-03-14
Add systemd unit file to activate socket with systemd

### LINKED ISSUE #156: dhcp plugin should ship with example systemd .socket file
The DHCP plugin can be socket-activated by systemd, but there is no example .socket file for it.  We should add one.

s1061123:
This changes to add sample systemd unit files to activate socket with systemd. Fix #156.


dcbw:
Awesome, thanks!.

/lgtm

Issue #282, opened 2019-03-26
bridge plugin does not cleanup bridge with no veth attached


pjferrell:
Bridge plugin does not currently cleanup unused bridges after all veth interfaces have been removed.  For backward compatibility with current behavior, plugin should offer a configuration option to cleanup bridge after all veth have been detached.

hujun-open:
+1, also need this function

PR #283, opened 2019-03-26
main/bridge: add option to cleanup unused bridge as part of cmdDEL (issue #282)

pjferrell:
implements issue #282 

this commit adds an optional configuration to cleanup bridge if no veth are attached.  cmdDEL implements a final check of bridge's hwaddr to determine if veth are attached, and whether to remove bridge.

Tests have been added to verify bridge deletion only occurs when enabled and all veth are detached.

### DIFF
--- a/plugins/main/bridge/bridge.go
+++ b/plugins/main/bridge/bridge.go
@@ -15,6 +15,7 @@
 package main
 
 import (
+	"bytes"
 	"encoding/json"
 	"errors"
 	"fmt"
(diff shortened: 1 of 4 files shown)


dcbw:
The large reason we haven't done something like this before is because of race conditions. It's completely possible to have two instances of the plugin running at the same time, one for ADD and one for DEL. The DEL could be deleting the bridge right after the ADD from the other instance has added an interface.

The solution to that is adding some locking to prevent concurrent plugin processes from doing stuff like this, and to have the code check whether any container interfaces are attached to the bridge before deleting it. Thus, if the DEL wins the race and deletes the bridge interface, the concurrent ADD will re-create the interface when it wins the lock.

pjferrell:
@dcbw Thanks for the feedback.  Would it be acceptable to implement a file lock at CNI plugin startup (acquired before ADD / DEL is processed)?  I could leverage the existing flock implementation in ipam host-local plugin (plugins/ipam/host-local/backend/disk/lock.go).

dcbw:
@pjferrell it's worth a shot; take a look at plugins/ipam/host-local/backend/disk/lock.go for an example of how that might be implemented.

PR #286, opened 2019-03-27
Windows: No error for no endpoint on DEL. Ipam clean-up on ADD failure

nagiesek:
We used to return error if no endpoint was found during delete. We now treat this as a success. If we fail during an add call, we now make a delete delegate call to the ipam to clean-up.


plugins/main/windows/win-overlay/win-overlay_windows.go
```diff
@@ -153,7 +163,8 @@ func cmdDel(args *skel.CmdArgs) error {
 
 	epName := hns.ConstructEndpointName(args.ContainerID, args.Netns, n.Name)
 
-	return hns.DeprovisionEndpoint(epName, args.Netns, args.ContainerID)
+	err = hns.DeprovisionEndpoint(epName, args.Netns, args.ContainerID)
```
dcbw: this looks odd; error doesn't seem to get used anywhere? I'm probably missing something.

plugins/main/windows/win-bridge/win-bridge_windows.go
```diff
@@ -119,6 +120,13 @@ func cmdHnsAdd(args *skel.CmdArgs, n *NetConf, cniVersion *string) error {
 		}
 		return hnsEndpoint, nil
 	})
+	defer func() {
```
nagiesek: Move down to Add function

plugins/main/windows/win-overlay/win-overlay_windows.go
dcbw: eg, return the error, or ignore it (with `_ = hns.DeprovisionEndpoint(...)` and `return nil`)

PR #287, opened 2019-04-01
ipvlan support default route

chenk008:
When I use flannel with ipVlan, I need the default route in container.
Add "isDefaultGateway" in ipVlan,  ipVlan can work  in Kubernetes with 'l3s' mode.

### DIFF
--- a/plugins/main/ipvlan/ipvlan.go
+++ b/plugins/main/ipvlan/ipvlan.go
@@ -203,7 +203,7 @@ func cmdAdd(args *skel.CmdArgs) error {
 
 	result.Interfaces = []*current.Interface{ipvlanInterface}
 
-	if n.IsDefaultGW{
+	if n.IsDefaultGW {
 		dst, err := mustCIDR("0.0.0.0/0")
 		if err != nil {
 			return err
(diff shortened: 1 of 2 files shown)


dcbw:
Typically this would be solved by adding the default route to the 'routes' section of the CNI configuration. Does that work for you?

```
{
    "cniVersion": "0.3.1",
    "name": "example",
    "type": "ipvlan",
    "ipam": {
        "type": "host-local",
        "subnet": "10.88.0.0/16",
        "routes": [
            { "dst": "0.0.0.0/0" }
        ]
    }
}
```

chenk008:
> Typically this would be solved by adding the default route to the 'routes' section of the CNI configuration. Does that work for you?
> 
> ```
> {
>     "cniVersion": "0.3.1",
>     "name": "example",
>     "type": "ipvlan",
>     "ipam": {
>         "type": "host-local",
>         "subnet": "10.88.0.0/16",
>         "routes": [
>             { "dst": "0.0.0.0/0" }
>         ]
>     }
> }
> ```

Thank you for reviewing.
I use ipvlan with flannel in kubernetes.The ipam config is generated by flannel,and flannel do not has a config for route.So i think default route is need.

dcbw:
Aha. So what we should do here is to modify the flannel plugin to accept an IPAM block like other plugins, and allow the 'routes' option in that block (but not the IP addressing stuff which flannel handles independently).

chenk008:
@dcbw I use "isDefaultGateway" inspired by this bridge plugin.When i use flannel with bright , i have set isDefaultGateway=true.
Maybe flannel also need support  ipam config , but now flannel is not support.The flannel will fail for "'delegate' dictionary must not have 'ipam' field, it'll be set by flannel"

PR #288, opened 2019-04-03
test: add coveralls support

dcbw:


dcbw:
So coveralls can't actually parse the data it's getting (or it's not getting any data) and I'm not sure what's going on there right now. But coveralls is at least aware of the builds and whatnot.

PR #291, opened 2019-04-09
[do not merge] CI breakage test

dcbw:
Only to test CI breakage with go 1.11.x by making a trivial change to ensure git master is actually broken in CI.


s1061123:
@dcbw I met it in multus-cni and `CGO_ENABLED=0` fixed it, FYI.

Issue #101, opened 2017-12-12
Plugins should include their build version


squeed:
Not knowing the exact plugin version makes debugging tricky, especially when it's not clear where the cni binaries come from.

The plugins should have some kind of a `--version` argument or equivalent.

lsm5:
atm, `git describe` on master branch tells me `v0.7.2-123-gb93d284` while there's already a v0.7.3 release. Not sure that's consistent.

lsm5:
@dcbw  ping ^

Issue #277, opened 2019-03-15
Review and address 'staticcheck' warnings


bboreham:
I ran `staticcheck`:

```
~/src/github.com/containernetworking/plugins$ staticcheck ./...
integration/integration_linux_test.go:152:16: should use raw string (`...`) with regexp.MustCompile to avoid having to escape twice (S1007)
pkg/ip/link_linux_test.go:36:9: should use String() instead of fmt.Sprintf (S1025)
pkg/ns/ns_linux.go:45:20: error strings should not be capitalized (ST1005)
pkg/ns/ns_linux.go:189:20: error strings should not be capitalized (ST1005)
pkg/utils/utils.go:25:2: const prefixLength is unused (U1000)
plugins/ipam/dhcp/client.go:9:2: const MaxDHCPLen is unused (U1000)
plugins/ipam/dhcp/client.go:38:6: func DhcpSendDecline is unused (U1000)
plugins/ipam/dhcp/daemon.go:36:7: const listenFdsStart is unused (U1000)
plugins/ipam/dhcp/daemon.go:156:25: error strings should not be capitalized (ST1005)
plugins/ipam/dhcp/daemon.go:168:21: error strings should not be capitalized (ST1005)
plugins/ipam/dhcp/daemon.go:171:21: error strings should not be capitalized (ST1005)
plugins/ipam/dhcp/daemon.go:177:20: error strings should not be capitalized (ST1005)
plugins/ipam/dhcp/lease.go:191:15: should use time.Until instead of t.Sub(time.Now()) (S1024)
plugins/ipam/dhcp/main.go:50:15: printf-style function with dynamic first argument and no further arguments should use print-style function instead (SA1006)
plugins/ipam/dhcp/options.go:101:10: should omit second index in slice, s[a:len(s)] is identical to s[a:] (S1010)
plugins/ipam/host-local/backend/allocator/config.go:100:11: should omit value from range; this loop is equivalent to `for idx := range ...` (S1005)
plugins/ipam/host-local/backend/allocator/config.go:125:9: should omit value from range; this loop is equivalent to `for i := range ...` (S1005)
plugins/ipam/host-local/backend/allocator/range.go:36:20: error strings should not be capitalized (ST1005)
plugins/ipam/host-local/backend/allocator/range.go:46:20: error strings should not be capitalized (ST1005)
plugins/ipam/host-local/backend/allocator/range_set.go:64:9: should omit value from range; this loop is equivalent to `for i := range ...` (S1005)
[...]

PR #295, opened 2019-04-15
plugins: correctly output build version, cosmetic cleanups

squeed:
Now that libcni has the ability to print a version message, plumb it through correctly.
    
While we're at it,
- fix import paths
- run gofmt
- add some more comments to sample
- add container runtime swappability for release

Fixes: #101


bboreham:
Fixes one of the points at #277

PR #298, opened 2019-04-17
macvlan: make master config as optional

s1061123:
To ease of config, make master config as optional parameter.
In default case, default route interface will be used for master.

### DIFF
--- a/plugins/main/macvlan/macvlan.go
+++ b/plugins/main/macvlan/macvlan.go
@@ -51,13 +51,33 @@ func init() {
 	runtime.LockOSThread()
 }
 
+func getDefaultRouteInterfaceName() string {
+	routeToDstIP, err := netlink.RouteList(nil, netlink.FAMILY_ALL)
+	if err != nil {
+		return ""
+	}
+
+	for _, v := range routeToDstIP {
+		l, _ := netlink.LinkByIndex(v.LinkIndex)
+		if v.Dst == nil {
+			return l.Attrs().Name
+		}
+	}
+
+	return ""
+}
+
 func loadConf(bytes []byte) (*NetConf, string, error) {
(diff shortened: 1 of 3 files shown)


plugins/main/macvlan/macvlan.go
```diff
@@ -51,13 +51,33 @@ func init() {
...
 }
 
+func getDefaultRouteInterfaceName() string {
+	routeToDstIP, err := netlink.RouteList(nil, netlink.FAMILY_ALL)
+	if err != nil {
+		return ""
+	}
+
+	for _, v := range routeToDstIP {
+		l, _ := netlink.LinkByIndex(v.LinkIndex)
```
matthewdupre: This could be moved inside the if.

plugins/main/macvlan/macvlan.go
```diff
@@ -51,13 +51,33 @@ func init() {
...
+		if v.Dst == nil {
+			return l.Attrs().Name
+		}
+	}
+
+	return ""
```
matthewdupre: Would be nice to pass the errors up: the caller gives a generic one: would be nice to know whether there was no default route, or whether netlink gave an error etc.

plugins/main/macvlan/macvlan.go
matthewdupre: Also nice to capture any error from here.

plugins/main/macvlan/macvlan.go
s1061123: thank you for your comments. I changed it.

s1061123:
@matthewdupre could you please take a look into it?

dcbw:
Updated LGTM

squeed:
lgtm

PR #300, opened 2019-04-18
host-device: add pciBusID property

### LINKED ISSUE #253: host-device: specify device with pci address
Currently host-device plugin supports specifying devices with three properties: ``device``, ``hwaddr``, ``kernelpath``. This is to request another property called ``pciaddr``.

Benefit of having ``pciaddr`` support is when using host-device along with other metaPlugins such as Multus. It can parse and config the device with given ``pciaddr`` that passed from Multus (Multus has the ability to retrieve devices allocated from kubelet device plugin and pass ``pciaddr`` of that device to delegated plugins).

zshi-redhat:
Fixes #253

### DIFF
--- a/plugins/main/host-device/README.md
+++ b/plugins/main/host-device/README.md
@@ -1,22 +1,40 @@
 # host-device
+
 Move an already-existing device into a container.
 
-This simple plugin will move the requested device from the host's network namespace
-to the container's. Nothing else will be done - no IPAM, no addresses.
+## Overview
 
-The device can be specified with any one of three properties:
+This simple plugin will move the requested device from the host's network namespace
+to the container's. IPAM configuration can be used for this plugin.
+
+## Network configuration reference
+
+The device can be specified with any one of four properties:
 * `device`: The device name, e.g. `eth0`, `can0`
 * `hwaddr`: A MAC address
 * `kernelpath`: The kernel device kobj, e.g. `/sys/devices/pci0000:00/0000:00:1f.6`
+* `pciBusID`: A PCI address of network device, e.g `0000:00:1f.6`
 
 For this plugin, `CNI_IFNAME` will be ignored. Upon DEL, the device will be moved back.
 
-A sample configuration might look like:
+## Example configuration
+
+A sample configuration with `device` property looks like:
 
 ```json
 {
 	"cniVersion": "0.3.1",
 	"type": "host-device",
 	"device": "enp0s1"
+}
+```
+
+A sample configuration with `pciBusID` property looks like:
+
+```json
+{
+	"cniVersion": "0.3.1",
+	"type": "host-device",
+	"pciBusID": "0000:3d:00.1"
 }
 ```
--- a/plugins/main/host-device/host-device.go
+++ b/plugins/main/host-device/host-device.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net"
+	"os"
 	"path/filepath"
 	"runtime"
 	"strings"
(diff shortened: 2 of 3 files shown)


plugins/main/host-device/host-device.go
```diff
@@ -38,12 +39,17 @@ import (
 	bv "github.com/containernetworking/plugins/pkg/utils/buildversion"
 )
 
+const (
+	sysBusPci = "/sys/bus/pci/devices"
```
jellonek: PTAL on https://github.com/golang/go/wiki/CodeReviewComments#initialisms - that should be named as `sysBusPCI`

plugins/main/host-device/host-device.go
```diff
@@ -278,6 +284,16 @@ func getLink(devname, hwaddr, kernelpath string) (netlink.Link, error) {
...
+	} else if len(pciaddr) > 0 {
+		netDir := filepath.Join(sysBusPci, pciaddr, "net")
+		if _, err := os.Lstat(netDir); err != nil {
+			return nil, fmt.Errorf("no net directory under pci device %s: %q", pciaddr, err)
+		}
+		fInfo, err := ioutil.ReadDir(netDir)
+		if err != nil {
+			return nil, fmt.Errorf("failed to read net directory %s: %q", netDir, err)
+		}
+		return netlink.LinkByName(fInfo[0].Name())
```
jellonek: There is no check if this list has at least single entry. I'm not sure if this directory can be empty, but if it's possible - code will panic there. If you are sure that it has to have at least a single entry - just leave a comment about that there in code.

plugins/main/host-device/host-device.go
zshi-redhat: fixed

plugins/main/host-device/host-device.go
zshi-redhat: added check for ``fInfo`` and return netlink by name when ``fInfo`` list larger than 0.

plugins/main/host-device/README.md
```diff
@@ -1,17 +1,25 @@
...
+to the container's. IPAM configuration can be used for this plugin.
+
+## Network configuration reference
 
-The device can be specified with any one of three properties:
+The device can be specified with any one of four properties:
 * `device`: The device name, e.g. `eth0`, `can0`
 * `hwaddr`: A MAC address
 * `kernelpath`: The kernel device kobj, e.g. `/sys/devices/pci0000:00/0000:00:1f.6`
+* `deviceID`: A PCI address of network device, e.g `0000:00:1f.6`
```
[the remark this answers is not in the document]
zshi-redhat: @s1061123  I didn't mark the deviceID as optional because it's similar to other existing options, such as hwaddr, device and kernelpath; at least one of these four options need to be provided.
Re the name of ``deviceID``, I give a second thought on it and maybe it should not be changed as that's the name used by device plugin [framework](https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/apis/deviceplugin/v1beta1/api.proto#L85) which in some cases might be referred with a different meaning other than PCI.

zshi-redhat:
@squeed @dcbw PTAL?

plugins/main/host-device/README.md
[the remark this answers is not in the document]
s1061123: In [above example](https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/apis/deviceplugin/v1beta1/api.proto#L77-L80), deviceID in k8s sometimes uses other than PCI address. So user may confuse if we name 'PCIBusID' as 'deviceID'. This plugin, `host-local` plugin is for host local device, not only for PCI also USB (i.e. USB network adapter) as well as other bus, so we should give appropriate name for that.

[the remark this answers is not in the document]
zshi-redhat: Ok, fair enough, updated a new version with ``pciBusID``.

PR #301, opened 2019-04-18
fix bug on ip revert if cmdAdd fails on macvlan and host-device

mars1024:
This PR try to fix bug on IP revert if cmdAdd fails on macvlan and host-device plugin.
Function **ipam.ExecDel** will check whether CNI_COMMAND environment variable is **ADD**, if not, it will not work as expected. So when we want to revert IP allocation in cmdAdd with environment variable CNI_COMMAND equals to ADD, we have to change CNI_COMMAND to DEL temporarily, otherwise IP release will not succeed.

### DIFF
--- a/plugins/main/host-device/host-device.go
+++ b/plugins/main/host-device/host-device.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net"
+	"os"
 	"path/filepath"
 	"runtime"
 	"strings"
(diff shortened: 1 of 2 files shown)


dcbw:
Seems OK to me, but I wonder if we should just make ipam.ExecDel() (well, and add/check/etc) do this automatically... the intent of ExecDel is pretty clear.

squeed:
We should clearly fix ExecDel, but this LGTM.

PR #307, opened 2019-04-28
Allow to configure empty ipam for macvlan

SchSeba:
This PR add the option to configure an empty ipam for the macvlan cni plugin.

When using the macvlan cni plugin with an empty ipam the requeted pod will get the macvlan interface but without any ip address.

One of the use cases for this feature is for projects that runs a dhcp server inside the pod like KubeVirt.
In KubeVirt we need to let the vm running inside the pod to make the dhcp request so it will be able to make a release an renew request when needed.

### DIFF
--- a/plugins/main/macvlan/macvlan.go
+++ b/plugins/main/macvlan/macvlan.go
@@ -169,6 +169,8 @@ func cmdAdd(args *skel.CmdArgs) error {
 		return err
 	}
 
+	isLayer3 := n.IPAM.Type != ""
+
 	netns, err := ns.GetNS(args.Netns)
 	if err != nil {
 		return fmt.Errorf("failed to open netns %q: %v", netns, err)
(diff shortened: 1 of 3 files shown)


plugins/main/macvlan/macvlan.go
```diff
@@ -189,56 +191,85 @@ func cmdAdd(args *skel.CmdArgs) error {
...
+			return nil
+		})
 		if err != nil {
-			return fmt.Errorf("failed to look up %q: %v", args.IfName, err)
+			return err
 		}
+	} else {
+		// For L2 just change interface status to up
+		err = netns.Do(func(_ ns.NetNS) error {
+			contVeth, err := net.InterfaceByName(args.IfName)
```
bboreham: Why do we need this lookup?

SchSeba: @bboreham Thanks for the review!

Just like we use it inside the L3 section here https://github.com/containernetworking/plugins/pull/307/files#diff-36359cba73159218a43a3afbafbb7f40R236

I need to find the interface the set it up.

bboreham: How do you expect `contVeth.Name` differs from `args.IfName`?

SchSeba: It should be the same I just want to preserve the same logic that was already implemented https://github.com/containernetworking/plugins/blob/master/plugins/main/macvlan/macvlan.go#L173

bboreham: I am totally failing to see the similarity.  The only other place `net.InterfaceByName()` is called in this file is where the result is actually needed, to pass to `GratuitousArpOverIface()`.

SchSeba: Oh I miss that you are compliantly right thanks for the comment! I remove that from the PR

PR #309, opened 2019-04-29
Flannel: Pass through runtimeConfig to delegate

nagiesek:
Currently we don't pass through the runtimeConfig from flannel down to the delegate call, I think it would be a good idea to do so.


nagiesek:
check for existence of runtimeConfig to fix test failure.

benmoss:
LGTM :+1:

PR #318, opened 2019-05-12
host-local: fix a bug about error assignment

mars1024:
Just as the title, I think host-local try to judge an error which can not be read forever because of a wrong assignment.

### DIFF
--- a/plugins/ipam/host-local/backend/disk/backend.go
+++ b/plugins/ipam/host-local/backend/disk/backend.go
@@ -105,7 +105,8 @@ func (s *Store) FindByKey(id string, ifname string, match string) (bool, error) 
 		if err != nil || info.IsDir() {
 			return nil
 		}
-		data, err := ioutil.ReadFile(path)
+		var data []byte
+		data, err = ioutil.ReadFile(path)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


squeed:
This shouldn't be a syntax error. Is this a `go lint` error? Unfortunately, `go lint` doesn't follow current conventions around error assignment.

mars1024:
> This shouldn't be a syntax error. Is this a `go lint` error? Unfortunately, `go lint` doesn't follow current conventions around error assignment.

Anyway, I think it's an error but not a warning.  If we try to assign **err** by `data, err := ioutil.ReadFile(path)`, this **err** will never be used beside this line, the **err** in next line `if err != nil ` is no longer the **err** we expected because we have an **err** in function input parameters `func(path string, info os.FileInfo, err error)` .

mars1024:
force-push for retest

bboreham:
What are you trying to achieve here?  I can't see what difference your changes make.
If it's a bug, can you construct a test that fails under the old code?

mars1024:
I think this is a mistake of IDE, I've tested with some cases, the code is working properly. Thanks for your comments. I'll close this.

Issue #260, opened 2019-02-13
bandwidth should be more careful about finding host-side veth pair


dcbw:
Bandwidth currently does this in getHostInterface():

```
	var err error
	for _, iface := range interfaces {
		if iface.Sandbox == "" { // host interface
			_, _, err = ip.GetVethPeerIfindex(iface.Name)
			if err == nil {
				return iface, err
			}
		}
	}
```

It should instead look for the sandbox interface, run GetVethPeerIfindex() on that inside the container namespace to get the ifindex of the host side, and then look up the host-side ifname using that ifindex.

The problem is that some plugins (eg https://github.com/containernetworking/plugins/pull/231) may return multiple host-side veth interfaces, only one of which is the container veth. Bandwidth doesn't filter the interfaces enough to figure that out.

mars1024:
try to fix in #321

PR #321, opened 2019-05-18
bandwidth: get bandwidth interface in host ns through container interface

mars1024:
try to fix [https://github.com/containernetworking/plugins/issues/260](https://github.com/containernetworking/plugins/issues/260)

We should find the interface in host ns which bandwidth plugin works on through container interface instead of host interfaces, because host interfaces may contain multiple host-side veth interfaces, it's hard for us to judge.

### DIFF
--- a/plugins/meta/bandwidth/bandwidth_linux_test.go
+++ b/plugins/meta/bandwidth/bandwidth_linux_test.go
@@ -621,6 +621,257 @@ var _ = Describe("bandwidth test", func() {
 		})
 	})
 
+	Describe("Getting the host interface which plugin should work on from veth peer of container interface", func() {
+		It("Should work with multiple host veth interfaces", func() {
+			conf := `{
+	"cniVersion": "0.4.0",
+	"name": "cni-plugin-bandwidth-test",
+	"type": "bandwidth",
+	"ingressRate": 8,
+	"ingressBurst": 8,
+	"egressRate": 16,
+	"egressBurst": 8,
+	"prevResult": {
+		"interfaces": [
+			{
+				"name": "%s",
+				"sandbox": ""
+			},
+			{
+				"name": "%s",
+				"sandbox": ""
+			},
+			{
+				"name": "%s",
+				"sandbox": ""
+			},
+			{
+				"name": "%s",
+				"sandbox": "%s"
+			}
+		],
+		"ips": [
+			{
+				"version": "4",
+				"address": "%s/24",
+				"gateway": "10.0.0.1",
+				"interface": 1
+			}
+		],
+		"routes": []
+	}
+}`
+
+			// create veth peer in host ns
+			vethName, peerName := "host-veth-peer1", "host-veth-peer2"
+			createVethInOneNs(hostNs.Path(), vethName, peerName)
+
+			conf = fmt.Sprintf(conf, vethName, peerName, hostIfname, containerIfname, containerNs.Path(), containerIP.String())
+			args := &skel.CmdArgs{
+				ContainerID: "dummy",
+				Netns:       containerNs.Path(),
+				IfName:      containerIfname,
+				StdinData:   []byte(conf),
+			}
+
+			Expect(hostNs.Do(func(netNS ns.NetNS) error {
+				defer GinkgoRecover()
+				r, out, err := testutils.CmdAdd(containerNs.Path(), args.ContainerID, "", []byte(conf), func() error { return cmdAdd(args) })
+				Expect(err).NotTo(HaveOccurred(), string(out))
+				result, err := current.GetResult(r)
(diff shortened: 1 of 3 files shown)


mars1024:
> LGTM but can you add a test case which will add on host 2 veth pairs, with only second one having a peer in some container, for which `getHostInterface` should be tested?

Sure. I'll do it later.

mars1024:
Updated with some test cases. cc @jellonek

plugins/meta/bandwidth/bandwidth_linux_test.go
```diff
@@ -621,6 +621,257 @@ var _ = Describe("bandwidth test", func() {
 		})
 	})
 
+	Describe("get the host interface which plugin should work on", func() {
```
jellonek: Please look how `Describe` messages are used in other test cases, like how are used in `It`. It could be something like:
```go
Describe("Getting host interface from veth pair"), func() {
    It("Should work with multiple host veth interfaces", ...
```

plugins/meta/bandwidth/bandwidth_linux_test.go
```diff
@@ -621,6 +621,257 @@ var _ = Describe("bandwidth test", func() {
...
+
+				Expect(qdiscs[0]).To(BeAssignableToTypeOf(&netlink.Tbf{}))
+				Expect(qdiscs[0].(*netlink.Tbf).Rate).To(Equal(uint64(1)))
+				Expect(qdiscs[0].(*netlink.Tbf).Limit).To(Equal(uint32(1)))
+				return nil
+			})).To(Succeed())
+
+		})
+
+		It("container interface has no veth peer", func() {
```
jellonek: ditto

plugins/meta/bandwidth/bandwidth_linux_test.go
```diff
@@ -621,6 +621,257 @@ var _ = Describe("bandwidth test", func() {
...
+				defer GinkgoRecover()
+
+				_, _, err := testutils.CmdAdd(containerNs.Path(), args.ContainerID, "", []byte(conf), func() error { return cmdAdd(args) })
+				Expect(err).To(HaveOccurred())
+
+				return nil
+			})).To(Succeed())
+		})
+
+		It("no interfaces", func() {
```
jellonek: ditto

plugins/meta/bandwidth/bandwidth_linux_test.go
```diff
@@ -621,6 +621,257 @@ var _ = Describe("bandwidth test", func() {
...
+				defer GinkgoRecover()
+
+				_, _, err := testutils.CmdAdd(containerNs.Path(), args.ContainerID, "", []byte(conf), func() error { return cmdAdd(args) })
+				Expect(err).To(HaveOccurred())
+
+				return nil
+			})).To(Succeed())
+		})
+
+		It("veth peer of container interface does not match any of host interfaces", func() {
```
jellonek: ditto

plugins/meta/bandwidth/bandwidth_linux_test.go
mars1024: Thanks for review, updated.

mars1024:
updated with some descriptions of test cases.

squeed:
this is awesome!
lgtm

czybjtu:
hi all. I use bandwith plugin with cilium in my test env. I find that `bandwith` plugin will always return error: `no veth peer of container interface found in host ns`. 

After reivewing src code, I find it that `iface.Sandbox` will always not empty and `iface.Name` is `eth0` and `link.Attrs().Name` is `lxcxxxxxx` which is the interface name in host side. So I'm curious about why we need this logic? Or did I misunderstand something?

https://github.com/containernetworking/plugins/blob/4a6147a1552064af80b4f7567b30c5174153c62a/plugins/meta/bandwidth/main.go#L153-L159

czybjtu:
here is my test code
```diff
+       log.Println("interfaces length:", len(interfaces))
        for _, iface := range interfaces {
+               log.Println("iface name:", iface.Name)
+               log.Println("sandbox: ", iface.Sandbox)
+               log.Println("link.Attrs().Name: ", link.Attrs().Name)
                if iface.Sandbox == "" && iface.Name == link.Attrs().Name {
                        return iface, nil
                }
        }
```

log 
```
2023/04/13 07:20:51 interfaces length: 1
2023/04/13 07:20:51 iface name: eth0
2023/04/13 07:20:51 sandbox:  /var/run/netns/cni-882cb34f-9dd9-6014-556b-724e5d2e383d
2023/04/13 07:20:51 link.Attrs().Name:  lxc1986c9428372
```

czybjtu:
This is a problem with cilium. Sorry for the inconvenience.

PR #325, opened 2019-05-26
bridge: fix ipMasq setup to use correct source address

nickethier:
It looks like a change in #279 which seems to expect the exact address be passed into `SetupIPMasq` broke ip masquerading for the bridge plugin. See discussion [in that PR](https://github.com/containernetworking/plugins/pull/279#issuecomment-495828664) for more details.

This PR adds a ginkgo spec that reproduces the issue and a change to fix it.

### DIFF
--- a/plugins/main/bridge/bridge_test.go
+++ b/plugins/main/bridge/bridge_test.go
@@ -17,11 +17,13 @@ package main
 import (
 	"encoding/json"
 	"fmt"
-	"github.com/vishvananda/netlink/nl"
 	"io/ioutil"
 	"net"
 	"os"
 	"strings"
+
+	"github.com/coreos/go-iptables/iptables"
+	"github.com/vishvananda/netlink/nl"
 
 	"github.com/containernetworking/cni/pkg/skel"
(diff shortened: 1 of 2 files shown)


plugins/main/bridge/bridge_test.go
```diff
@@ -178,6 +186,11 @@ func (tc testCase) subnetConfig() string {
 	return conf
 }
 
+func (tc testCase) ipMasqConfig() string {
+	conf := fmt.Sprintf(ipMasqConfStr, tc.ipMasq)
+	return conf
```
mars1024: I think you can just `return fmt.Sprintf(ipMasqConfStr, tc.ipMasq)` here

mars1024:
ref [https://github.com/containernetworking/plugins/pull/279#issuecomment-495847441](https://github.com/containernetworking/plugins/pull/279#issuecomment-495847441)

mars1024:
lgtm

mars1024:
I think this need more maintainers to take a look at :)

squeed:
Good catch! LGTM

Issue #327, opened 2019-06-01
host-local should support idempotent allocation


mars1024:
I think host-local plugin should support idempotent allocation, which means that if we try to request an IP with same container ID several times, the allocator should return the same IP at any time unless we free it.

This will help to avoid **ip leak**. Specially in this scenario, [CRI supports set up the same sandbox container again and again if it exists and its network has not been configured successfully](https://github.com/alibaba/pouch/blob/29753d7d218380b3459ed9e607e1898492bfedab/cri/v1alpha2/cri.go#L363).

squeed:
Hmm. The spec, as written, says that runtimes can only call ADD once. That was part of the motivation for creating the CHECK action, so they could see that networking was still up.

I don't think we want to change the spec to allow idempotency. This is mostly because it's difficult to imagine how certain plugins could implement all of their functionality - along with chained plugins - in an idempotent way.

For example, the `sbr` chained-plugin deletes the routes created by, say, `ptp`. If we were to re-run ADD, then the routing table would be either broken or misconfigured as the process proceeded.

So, I don't think there's a safe-way to re-ADD an existing container.

mars1024:
Thanks for you comment, @squeed , I still think idempotency should be one of basic abilities which an IPAM server should have, it does not sound reasonable for allocator to allocate different IPs for several request with same sandbox ID.
I also think re-ADD is not safe, but it's, maybe, CRI's job to deal with it with some special operation like stopping sandbox container ? Anyway, I don't really care about it. 
What I care about is making `host-local` plugin more robust in some extreme cases like above, and I think this change does not affect any existing behavior or usage, it's more like a nice-to-have feature to make `host-local` a real IPAM server.

dcbw:
@mars1024 the CNI spec requires that ADD only be called once per container. Thus there is no possibility that ADD will be called twice for the same container ID, ifname, and network name.

jellonek:
That also means the spec does not handle ADD/DEL/ADD sequence and for sure at least telecoms need such functionality like dynamic reconfiguration without killing of the container.

mars1024:
Just as I talked above, without regard of the reasonable or unreasonable usage scenarios, what I want to do is just making `host-local` more robust in any possible scenario with minimal cost. I  think it's an **enhancement** rather than a **fix**.
If the issue is not accepted after reconsideration, I'll close this ~
Please pay another attention to this issue and PR in maintainer meeting of this week, thanks ~ @dcbw  @squeed

[...]

PR #328, opened 2019-06-01
host-local support idempotent allocation

mars1024:
try to fix [https://github.com/containernetworking/plugins/issues/327](https://github.com/containernetworking/plugins/issues/327)

### DIFF
--- a/plugins/ipam/host-local/backend/allocator/allocator.go
+++ b/plugins/ipam/host-local/backend/allocator/allocator.go
@@ -40,7 +40,7 @@ func NewIPAllocator(s *RangeSet, store backend.Store, id int) *IPAllocator {
 	}
 }
 
-// Get alocates an IP
+// Get allocates an IP
 func (a *IPAllocator) Get(id string, ifname string, requestedIP net.IP) (*current.IPConfig, error) {
 	a.store.Lock()
 	defer a.store.Unlock()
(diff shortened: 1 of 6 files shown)


dcbw:
Discussed this and we agree that this fixes a bug in host-local and is better than the current behavior. Could you update the plugin to instead return an error when an allocation exists for a given ID, since the spec mandates that plugins "must" not call ADD for the same tuple?

mars1024:
> @mars1024 Discussed this and we agree that this fixes a bug in host-local and is better than the current behavior. Could you update the plugin to instead return an error when an allocation exists for a given ID, since the spec mandates that plugins "must" not call ADD for the same tuple?

OK, I'll update this soon. :)

mars1024:
Updated!  cc @dcbw  
Now host-local will support idempotent allocation in this way, if a given ID has an allocated IP, the follow-up requests will be rejected until it release the allocated IP.

mars1024:
@dcbw @squeed @bboreham @rosenhouse If you have time, please take a look at this updated PR this week, thanks a lot ~

squeed:
lgtm, thanks!

mars1024:
@dcbw @bboreham Could you help to promote the updated PR this week?

dcbw:
LGTM

squeed:
Done! We expect to be tagging another release next week.

PR #331, opened 2019-06-04
Loopback dsr & L2Tunnel

nagiesek:
Two seperate items here. The first adds an optional feature to easily allow for loopback dsr. The second relaxes the network type requirement to allow l2tunnel networks to be used with the l2bridge plugin.

### DIFF
--- a/pkg/hns/netconf_windows.go
+++ b/pkg/hns/netconf_windows.go
@@ -17,6 +17,9 @@ package hns
 import (
 	"bytes"
 	"encoding/json"
+	"fmt"
+	"net"
+
 	"github.com/Microsoft/hcsshim/hcn"
 	"github.com/buger/jsonparser"
 	"github.com/containernetworking/cni/pkg/types"
@@ -26,9 +29,16 @@ import (
 // NetConf is the CNI spec
 type NetConf struct {
 	types.NetConf
+	// ApiVersion is either 1 or 2, which specifies which hns APIs to call
+	ApiVersion int `json:"ApiVersion"`
+	// V2 Api Policies
 	HcnPolicyArgs []hcn.EndpointPolicy `json:"HcnPolicyArgs,omitempty"`
-	Policies      []policy             `json:"policies,omitempty"`
-	RuntimeConfig RuntimeConfig        `json:"runtimeConfig"`
+	// V1 Api Policies
+	Policies []policy `json:"policies,omitempty"`
+	// Options to be passed in by the runtime
+	RuntimeConfig RuntimeConfig `json:"runtimeConfig"`
+	// If true, adds a policy to endpoints to support loopback direct server return
+	LoopbackDSR bool `json:"loopbackDSR"`
 }
 
 type RuntimeDNS struct {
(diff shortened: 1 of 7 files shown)


pkg/hns/netconf_windows.go
```diff
@@ -45,6 +54,37 @@ type policy struct {
...
+func GetDefaultDestinationPrefix(ip *net.IP) string {
+	destinationPrefix := "0.0.0.0/0"
+	if ipv6 := ip.To4(); ipv6 == nil {
+		destinationPrefix = "::/0"
+	}
+	return destinationPrefix
+}
+
+func (n *NetConf) applyLoopbackDSR(ip *net.IP) {
+	value := fmt.Sprintf(`"Destinations" : ["%s"]`, ip.String())
```
jellonek: Please use `%q` instead of `"%s"`.

squeed: hmm, probably not necessary in this case since it's just building json.

pkg/hns/netconf_windows.go
```diff
@@ -26,9 +29,15 @@ import (
 // NetConf is the CNI spec
 type NetConf struct {
 	types.NetConf
+	ApiVersion    int                  `json:"ApiVersion"`
 	HcnPolicyArgs []hcn.EndpointPolicy `json:"HcnPolicyArgs,omitempty"`
 	Policies      []policy             `json:"policies,omitempty"`
 	RuntimeConfig RuntimeConfig        `json:"runtimeConfig"`
+	OptionalFlags OptionalFlags        `json:"optionalFlags"`
```
squeed: Is there any reason to push this down a level, configuration-wise? It seems LoopbackDSR could just be on this level?

nagiesek: No good ones, I'll change it back.

plugins/main/windows/win-bridge/win-bridge_windows.go
```diff
@@ -103,7 +103,7 @@ func cmdHnsAdd(args *skel.CmdArgs, n *NetConf) (*current.Result, error) {
 		return nil, fmt.Errorf("network %v not found", networkName)
 	}
 
-	if !strings.EqualFold(hnsNetwork.Type, "L2Bridge") {
+	if !strings.EqualFold(hnsNetwork.Type, "L2Bridge") || !strings.EqualFold(hnsNetwork.Type, "L2Tunnel") {
```
lzang: Has a similar check in the function cmdHcnAdd. Should that be changed as well?

nagiesek: Good Catch.

pkg/hns/netconf_windows.go
```diff
@@ -26,9 +29,11 @@ import (
 // NetConf is the CNI spec
 type NetConf struct {
 	types.NetConf
+	ApiVersion    int                  `json:"ApiVersion"`
 	HcnPolicyArgs []hcn.EndpointPolicy `json:"HcnPolicyArgs,omitempty"`
 	Policies      []policy             `json:"policies,omitempty"`
 	RuntimeConfig RuntimeConfig        `json:"runtimeConfig"`
+	LoopbackDSR   bool                 `json:"loopbackDSR"`
```
dcbw: Can you add some code comments to describe what the fields here are/do?

squeed:
Seems fine now. Can you add an entry in to the plugin's README (and squash your commits). Then we can merge.

dcbw:
@nagiesek can you rebase to fix the conflict? Then we can merge. Thanks!

nagiesek:
@dcbw

squeed:
awesome, looks great!

PR #336, opened 2019-06-17
Unmount and remove after closing ns

smarkm:


squeed:
I don't think we want to do this - we want the namespace to stick around after we've set up the networking :-)

smarkm:
Yes, maybe just remove by each test suit file when running test suit

Issue #316, opened 2019-05-08
Switch to go modules


squeed:
Godeps has served us well. Time to get with the program.

nagiesek:
If no one is working on this, I can take this up.

squeed:
Go for it!

PR #337, opened 2019-06-18
Go mod update

nagiesek:


nagiesek:
#316

squeed:
Argh, looks like go mod moved the ethtool dependency backwards in time...

PR #344, opened 2019-06-21
Make host-side veth name configurable. 

cadmuxe:
update pkg/ip/link_linux.go:SetupVeth to support host-side veth name config.
GKE will have an internal plugin which requires customized host-side veth name, we would like to leverage this lib.

### DIFF
--- a/pkg/ip/link_linux.go
+++ b/pkg/ip/link_linux.go
@@ -60,11 +60,15 @@ func peerExists(name string) bool {
 	return true
 }
 
-func makeVeth(name string, mtu int) (peerName string, veth netlink.Link, err error) {
+func makeVeth(name, vethPeerName string, mtu int) (peerName string, veth netlink.Link, err error) {
 	for i := 0; i < 10; i++ {
-		peerName, err = RandomVethName()
-		if err != nil {
-			return
+		if vethPeerName != "" {
+			peerName = vethPeerName
+		} else {
+			peerName, err = RandomVethName()
+			if err != nil {
+				return
+			}
 		}
 
 		veth, err = makeVethPair(name, peerName, mtu)
@@ -124,9 +128,10 @@ func ifaceFromNetlinkLink(l netlink.Link) net.Interface {
 // SetupVeth sets up a pair of virtual ethernet devices.
 // Call SetupVeth from inside the container netns.  It will create both veth
 // devices and move the host-side veth into the provided hostNS namespace.
+// hostVethName: If hostVethName is not specified, the host-side veth name will use a random string.
 // On success, SetupVeth returns (hostVeth, containerVeth, nil)
-func SetupVeth(contVethName string, mtu int, hostNS ns.NetNS) (net.Interface, net.Interface, error) {
-	hostVethName, contVeth, err := makeVeth(contVethName, mtu)
+func SetupVeth(contVethName, hostVethName string, mtu int, hostNS ns.NetNS) (net.Interface, net.Interface, error) {
+	hostVethName, contVeth, err := makeVeth(contVethName, hostVethName, mtu)
 	if err != nil {
 		return net.Interface{}, net.Interface{}, err
 	}
--- a/pkg/ip/link_linux.go
+++ b/pkg/ip/link_linux.go
@@ -77,7 +77,7 @@ func makeVeth(name, vethPeerName string, mtu int) (peerName string, veth netlink
 			return
 
 		case os.IsExist(err):
-			if peerExists(peerName) {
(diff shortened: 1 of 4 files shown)


cadmuxe:
@lzang

pkg/ip/link_linux.go
```diff
@@ -60,11 +60,15 @@ func peerExists(name string) bool {
 	return true
 }
 
-func makeVeth(name string, mtu int) (peerName string, veth netlink.Link, err error) {
+func makeVeth(name, vethPeerName string, mtu int) (peerName string, veth netlink.Link, err error) {
```
mars1024: I don't think veth-peer name in host-side can be configurable without some checks, because interface name actually has some constraints, like **length** and **uniqueness**. It's dangerous to put this as an input parameter. I think random name is a safe and general option for us.

jellonek: Exactly as you described that - this randomness should be an option, but having a possibility to set a particular value, as its provided by kernel - is also an option and having this on the lib side (e.g. to enforce some naming scheme) seems to be very valuable.
On the same time I think that validation of these constraints should not be done on lib level. These rules will be enforced by kernel, so if user will provide incorrect values - he will simply receive an error.

btw. what exactly do you see dangerous in leaving such option for an user?

cadmuxe: Thank you for reviewing, like Piotr said, I think we can rely on kernel. Actually, the uniqueness of current logic is enforced by kernel.

cadmuxe:
> It seems that this code does not make any **real** difference. Could you describe the reason for this PR? Thanks ~

Hi Bruce, Thank you for reviewing this PR. GKE has an internal plugin which need the customized host-side veth, we would like to use this lib. I have updated the description. Thanks.

pkg/ip/link_linux.go
```diff
@@ -60,11 +60,15 @@ func peerExists(name string) bool {
...
 	for i := 0; i < 10; i++ {
```
mars1024: If `vethPeerName` is specified, the retry here will no longer be needed.

mars1024:
> > It seems that this code does not make any **real** difference. Could you describe the reason for this PR? Thanks ~
> 
> Hi Bruce, Thank you for reviewing this PR. GKE has an internal plugin which need the customized host-side veth, we would like to use this lib. I have updated the description. Thanks.

Thanks, I've got your point. But what if bringing a new function called `SetupVethWithName` instead of just changing the origin function, maybe it's a better way.

pkg/ip/link_linux.go
cadmuxe: yeah, you are right. I have made it skip the retry if the vethPeerName is set.

cadmuxe:
> > > It seems that this code does not make any **real** difference. Could you describe the reason for this PR? Thanks ~
> > 
> > 
> > Hi Bruce, Thank you for reviewing this PR. GKE has an internal plugin which need the customized host-side veth, we would like to use this lib. I have updated the description. Thanks.
> 
> Thanks, I've got your point. But what if bringing a new function called `SetupVethWithName` instead of just changing the origin function, maybe it's a better way.

I think to reuse the same code as much as possible will make the maintain easier. And the change is small, make a big chunk of duplicates probably not worth it.

pkg/ip/link_linux.go
```diff
@@ -73,7 +77,7 @@ func makeVeth(name string, mtu int) (peerName string, veth netlink.Link, err err
 			return
 
 		case os.IsExist(err):
-			if peerExists(peerName) {
+			if peerExists(peerName) && vethPeerName == "" {
```
cadmuxe: Only when vethPeerName was empty, it would go with the current retry logic. @mars1024

mars1024: Looks good ~

bboreham:
> > But what if bringing a new function called `SetupVethWithName` instead of just changing the origin function, maybe it's a better way.

> I think to reuse the same code as much as possible will make the maintain easier. And the change is small, make a big chunk of duplicates probably not worth it.

You can have the old function call the new one.  This will add ~4 lines and save everyone calling this library some edits.

cadmuxe:
> > > But what if bringing a new function called `SetupVethWithName` instead of just changing the origin function, maybe it's a better way.
> 
> > I think to reuse the same code as much as possible will make the maintain easier. And the change is small, make a big chunk of duplicates probably not worth it.
> 
> You can have the old function call the new one. This will add ~4 lines and save everyone calling this library some edits.

Thank you for reviewing, good idea, I have updated the pr.

squeed:
LGTM

PR #346, opened 2019-06-26
replace context.TODO with timeout context

### LINKED ISSUE #228: Clear up context.TODO() calls
Some calls to `context.TODO()` were added in #227: we need to decide what context they should be, and use that one instead.

mars1024:
try to resolve #228

### DIFF
--- a/pkg/ipam/ipam.go
+++ b/pkg/ipam/ipam.go
@@ -16,18 +16,51 @@ package ipam
 
 import (
 	"context"
+	"time"
+
 	"github.com/containernetworking/cni/pkg/invoke"
 	"github.com/containernetworking/cni/pkg/types"
 )
 
+const (
+	delegateTimeout = 30 * time.Second
+)
+
+// ExecAdd delegates ADD action of CNI plugin
 func ExecAdd(plugin string, netconf []byte) (types.Result, error) {
-	return invoke.DelegateAdd(context.TODO(), plugin, netconf, nil)
+	ctx, cancel := context.WithTimeout(context.Background(), delegateTimeout)
+	defer cancel()
+
+	return ExecAddWithContext(ctx, plugin, netconf)
 }
 
+// ExecCheck delegates CHECK action of CNI plugin
 func ExecCheck(plugin string, netconf []byte) error {
-	return invoke.DelegateCheck(context.TODO(), plugin, netconf, nil)
+	ctx, cancel := context.WithTimeout(context.Background(), delegateTimeout)
+	defer cancel()
+
+	return ExecCheckWithContext(ctx, plugin, netconf)
 }
 
+// ExecDel delegates DEL action of CNI plugin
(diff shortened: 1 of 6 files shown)


mars1024:
@bboreham we can have a further discussion here, thanks ~

bboreham:
What is the use-case for this timeout?  Why doesn't the top-level caller time out and kill the process?

mars1024:
> What is the use-case for this timeout? Why doesn't the top-level caller time out and kill the process?

Thanks for your comment, @bboreham .
I think, for example, `ExecAdd` has not a context input but it has to call `invoke.DelegateAdd` with a context, `ExecAdd` is a top-level caller to `invoke.DelegateAdd` indeed, we should make sure that if `ExecAdd` returns, all goroutines which belongs to this should receive the termination signal and exit rather than waiting for the process's exit.
Moreover, i don't think we have a proper parent context passed from CNI users (maybe CRI), if `ExecAdd` is hung, the process may be hung either, this maybe a use-case for this timeout.
And I also think we can add a timeout-context framework to all CNI plugins, this can make all functions which needs context to have a parent context, we can control the cancel of context (with timeout or signal) in a uniform way.

mars1024:
Maybe we can add a default context to `cni/pkg/skel` which can be passed to cmdAdd, cmdCheck and cmdDel.

mars1024:
A user-case [https://github.com/containernetworking/cni/issues/567](https://github.com/containernetworking/cni/issues/567) and [https://github.com/containernetworking/cni/pull/568](https://github.com/containernetworking/cni/pull/568)  @bboreham

bboreham:
We've fixed the issue of not being able to time out the top-level call - that's in libcni.
After the fix, if the top-level call times out it will kill the plugin process.  This should kill any child processes (we may need to do something to ensure that).

> i don't think we have a proper parent context passed from CNI users (maybe CRI)

A Go `Context` is not something you can pass from one process to another.

mars1024:
> We've fixed the issue of not being able to time out the top-level call - that's in libcni.
> After the fix, if the top-level call times out it will kill the plugin process. This should kill any child processes (we may need to do something to ensure that).
> 
> > i don't think we have a proper parent context passed from CNI users (maybe CRI)
> 
> A Go `Context` is not something you can pass from one process to another.

yeah, libcni has supported context, but the upstream caller of  `ipam.ExecAdd` is not libcni, maybe a CNI plugin or even a daemon will call it, so `ipam.ExecAdd` should have a **default** timeout context to avoid being hung, and if caller has a context, maybe it can call `ipam.ExecAddWithContext` instead of `ipam.ExecAdd`, we can deprecate `ipam.Exec` in future.

dcbw:
@mars1024 I'm leaning towards just updating Exec*() with a context argument rather than adding new functions. Then of course we update the plugins themselves, but that's fine.

PR #353, opened 2019-07-10
Increase IfbDeviceName size and refactor hashing functions, fix

### LINKED ISSUE #347: ifb device name hash collisions
Hi all! (long time, I'm back to this for at least a tiny bit)

We got this issue reported in Cloud Foundry: https://github.com/cloudfoundry/silk-release/issues/13

Summary: the bandwidth plugin [creates an IFB device with a name based on only 16 bits of hash material of the container id](https://github.com/containernetworking/plugins/blob/0eddc554c0747200b7b112ce5322dcfa525298cf/plugins/meta/bandwidth/main.go#L114-L122):
```go
func getIfbDeviceName(networkName string, containerId string) (string, error) {
	hash := sha1.New()
	_, err := hash.Write([]byte(networkName + containerId))
	if err != nil {
		return "", err
	}

	return fmt.Sprintf("%x", hash.Sum(nil))[:4], nil
}
```

When you do some basic [birthday problem](https://en.wikipedia.org/wiki/Birthday_problem) math, you discover that with a hundred containers on a single host, the chance of has collision (more than one IFB device with the same name) is close to 10%.

Since the interface name length can be up to 15 characters, we're thinking of mitigating this by changing the above `[:4]` to `[:12]` or `[:14]`.  That should reduce the chance of collision to below 1 in 10^7 even with 500 containers per host.

Anyone got other ideas for less-probabilistic fixes?

mike1808:
* Extract all hashing usage into `pkg/utils/hash` and refactored all usages
* Refactor duplication FormatChainName
* Fixes #347 

Co-authored-by: Cameron Moreau <cmoreau@pivotal.io>

cc @rosenhouse

### DIFF
--- a/pkg/utils/utils.go
+++ b/pkg/utils/utils.go
@@ -22,20 +22,39 @@ import (
 const (
 	maxChainLength = 28
 	chainPrefix    = "CNI-"
-	prefixLength   = len(chainPrefix)
 )
 
-// Generates a chain name to be used with iptables.
-// Ensures that the generated chain name is exactly
-// maxChainLength chars in length
+// FormatChainName generates a chain name to be used
+// with iptables. Ensures that the generated chain
+// name is exactly maxChainLength chars in length.
 func FormatChainName(name string, id string) string {
-	chainBytes := sha512.Sum512([]byte(name + id))
-	chain := fmt.Sprintf("%s%x", chainPrefix, chainBytes)
-	return chain[:maxChainLength]
+	return MustFormatChainNameWithPrefix(name, id, "")
+}
+
+// MustFormatChainNameWithPrefix generates a chain name similar
+// to FormatChainName, but adds a custom prefix between
+// chainPrefix and unique identifier. Ensures that the
+// generated chain name is exactly maxChainLength chars in length.
+// Panics if the given prefix is too long.
+func MustFormatChainNameWithPrefix(name string, id string, prefix string) string {
+	return MustFormatHashWithPrefix(maxChainLength, chainPrefix+prefix, name+id)
 }
 
 // FormatComment returns a comment used for easier
 // rule identification within iptables.
 func FormatComment(name string, id string) string {
 	return fmt.Sprintf("name: %q id: %q", name, id)
+}
+
+const MaxHashLen = sha512.Size * 2
+
+// MustFormatHashWithPrefix returns a string of given length that begins with the
+// given prefix. It is filled with entropy based on the given string toHash.
+func MustFormatHashWithPrefix(length int, prefix string, toHash string) string {
+	if len(prefix) >= length || length > MaxHashLen {
+		panic("invalid length")
+	}
+
+	output := sha512.Sum512([]byte(toHash))
+	return fmt.Sprintf("%s%x", prefix, output)[:length]
 }
--- a/pkg/utils/utils_test.go
+++ b/pkg/utils/utils_test.go
@@ -15,37 +15,151 @@
 package utils
 
 import (
+	"fmt"
+	"strings"
+
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 )
 
 var _ = Describe("Utils", func() {
-	It("must format a short name", func() {
-		chain := FormatChainName("test", "1234")
-		Expect(len(chain)).To(Equal(maxChainLength))
-		Expect(chain).To(Equal("CNI-2bbe0c48b91a7d1b8a6753a8"))
+	Describe("FormatChainName", func() {
+		It("must format a short name", func() {
+			chain := FormatChainName("test", "1234")
+			Expect(len(chain)).To(Equal(maxChainLength))
+			Expect(chain).To(Equal("CNI-2bbe0c48b91a7d1b8a6753a8"))
+		})
+
+		It("must truncate a long name", func() {
+			chain := FormatChainName("testalongnamethatdoesnotmakesense", "1234")
+			Expect(len(chain)).To(Equal(maxChainLength))
+			Expect(chain).To(Equal("CNI-374f33fe84ab0ed84dcdebe3"))
+		})
+
+		It("must be predictable", func() {
+			chain1 := FormatChainName("testalongnamethatdoesnotmakesense", "1234")
+			chain2 := FormatChainName("testalongnamethatdoesnotmakesense", "1234")
+			Expect(len(chain1)).To(Equal(maxChainLength))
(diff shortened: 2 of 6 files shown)


pkg/utils/hash/hash.go
```diff
@@ -0,0 +1,32 @@
...
+import (
+	"crypto/sha512"
+	"fmt"
+)
+
+const maxLen = sha512.Size * 2
+
+func New(input string, length int) (string, error) {
+	if length <= 0 || length > maxLen {
+		return "", fmt.Errorf("only length between 1 and 128 are supported")
```
jellonek: Hmmm... IMO instead of 128 there should be used `maxLen` value, so if someday we would switch from sha512 to something other - this would be still correct.

pkg/utils/utils.go
```diff
@@ -29,9 +30,13 @@ const (
 // Ensures that the generated chain name is exactly
 // maxChainLength chars in length
 func FormatChainName(name string, id string) string {
-	chainBytes := sha512.Sum512([]byte(name + id))
-	chain := fmt.Sprintf("%s%x", chainPrefix, chainBytes)
-	return chain[:maxChainLength]
+	return FormatChainNameWithPrefix(name, id, "")
```
jellonek: Could you update there docstring (and also add one to next func) according to https://github.com/golang/go/wiki/CodeReviewComments#doc-comments and https://golang.org/doc/effective_go.html#commentary
Especially I'm asking about:
```
If every doc comment begins with the name of the item it describes, you can use the doc subcommand of the go tool and run the output through grep. Imagine you couldn't remember the name "Compile" but were looking for the parsing function for regular expressions, so you ran the command,

$ go doc -all regexp | grep -i parse
```

pkg/utils/utils_test.go
```diff
@@ -20,32 +20,70 @@ import (
 )
 
 var _ = Describe("Utils", func() {
-	It("must format a short name", func() {
-		chain := FormatChainName("test", "1234")
-		Expect(len(chain)).To(Equal(maxChainLength))
-		Expect(chain).To(Equal("CNI-2bbe0c48b91a7d1b8a6753a8"))
-	})
+	Context("FormatChainName", func() {
```
rosenhouse: Idiomatic Ginkgo would have this be a `Describe` instead of `Context`.  So that the full test can be read in English like
> Describe [the function] FormatChainName [, ] It must format a short name

pkg/utils/utils.go
```diff
@@ -29,9 +30,13 @@ const (
...
+}
+
+func FormatChainNameWithPrefix(name string, id string, prefix string) string {
+	length := maxChainLength - prefixLength - len(prefix)
+	chain, _ := hash.New(name+id, length)
```
rosenhouse: Discarding this error seems risky.  The caller could pass in a long `prefix` string which could cause `length` to become less than 1.  Better to either
1. return the error up to the caller, or
2. alternately, have the `hash.New` internally `panic`.  But "[library functions should avoid panic](https://golang.org/doc/effective_go.html#panic)" and I'd only suggest doing that if we were confident that it was always a programmer error and not a user-input or environment issue.

plugins/meta/bandwidth/bandwidth_linux_test.go
```diff
@@ -105,7 +105,7 @@ var _ = Describe("bandwidth test", func() {
 		hostIP = net.IP{169, 254, 0, 1}
 		containerIP = net.IP{10, 254, 0, 1}
 		hostIfaceMTU = 1024
-		ifbDeviceName = "5b6c"
+		ifbDeviceName = "a8eda89404b7c8"
```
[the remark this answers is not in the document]
cameronmoreau: That makes sense 👍 

Since "b" is a valid hex character, @mike1808 and I have decided on the prefix "bw" to avoid confusion

mike1808:
Thanks for the review! Check new changes for the fixes @rosenhouse @jellonek

plugins/meta/bandwidth/main.go
```diff
@@ -112,13 +114,11 @@ func validateRateAndBurst(rate int, burst int) error {
 }
 
 func getIfbDeviceName(networkName string, containerId string) (string, error) {
-	hash := sha1.New()
-	_, err := hash.Write([]byte(networkName + containerId))
+	hash, err := hash.New(networkName+containerId, maxIfbDeviceLength-3)
```
shawn-ogg: Is `-3` correct? I see that you append "bw" to the return value which has only a length of `2`

mike1808:
Hey @shawn-ogg , do you have a specific use-case where you would run multiple bandwidth plugins per container?  We'd like to do more research on this, but our assumption is that every `containerId` would be unique (not sure about `networkName`).

Correct me if I'm wrong but, I think the current (on master) implementation will result in all `sha1(networkName, containerId)` being unique, but the collision happens when those sha's share the same first 4 digits.

Also the [cmdDel() method](https://github.com/containernetworking/plugins/blob/966bbcb8a572aa4c253ea9650888ad2306e3c8f2/plugins/meta/bandwidth/main.go#L242) relies on `getIfbDeviceName()` to be deterministic (given the same name and containerId).

cc @cameronmoreau

plugins/meta/bandwidth/main.go
rosenhouse: Yeah, it'd much rather have 
```golang
const ifbDeviceNamePrefix = "bw"
```
and
```golang
hash.New(..., maxIfbDeviceLength-len(ifbDeviceNamePrefix))
```

pkg/utils/utils.go
rosenhouse: You're still discarding this error, no?

cameronmoreau:
> In general I think you just made the possibility of an name collision smaller (by a lot). But the main issue still remains: the hash is calculated from the `networkName` and `containerId`. If there is a name collision we stuck, because a rerun would still end up with the same hash.
> One solution could be to add some randomness (e.g. the timestamp) to it. Of course only if there is no other place where you must deduce the ifname using the hash function (I'm not aware of it).
> 
> Another approach would be to use random strings:
> (e.g. https://github.com/dchest/uniuri)
> and implement a retry in the unlikely event of a clash.

@shawn-ogg I think you may have found a bug (or new features?) in the existing bandwidth plugin 🎉I don't believe you can run multiple bandwidth plugins in a single container today, as the `networkName` and `containerId` will not be different. What are your thoughts on opening a separate issue for this?

cc @mike1808 @rosenhouse

squeed:
lgtm!

PR #354, opened 2019-07-17
Update vendored coreos/ip-tables

mccv1r0:


nagiesek:
Could you reflect this within go.mod?

Issue #355, opened 2019-07-17
Verify vendor integrity in CI


squeed:
Now that we can do `go mod verify`, we should do that in CI.

squeed:
Gabe suggested adding this to the end of `test.sh`. I'd also like to see it in `release.sh`

squeed:
Unfortunate discovery: `go mod verify` doesn't actually check that `vendor/` is correct. It will, someday, but it doesn't right now. At this point, it really doesn't add any value.

PR #356, opened 2019-07-17
Run 'go mod verify' at the end of tests

bboreham:
This ensures that dependencies match the list in go.sum.

In particular, this will run as part of CI so we catch cases where vendored files are changed directly.

### DIFF
--- a/test_linux.sh
+++ b/test_linux.sh
@@ -57,3 +57,10 @@ if [ -n "${vetRes}" ]; then
 	echo -e "govet checking failed:\n${vetRes}"
 	exit 255
 fi
+
+echo "Checking go mod verify..."
+modRes=$(go mod verify)
+if [ -n "${modRes}" ]; then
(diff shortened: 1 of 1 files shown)


test_linux.sh
```diff
@@ -57,3 +57,10 @@ if [ -n "${vetRes}" ]; then
 	echo -e "govet checking failed:\n${vetRes}"
 	exit 255
 fi
+
+echo "Checking go mod verify..."
+modRes=$(go mod verify)
+if [ -n "${modRes}" ]; then
```
jellonek: The test there should be like:
```sh
if [ $? -gt 0 ]; then
...
```

jellonek:
Related issue: #355

squeed:
It looks like `go mod verify` doesn't do what we want anyways. Instead, we need to "go mod vendor" and see if any files have changed. Bah.

bboreham:
Closing because this isn't actually what we want.

PR #357, opened 2019-07-19
host-device: revert name setting to make retries idempotent

mars1024:
If moving device to host network namespace fails, the retries will always fail because the interface with name **ifName** does not exist anymore, so we have to revert name setting, to make retires idempotent.

### DIFF
--- a/plugins/main/host-device/host-device.go
+++ b/plugins/main/host-device/host-device.go
@@ -218,14 +218,23 @@ func moveLinkOut(containerNs ns.NetNS, ifName string) error {
 		}
 
 		// Devices can be renamed only when down
-		if err := netlink.LinkSetDown(dev); err != nil {
+		if err = netlink.LinkSetDown(dev); err != nil {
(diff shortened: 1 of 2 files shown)


mars1024:
> @mars1024 Could you add a testcase for this? It should basically be:
> 
> * testutils.CmdAddWithArgs() to move the dummy interface to the container
> * create a dummy interface with same name as the host device's original name
> * testutils.CmdDelWithArgs() to force an error, check that the error occurred
> * test that the interface still exists in the container's namespace with the CNI_IFNAME
> * delete the dummy interface from the fake host namespace
> * testutils.CmdDelWithArgs() again and expect that it succeeds

Yes, of course, thanks for your detailed suggestion, @dcbw , I'll update this soon ~

mars1024:
Updated! Please go on reviewing, thanks ~ @dcbw

squeed:
great tests, thanks.

PR #364, opened 2019-08-07
Remove link Down/Up in MAC address change to prevent route flush

s1061123:


plugins/meta/tuning/tuning.go
```diff
@@ -93,15 +93,7 @@ func changeMacAddr(ifName string, newMacAddr string) error {
...
-	err = netlink.LinkSetDown(link)
-	if err != nil {
-		return fmt.Errorf("failed to set %q down: %v", ifName, err)
-	}
-	err = netlink.LinkSetHardwareAddr(link, addr)
-	if err != nil {
-		return fmt.Errorf("failed to set %q address to %q: %v", ifName, newMacAddr, err)
-	}
-	return netlink.LinkSetUp(link)
+	return netlink.LinkSetHardwareAddr(link, addr)
```
jellonek: :+1:

btw. It would be nice to send also after that GARP as in https://github.com/higebu/gogarp
... or maybe that should go through a separate PR?

dcbw: I also wonder if all network drivers actually support this?

squeed:
seems fine.

plugins/meta/tuning/tuning.go
s1061123: @jellonek good question. for now, cni is invoked at initialization phase, so I suppose no neighbor is associated, so garp is not so strongly required, but I agree we may send garp to let others know the mac. btw, I'm wondering why netlink in kernel does not send garp...

@dcbw well, interesting question :simple_smile:.
In the kernel, ndo_set_mac_address() of the interface drive is invoked to change the MAC address at dev_set_mac_address().
https://github.com/torvalds/linux/blob/9c7db5004280767566e91a33445bf93aa479ef02/net/core/dev.c#L7820-L7843

As far as I looked the driver in linux/drivers/net/ethernet, most driver supports ndo_set_mac_address(), but it does not guarantee that all driver supports ndo_set_mac_address(). Should we mention that in README?

jellonek: Btw. we are already doing something similar in case of https://github.com/containernetworking/plugins/blob/master/plugins/main/ptp/ptp.go#L145 - one liner, but I agree that this could go through a separate PR in which we can discuss if that's needed/harmful separately.

For second thing - if we are not sure if all drivers do that - we should probably have at least optional possibility to up/down interface, because only mentioning that in README seems to be not enough...

PR #374, opened 2019-08-19
loopback support CNI CHECK

mars1024:
Loopback plugin should support CNI CHECK since it supports CNI version 0.4.0


squeed:
Excellent, thanks!

plugins/main/loopback/loopback.go
```diff
@@ -78,6 +81,18 @@ func main() {
 }
 
 func cmdCheck(args *skel.CmdArgs) error {
-	// TODO: implement
-	return nil
+	args.IfName = "lo"
```
bboreham: There's a comment on the other places we force this value: probably good to have the same here.

mars1024: thanks, will update it

PR #341, opened 2019-06-18
bump containernetworking/cni to v0.7.1

mars1024:
ref to [https://github.com/containernetworking/plugins/issues/340](https://github.com/containernetworking/plugins/issues/340) and [https://github.com/containernetworking/plugins/issues/303](https://github.com/containernetworking/plugins/issues/303)

1. bump containernetworking/cni to v0.7.1
2. use delegateArgs in `ipam.ExecDel` instead of setting/unsetting environment

PR #377, opened 2019-08-23
bump up libcni to v0.7.1

mars1024:
#341 try to bump up libcni to v0.7.1 and do some code changes, but I found libcni was downgraded to v0.7.0 by switching from glide to go module, so I think we should bump up libcni again.


mars1024:
@squeed @dcbw @bboreham  PTAL, thanks!

PR #380, opened 2019-08-31
add dnsmasq plugin

baude:
the dnsmasq plugin is designed to help with resolving names between pods
or containers. it uses a dnsmasq instance that only listens on the
network interface in question.  otherwise, the dnsmasq instance will act
as a caching server to resolve "outside" requests.

Signed-off-by: baude <bbaude@redhat.com>


plugins/meta/dnsmasq/README.md
```diff
@@ -0,0 +1,44 @@
+# dnsmasq plugin
+
+## Overview
+
+This plugin creates firewall rules to allow traffic to/from container IP address via the host network .
```
squeed: Looks like this paragraph needs to be deleted :-)

jellonek: Please remove a space at the end, before the dot.

bboreham:
I can't figure out the requirement you are trying to meet.
Could you file an issue stating that?

baude:
> I can't figure out the requirement you are trying to meet.
> Could you file an issue stating that?

@bboreham would you accept a better written commit message?

bboreham:
Maybe. It’s just better to have an issue, then we could have multiple PRs attempt to fix it, if we close this PR the issue remains, etc.

tdfischer:
Chiming in with my own use case. I, like many others, have a home server that runs a sort of local container cloud. A couple of services I run utilize UPNP to discover devices and multimedia sources on my home LAN, plus some automatic port forwarding at my gateway, meaning that they're all using --host=net and competing to listen on port 8080. The solution I've come up with is to put everything onto a macvlan network which itself is on top of a real vlan device. My containers can see everything and they're not competing for that single port anymore. I use the dhcp plugin to allow my existing pfsense gateway to assign IP addresses over that VLAN; I could also use ipvlan here, but that doesn't change the problem.

Once these containers are on the ipvlan or macvlan, I can no longer use the regular port forwarding tools. If I need to access these containers through, eg, my nginx reverse proxy that applies letsencrypt SSL for public internet access, I effectively would need to assign a static IP address to maintain access across reboots and container restarts. This is where I need either this dnsmasq plugin (for ipvlan), or the answer in #435 (for macvlan). This dnsmasq plugin would let me point my existing pfsense DNS server at it for resolving .cloud domains whereas #435 would let my existing pfsense server handle that instead.

Hopefully this long explanation makes sense.

lucab:
@jellonek @squeed @matthewdupre for your consideration, even if this PR hasn't landed here yet, the same logic as a [standalone plugin](https://github.com/containers/dnsname) is already [causing](https://github.com/coreos/fedora-coreos-tracker/issues/519) [troubles](https://github.com/containers/dnsname/issues/22) to users and distributions.

> The DNSMasq services are not configured or managed by systemd but rather only by the plugin itself.

In particular, this part assumes that a `dnsmasq` binary exists in the host OS, it can be owned/run by the CNI plugin and its build-time configuration and its version suits whatever are the requirements here. This does not match reality on modern Linux distributions, where:
 * the binary may not be found in PATH, if a `dnsmasq` service is not meant to run on that host (e.g. because it's not useful in this environment, or because it runs as a containerized service instead)
 * the binary found in PATH may be tailored to users' need regarding versioning/patching/build-tuning, and it is anyway not a CLI utility

In short, this proposed plugin takes a design decision to use an on-host `dnsmasq` to implement part of its logic. However, that binary is commonly regarded as a user-service, not a widespread CLI utility.
This clashes with real world Linux distributions tailored at containerized workloads, which is where this plugin is most likely to run. Additionally, `dnsmasq` is a component with a [less-than-stellar security record](https://security.googleblog.com/2017/10/behind-masq-yet-more-dns-and-dhcp.html).
A memory-safe and self-contained design for the same feature would better suit CNI consumers.

Issue #381, opened 2019-08-31
Pods creating failed because of "cannot convert: no valid IP addresses" in master branch


frui0101:
# Description: 
+ in master branch
+ kube-flannel.yaml together with loopback cni

# WR:
+ works in v0.7
+ works if replace conf.CNIVersion(types.PrintResult(result, conf.CNIVersion) in loopback cmdAdd procedure to types.PrintResult(result, "0.3.1") directly in master branch. conf.CNIVersion default is 0.2.0 which supplied by k8s, no place to change it by configuration.

Funky-Hamster:
Did you find the solution? I got the same issue and I think the updates recently caused that.

squeed:
Can you post your CNI config?

mccv1r0:
Podman has this issue too: `https://bugzilla.redhat.com/show_bug.cgi?id=1748214`

dcbw:
Ok, I think this because loopback has never returned an actual result, always an empty one. Which is kinda wrong, I suppose. Here's code in loopback.go:

```
	var result types.Result
	if conf.PrevResult != nil {
		// If loopback has previous result which passes from previous CNI plugin,
		// loopback should pass it transparently
		result = conf.PrevResult
	} else {
		result = &current.Result{}
	}

	return types.PrintResult(result, conf.CNIVersion)
```
When there's no PrevResult loopback will return an empty current.Result{}. If the loopback config specified CNI Version 0.2.0, PrintResult will convert the empty current result to a 0.2.0 result. The conversion code for that has:

```
	if oldResult.IP4 == nil && oldResult.IP6 == nil {
		return nil, fmt.Errorf("cannot convert: no valid IP addresses")
	}
```
which of course will be triggered because loopback never returns IP addresses.

The original problem was that loopback did not conform to the CNI specification by ensuring that it's result was the same version as the incoming configuration. That was fixed by https://github.com/containernetworking/plugins/pull/374 and now it returns the result in the right version.

In any case we have two options:
1. relax the requirement that down-converting 0.3.0+ -> 0.2.0 must have an IP address
2. fix loopback to technically conform to the specification and return a real result with the interface and IP address it assigned

PR #383, opened 2019-09-06
When prevResults are not supplied to loopback plugin, create results to return

mccv1r0:
When prevResults are not returned to loopback plugin, create results to return based on the lo interface and IP address assigned inside container. 

Fixes [381](https://github.com/containernetworking/plugins/issues/381)


bboreham:
LGTM if we can have one or two tests.

PR #389, opened 2019-09-19
Use uint64 for Bandwidth plugin

CallMeFoxie:
This is the first out of several required PRs across CNI plugins, CRI-O and Kubelet.

We need to use something bigger than int as it goes only into 2Gb for rate and burst, which is not enough for 10Gbps networking anymore.
Since we use bits in JSON and can feed upto 4GByte into tbf (at least per current implementation), we need to use uint64 even for burst and then number check it for uint32 max * 8.

Thanks for looking into this
Ashley

### DIFF
--- a/plugins/meta/bandwidth/main.go
+++ b/plugins/meta/bandwidth/main.go
@@ -17,6 +17,7 @@ package main
 import (
 	"encoding/json"
 	"fmt"
+	"math"
 
 	"github.com/vishvananda/netlink"
 
@@ -37,11 +38,11 @@ const ifbDevicePrefix = "bwp"
 // BandwidthEntry corresponds to a single entry in the bandwidth argument,
 // see CONVENTIONS.md
 type BandwidthEntry struct {
-	IngressRate  int `json:"ingressRate"`  //Bandwidth rate in bps for traffic through container. 0 for no limit. If ingressRate is set, ingressBurst must also be set
-	IngressBurst int `json:"ingressBurst"` //Bandwidth burst in bits for traffic through container. 0 for no limit. If ingressBurst is set, ingressRate must also be set
+	IngressRate  uint64 `json:"ingressRate"`  //Bandwidth rate in bps for traffic through container. 0 for no limit. If ingressRate is set, ingressBurst must also be set
+	IngressBurst uint64 `json:"ingressBurst"` //Bandwidth burst in bits for traffic through container. 0 for no limit. If ingressBurst is set, ingressRate must also be set
 
-	EgressRate  int `json:"egressRate"`  //Bandwidth rate in bps for traffic through container. 0 for no limit. If egressRate is set, egressBurst must also be set
-	EgressBurst int `json:"egressBurst"` //Bandwidth burst in bits for traffic through container. 0 for no limit. If egressBurst is set, egressRate must also be set
+	EgressRate  uint64 `json:"egressRate"`  //Bandwidth rate in bps for traffic through container. 0 for no limit. If egressRate is set, egressBurst must also be set
+	EgressBurst uint64 `json:"egressBurst"` //Bandwidth burst in bits for traffic through container. 0 for no limit. If egressBurst is set, egressRate must also be set
 }
 
 func (bw *BandwidthEntry) isZero() bool {
@@ -101,7 +102,7 @@ func getBandwidth(conf *PluginConf) *BandwidthEntry {
 	return conf.BandwidthEntry
 }
 
-func validateRateAndBurst(rate int, burst int) error {
+func validateRateAndBurst(rate, burst uint64) error {
 	switch {
 	case burst < 0 || rate < 0:
 		return fmt.Errorf("rate and burst must be a positive integer")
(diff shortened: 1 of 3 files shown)


CallMeFoxie:
Here's a fix for CRI-O https://github.com/cri-o/cri-o/pull/2834

CallMeFoxie:
And for kubernetes: https://github.com/kubernetes/kubernetes/pull/82871

squeed:
hah, good catch.
LGTM.

plugins/meta/bandwidth/main.go
```diff
@@ -101,14 +102,16 @@ func getBandwidth(conf *PluginConf) *BandwidthEntry {
...
-func validateRateAndBurst(rate int, burst int) error {
+func validateRateAndBurst(rate, burst uint64) error {
 	switch {
 	case burst < 0 || rate < 0:
 		return fmt.Errorf("rate and burst must be a positive integer")
 	case burst == 0 && rate != 0:
 		return fmt.Errorf("if rate is set, burst must also be set")
 	case rate == 0 && burst != 0:
 		return fmt.Errorf("if burst is set, rate must also be set")
+	case burst/8 > math.MaxUint32:
```
mars1024: Rate can be more than 4 GB here?

mars1024: nit: Could you help to add some extra tests for `validateRateAndBurst`?

CallMeFoxie: > Rate can be more than 4 GB here?

Can be as it comes in argument as uint64 but later gets converted to a byte and casted to uint32. Kept the same logic in this validity check as is used later down the path for call to tc.

mars1024: OK, got your point.

jellonek: This new case deserves for new test case. Can you add it in this PR?

CallMeFoxie: Yes, just need to check out the testing framework, will try to make some time tomorrow.

bboreham:
I'm interested: where does the limitation come from?  `int` is 64-bit on many platforms.

jellonek:
Still `int` size is platform specific and having there a more precise type seems fine for me.

CallMeFoxie:
actually int when casted to JSON is only 32bit even on amd64 platform it seems :) at least based on the errors in logs.

jellonek:
Oh, that even more interesting :D

bboreham:
Could you post one of those errors here?

CallMeFoxie:
That was on CRI-O about a week ago, something about not being able to cast <uint64 max in decimal> to int, coming from the Bandwidth plugin. Sadly I don't keep these logs :(

CallMeFoxie:
(CRI-O was using uint64 at the time)

mars1024:
> I'm interested: where does the limitation come from? `int` is 64-bit on many platforms.

But we can only use half of int because only positive bandwidth will be accepted... For me, I'm more interested about why tbf has a 4GB burst limit...

bboreham:
"burst" perhaps doesn't mean what you think it does.  It's the size of bucket used in the leaky-bucket algorithm, effectively it governs the granularity of the limiter. Nobody needs to be less granular than 4GB.

Some hints at https://linux.die.net/man/8/tc-tbf.

bboreham:
Came back to say I somewhat doubt my own understanding of this thing. Filed #392 for one aspect.

squeed:
lgtm.

dcbw:
/lgtm

Though @CallMeFoxie if you could paste in the error you got from CRIO that would make @bboreham happy...

jellonek:
btw. we are still waiting for this unittest, right?

CallMeFoxie:
The error is:
```
Oct 04 14:32:54 somenode kubelet[26269]: E1004 14:32:54.699051   26269 pod_workers.go:191] Error syncing pod 4146ace4-ab18-47f0-9fad-7b25f5748f7b ("test-bw-85c6fc4999-bh8l5_default(4146ace4-ab18-47f0-9fad-7b25f5748f7b)"), skipping: failed to "CreatePodSandbox" for "test-bw-85c6fc4999-bh8l5_default(4146ace4-ab18-47f0-9fad-7b25f5748f7b)" with CreatePodSandboxError: "CreatePodSandbox for pod \"test-bw-85c6fc4999-bh8l5_default(4146ace4-ab18-47f0-9fad-7b25f5748f7b)\" failed: rpc error: code = Unknown desc = failed to create pod network sandbox k8s_test-bw-85c6fc4999-bh8l5_default_4146ace4-ab18-47f0-9fad-7b25f5748f7b_0(1a150fb5fc9b8625a4c9a933ddf5cd5dd9891b7159a686c41415014170395686): failed to parse network configuration: json: cannot unmarshal number 18446744073709551615 into Go struct field BandwidthEntry.egressBurst of type int"
```

bboreham:
So if you were to merge this PR that call would error with `burst cannot be more than 4GB`.

PR #394, opened 2019-09-30
bridge: check vlan id when loading net conf

mars1024:
Check vlan id if in [0, 4094] when loading net conf to avoid unexpected error from kernel.

### DIFF
--- a/plugins/main/bridge/bridge.go
+++ b/plugins/main/bridge/bridge.go
@@ -76,7 +76,7 @@ func loadNetConf(bytes []byte) (*NetConf, string, error) {
 		return nil, "", fmt.Errorf("failed to load netconf: %v", err)
 	}
 	if n.Vlan < 0 || n.Vlan > 4094 {
-		return nil, "", fmt.Errorf(`invalid VLAN ID %d (must be between 0 and 4094)`, n.Vlan)
+		return nil, "", fmt.Errorf("invalid VLAN ID %d (must be between 0 and 4094)", n.Vlan)
 	}
 	return n, n.CNIVersion, nil
 }
--- a/plugins/main/bridge/bridge.go
+++ b/plugins/main/bridge/bridge.go
@@ -75,6 +75,9 @@ func loadNetConf(bytes []byte) (*NetConf, string, error) {
 	if err := json.Unmarshal(bytes, n); err != nil {
 		return nil, "", fmt.Errorf("failed to load netconf: %v", err)
 	}
+	if n.Vlan < 0 || n.Vlan > 4094 {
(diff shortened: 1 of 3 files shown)


mars1024:
Recently I met test errors like `verifying golang.org/x/tools@v0.0.0-20190929041059-e7abfedfabcf: golang.org/x/tools@v0.0.0-20190929041059-e7abfedfabcf: Get https://sum.golang.org/lookup/golang.org/x/tools@v0.0.0-20190929041059-e7abfedfabcf: dial tcp 108.177.111.141:443: i/o timeout
The command "go get golang.org/x/tools/cmd/cover" failed and exited with 1 during .` several times, I had to force-push to test again, any ideas about this? @bboreham @squeed @dcbw @jellonek

squeed:
lgtm

plugins/main/bridge/bridge.go
```diff
@@ -75,6 +75,9 @@ func loadNetConf(bytes []byte) (*NetConf, string, error) {
 	if err := json.Unmarshal(bytes, n); err != nil {
 		return nil, "", fmt.Errorf("failed to load netconf: %v", err)
 	}
+	if n.Vlan < 0 || n.Vlan > 4094 {
+		return nil, "", fmt.Errorf(`invalid VLAN ID %d (must be between 0 and 4094)`, n.Vlan)
```
jellonek: Why there you are using backtick instead of commonly used `"`?

dcbw: @mars1024 we all agree that if you change to " then we'll merge :) Thanks!

mars1024: @jellonek @dcbw definitely OK, btw, I will change the [usage](https://github.com/containernetworking/plugins/blob/15894b36a0a679028768bf02498618992e808136/plugins/main/vlan/vlan.go#L59) in vlan plugin too.

mars1024: Updated, thanks!

plugins/main/vlan/vlan.go
```diff
@@ -53,10 +53,10 @@ func loadConf(bytes []byte) (*NetConf, string, error) {
 		return nil, "", fmt.Errorf("failed to load netconf: %v", err)
 	}
 	if n.Master == "" {
-		return nil, "", fmt.Errorf(`"master" field is required. It specifies the host interface name to create the VLAN for.`)
+		return nil, "", fmt.Errorf("\"master\" field is required. It specifies the host interface name to create the VLAN for.")
```
jellonek: This was the precise example for backtick usage, but ok, lets keep everywhere `"` :)

mars1024: Yeap, I also read some docs about double quotes and back quote... It says that back quote is usually used for multi lines, html and regular expression. So I think both are OK here.

Issue #339, opened 2019-06-18
do we need a ipam plugin based on rpc


mars1024:
Recently I'm working on a CNI plugin which need to use a global centralized IPAM, and I implemented a client/server IPAM framework based on grpc, I'm wondering if we need a generic rpc-based IPAM plugin, which including a rpc proto file and a CNI plugin which allocate/release IP through grpc.
The rpc proto file is just like below, I think we can make it a standard,
```
service IPAMService {
    rpc AllocateIP (AllocateIPRequest) returns (AllocateIPResponse) {}
    rpc ReleaseIP (ReleaseIPRequest) returns (ReleaseIPResponse) {}
}

message AllocateIPRequest {
    string PodName = 1;
    string PodNamespace = 2;
    string SandboxID = 3;
    string NetNS = 4;
    string IfName = 5;
}
...
```

Then, we can build a centralized IPAM server based on **IPAMService** in proto file, after that, we can use it easily with this plugin.

Anyone has some suggestions?

MikeZappa87:
@mars1024 as I am looking at rpc right now. Possibly we could revive the rpc efforts.

PR #395, opened 2019-10-02
Add prune option to remove unused bridge so far

s1061123:
This change introduces new option 'prune' in bridge CNI plugin. With `prune:true`, bridge CNI will remove the bridge if bridge is no longer used.

### DIFF
--- a/plugins/main/bridge/bridge.go
+++ b/plugins/main/bridge/bridge.go
@@ -20,9 +20,12 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net"
+	"os"
+	"path/filepath"
 	"runtime"
 	"syscall"
 
+	"github.com/alexflint/go-filemutex"
 	"github.com/j-keck/arping"
 	"github.com/vishvananda/netlink"
 
@@ -41,6 +44,7 @@ import (
 var debugPostIPAMError error
 
 const defaultBrName = "cni0"
+const defaultLockDir = "/var/lib/cni/bridge"
 
 type NetConf struct {
 	types.NetConf
@@ -53,12 +57,19 @@ type NetConf struct {
 	HairpinMode  bool   `json:"hairpinMode"`
 	PromiscMode  bool   `json:"promiscMode"`
 	Vlan         int    `json:"vlan"`
+	Prune        bool   `json:"prune"`
+	LockDir      string `json:"lockDir"`
 }
 
 type gwInfo struct {
 	gws               []net.IPNet
 	family            int
 	defaultRouteFound bool
+}
+
+// BridgeLock wraps os.File to be used as a lock using flock for bridge
+type BridgeLock struct {
+	f *filemutex.FileMutex
 }
 
 func init() {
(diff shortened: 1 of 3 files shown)


dcbw:
@s1061123 pruning the bridge needs to be locked against concurrent DEL/ADD. For example if the last container is being deleted and then a different container starts to be added, these will race and end up with sadness.

plugins/main/bridge/bridge.go
```diff
@@ -589,6 +590,30 @@ func cmdDel(args *skel.CmdArgs) error {
 		}
 	}
 
+	// Remove bride if prunening is required
+	if n.Prune == true {
+		br, errprune := bridgeByName(n.BrName)
+
+		if errprune == nil {
+			brIndex := br.Attrs().Index
```
jellonek: Shouldn't this index be read from link gathered in LinkList below? In theory it LinkList could return other index for this bridge (if it was in between deleted/readded)? Very edge situation, so i'm not insisting on that, but imo you should once use LinkList, then double reiterate over it - once to find bridge, second time - using the same list - to step over master indexes of all links.

plugins/main/bridge/bridge.go
```diff
@@ -589,6 +590,30 @@ func cmdDel(args *skel.CmdArgs) error {
...
+			links, err := netlink.LinkList()
+			if err != nil {
+				return err
+			}
+
+			portCount := 0
+			for _, l := range links {
+				if l.Attrs().MasterIndex == brIndex {
+					portCount = portCount + 1
+				}
```
mccv1r0: Add break?

s1061123: fixed.

plugins/main/bridge/bridge.go
s1061123: okey. changed.

plugins/main/bridge/bridge.go
```diff
@@ -589,6 +655,39 @@ func cmdDel(args *skel.CmdArgs) error {
...
+				br = l
+				break
+			}
+		}
+
+		if brIndex == 0 {
+			return fmt.Errorf("cannot found bridge")
+		}
+
+		portCount := 0
```
jellonek: As there is used break after first found interface - keeping that as `int` instead of `bool` (lets say: `hasSlaves`) has less sense now.

plugins/main/bridge/bridge.go
```diff
@@ -53,6 +57,8 @@ type NetConf struct {
 	HairpinMode  bool   `json:"hairpinMode"`
 	PromiscMode  bool   `json:"promiscMode"`
 	Vlan         int    `json:"vlan"`
+	Prune        bool   `json:"prune"`
+	LockDir      string `json:"lockDir"`
```
mars1024: What's this field used for? Seems useless...

mars1024:
> @s1061123 pruning the bridge needs to be locked against concurrent DEL/ADD. For example if the last container is being deleted and then a different container starts to be added, these will race and end up with sadness.
> 
> But locking will serialize all access to the bridge plugin, which will slow down container networking operations. Which is the reason we haven't done this until now.
> 
> What we really want to do is add global (eg not per-container) plugin operations like "init" and "cleanup" like we have for VERSION. And we had been punting this to gRPC but that's been slow going so far.
> 
> BUT... Casey points out that we can reduce the cost of serialization by being careful about the locking. For ADD, lock everything until the container interface is added to the bridge. For DEL, lock the check for whether the bridge has interfaces or not.

@dcbw @squeed  gRPC CNI? Looks great! Could I know some details about this?  related issue [https://github.com/containernetworking/plugins/issues/339](https://github.com/containernetworking/plugins/issues/339)

plugins/main/bridge/bridge.go
s1061123: optional parameter to specify directory where plugin configure lock file.
We could remove it and change always use default dir. Any preferences?

mars1024:
I still think this feature will slow down the container networking operations just as @dcbw said above.
Ant I think maybe there is another way, we can have a disk file to count the number of interfaces which have been attached on the bridge and another file to lock the write operation on this file, this way will reduce the time interval between lock/unlock (an file IO against a container networking operation) to help reduce concurrency race.

plugins/main/bridge/bridge.go
mars1024: Oh... a mistake, maybe we should put this in docs..

s1061123:
@mars1024 thank you for your good comments. I just wrote this PR to improve the operation but I agree that it affects everything (performance, usability, stability and so on).

So I am thinking that we close it and postpone this feature to support it in gRPC version CNI (with various features).

s1061123:
So waiting for other opinion a few days and close it.

s1061123:
@mars1024 In this case, bridge may cause an issue if user add physical interface or another veth which is created by other than bridge CNI because this lock/counter is only used in bridge cni.

mars1024:
> @mars1024 In this case, bridge may cause an issue if user add physical interface or another veth which is created by other than bridge CNI because this lock/counter is only used in bridge cni.

Oh, I think this case is out of our design, we can add some comments in README to avoid users from using `prune` option in this case.

mars1024:
@s1061123  How about add a judging condition before cleaning up bridge like
```
	if cfg.Prune && counter.Get() == 0 {
                if ! bridgeHasSlaves() {
                    // clean up bridge
                }
	}
```

PR #397, opened 2019-10-07
contributing doc: describe cnitool installation

oshothebig:
I found a description to install `cnitool` is missing in the current `CONTRIBUTING.md` while `cnitool` is necessary to be installed to run the tests.


squeed:
Aha, good point. Thanks for the update.

oshothebig:
Thanks for merging.

PR #400, opened 2019-10-11
static: prioritize the input sources for IPs

### LINKED ISSUE #399: static: prioritize between runtimeConfig, CNI_ARGS and json "args"
Currently static ipam plugin supports following input for IPs:

- runtimeConfig
- CNI_ARGS
- "args" in json

When the 'runtimeConfig' and 'CNI_ARGS' are used at same time, we should prioritize which one should be used.

s1061123:
This change introduce priorities for IPs input among CNI_ARGS, 'args' and runtimeConfig.
Fix #399

Signed-off-by: Tomofumi Hayashi <tohayash@redhat.com>

### DIFF
--- a/plugins/ipam/static/README.md
+++ b/plugins/ipam/static/README.md
@@ -60,3 +60,9 @@ The plugin also support following [capability argument](https://github.com/conta
 The following [args conventions](https://github.com/containernetworking/cni/blob/master/CONVENTIONS.md#args-in-network-config) are supported:
 
 * `ips` (array of strings): A list of custom IPs to attempt to allocate, with prefix (e.g. '10.10.0.1/24')
+
+Notice: If some of above are used at same time, only one will work according to the priorities below
+
+1. [capability argument](https://github.com/containernetworking/cni/blob/master/CONVENTIONS.md)
+1. [args conventions](https://github.com/containernetworking/cni/blob/master/CONVENTIONS.md#args-in-network-config)
+1. [CNI_ARGS](https://github.com/containernetworking/cni/blob/master/SPEC.md#parameters)
--- a/plugins/ipam/static/main.go
+++ b/plugins/ipam/static/main.go
@@ -145,11 +145,44 @@ func LoadIPAMConfig(bytes []byte, envArgs string) (*IPAMConfig, string, error) {
 		return nil, "", err
 	}
 
-	if len(n.RuntimeConfig.IPs) != 0 {
-		// args IP overwrites IP, so clear IPAM Config
-		n.IPAM.Addresses = make([]Address, 0, len(n.RuntimeConfig.IPs))
-		for _, addr := range n.RuntimeConfig.IPs {
-			n.IPAM.Addresses = append(n.IPAM.Addresses, Address{AddressStr: addr})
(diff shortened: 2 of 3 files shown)


plugins/ipam/static/README.md
```diff
@@ -60,3 +60,9 @@ The plugin also support following [capability argument](https://github.com/conta
 The following [args conventions](https://github.com/containernetworking/cni/blob/master/CONVENTIONS.md#args-in-network-config) are supported:
 
 * `ips` (array of strings): A list of custom IPs to attempt to allocate, with prefix (e.g. '10.10.0.1/24')
+
+Note: If some of above is used at same time, one input sources are used as following priority:
```
mars1024: nit:  `Notice: If some of above are used at same time, only one will work according to the priorities below`

plugins/ipam/static/main.go
```diff
@@ -145,8 +145,49 @@ func LoadIPAMConfig(bytes []byte, envArgs string) (*IPAMConfig, string, error) {
...
+
+				addr := Address{
+					Address:    net.IPNet{IP: ip, Mask: subnet.Mask},
+					AddressStr: ipstr,
+				}
+				n.IPAM.Addresses = append(n.IPAM.Addresses, addr)
+			}
+		}
+
+		if e.GATEWAY != "" {
```
mars1024: Another small question, why do we need GATEWAY in this way? I don't see the other two ways need `gateway` info.

plugins/ipam/static/README.md
```diff
@@ -60,3 +60,9 @@ The plugin also support following [capability argument](https://github.com/conta
...
+
+1. [args conventions](https://github.com/containernetworking/cni/blob/master/CONVENTIONS.md#args-in-network-config)
+1. [capability argument](https://github.com/containernetworking/cni/blob/master/CONVENTIONS.md)
```
dcbw: @s1061123 Casey and I think capability args should be #1, then args, then CNI_ARGS. Does that look OK?

squeed:
Can you add some tests for this? Thanks!

plugins/ipam/static/README.md
s1061123: fixed.

plugins/ipam/static/main.go
s1061123: Thank you for your comment! I assume that user already configured GATEWAY in its CNI config, however, current static ipam supports GATEWAY as CNI_ARG, hence I just keep it.

plugins/ipam/static/README.md
s1061123: Sure. Changed.

s1061123:
@squeed thank you for your comment! I added a testcase.

plugins/ipam/static/main.go
s1061123: agree.

bboreham:
Am I right in thinking there was a priority order before, and this PR changes the ordering?
Just wanted to check how we announce this in release notes, etc.

squeed:
Given that this is a (small) change in behavior, we should probably bump the version when we next do a release.

PR #401, opened 2019-10-17
testutils: newNS() works in a rootless user namespace

giuseppe:
When running in a user namespace created by an unprivileged user the
owner of /var/run will be reported as the unknown user (as defined in
/proc/sys/kernel/overflowuid) so any access to the directory will
fail.

If the XDG_RUNTIME_DIR environment variable is set, check whether the
current user is also the owner of /var/run.  If the owner is different
than the current user, use the $XDG_RUNTIME_DIR/netns directory.

Signed-off-by: Giuseppe Scrivano <gscrivan@redhat.com>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: go
 sudo: required
-dist: trusty
+dist: xenial
 
 go:
   - 1.11.x
--- a/pkg/testutils/netns_linux.go
+++ b/pkg/testutils/netns_linux.go
@@ -22,16 +22,35 @@ import (
 	"runtime"
 	"strings"
 	"sync"
+	"syscall"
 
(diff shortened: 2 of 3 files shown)


giuseppe:
@dcbw PTAL

squeed:
Makes sense - is there any documentation about this?

giuseppe:
I've added some tests to travis for running as non privileged user.

I needed to upgrade the version of Ubuntu to Xenial as we need `unshare -r` for creating the user namespace.

giuseppe:
> Makes sense - is there any documentation about this?

I can add that.  What would the best place be?

bboreham:
I found a detailed description of `XDG_RUNTIME_DIR` at https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html

PR #403, opened 2019-10-18
Sending GratuitousArp in case of MAC address update

s1061123:
This change sends gratuitous ARP when MAC address is changed to
let other devices to know the MAC address update.

### DIFF
--- a/plugins/meta/tuning/tuning.go
+++ b/plugins/meta/tuning/tuning.go
@@ -25,6 +25,7 @@ import (
 	"path/filepath"
 	"strings"
 
+	"github.com/j-keck/arping"
 	"github.com/vishvananda/netlink"
 
 	"github.com/containernetworking/cni/pkg/skel"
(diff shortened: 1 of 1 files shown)


squeed:
very good idea, :+1:

plugins/meta/tuning/tuning.go
```diff
@@ -208,6 +209,17 @@ func cmdAdd(args *skel.CmdArgs) error {
...
 				return err
 			}
+
+			containerIF, err := net.InterfaceByName(args.IfName)
+			if err != nil {
+				return err
+			}
+			for _, ipc := range result.IPs {
+				if ipc.Version == "4" {
+					_ = arping.GratuitousArpOverIface(ipc.Address.IP, *containerIF)
```
mars1024: lgtm except a nit: maybe `arping.GratuitousArpOverIfaceByName` is better here

PR #405, opened 2019-10-25
vlan: add MTU validation to loadNetConf

### LINKED ISSUE #393: improve MTU assignment and validation

mars1024:
Try to fix #393 

Same like #404 , vlan plugin has the same problem if we want to create a vlan interface from a parent(master) interface with greater MTU.

```
[root@test ~]# ip link show eth2
11: eth2: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc pfifo_fast state UP mode DEFAULT group default qlen 1000
    link/ether fa:f2:3b:5a:33:02 brd ff:ff:ff:ff:ff:ff
[root@test ~]# ip link add link eth2 name eth2.100 mtu 2000 type vlan id 100
RTNETLINK answers: Invalid argument
[root@test ~]# ip link add link eth2 name eth2.100 mtu 1200 type vlan id 100
[root@test ~]# ip link show eth2.100
15: eth2.100@eth2: <BROADCAST,MULTICAST> mtu 1200 qdisc noop state DOWN mode DEFAULT group default qlen 1000
link/ether fa:f2:3b:5a:33:02 brd ff:ff:ff:ff:ff:ff
```

### DIFF
--- a/plugins/main/vlan/vlan.go
+++ b/plugins/main/vlan/vlan.go
@@ -58,7 +58,25 @@ func loadConf(bytes []byte) (*NetConf, string, error) {
 	if n.VlanId < 0 || n.VlanId > 4094 {
 		return nil, "", fmt.Errorf("invalid VLAN ID %d (must be between 0 and 4095 inclusive)", n.VlanId)
 	}
+
+	// check existing and MTU of master interface
+	masterMTU, err := getMTUByName(n.Master)
+	if err != nil {
+		return nil, "", err
+	}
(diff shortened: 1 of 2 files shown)


squeed:
:+1:

Issue #370, opened 2019-08-13
failed to create bridge "cni0": could not add "cni0": operation not supported


ChenQi1989:
I'm trying to setup k8s using flannel. But I'm blocked by this error. Hope someone could give me some suggestion.

I used the following commands:
1. swapoff -a
2. kubeadm init --pod-network-cidr=10.244.0.0/16
3. sysctl net.bridge.bridge-nf-call-iptables=1
4. mkdir -p $HOME/.kube
    cp -i /etc/kubernetes/admin.conf $HOME/.kube/config
    chown $(id -u):$(id -g) $HOME/.kube/config
5. kubectl apply -f https://raw.githubusercontent.com/coreos/flannel/master/Documentation/kube-flannel.yml

I also added some output in kubelet and link_linux.go to help debugging. But I'm now blocked here. Below are the related logs. Any idea?

Aug 13 03:30:39 yocto-k8s-master kubelet[9968]: I0813 03:30:39.018857    9968 raw_exec.go:40] chenqi: raw_exec.ExecPlugin &{/opt/cni/bin/flannel [/opt/cni/bin/flannel] [PATH=/usr/local/sbin:/usr/local/bin:/usr
/sbin:/usr/bin:/sbin:/bin INVOCATION_ID=ed7608a7f620418695c029825e3d374d CNI_ARGS=IgnoreUnknown=1;K8S_POD_NAMESPACE=kube-system;K8S_POD_NAME=coredns-5f7fc64c95-x4l8t;K8S_POD_INFRA_CONTAINER_ID=ce1302be7aedaf89
32763b461a287fa5437a3889580fa615c7422f43f65d4c54 KUBELET_KUBEADM_ARGS=--cgroup-driver=systemd --network-plugin=cni --pod-infra-container-image=k8s.gcr.io/pause:3.1 --resolv-conf=/run/systemd/resolve/resolv.con
f CNI_IFNAME=eth0 LANG=C JOURNAL_STREAM=8:3119020 KUBELET_KUBECONFIG_ARGS=--bootstrap-kubeconfig=/etc/kubernetes/bootstrap-kubelet.conf --kubeconfig=/etc/kubernetes/kubelet.conf CNI_CONTAINERID=ce1302be7aedaf8
[...]

PR #408, opened 2019-11-04
ensure iptables chain creation is idempotent

tgross:
Concurrent use of the `portmap` and `firewall` plugins can result in errors during iptables chain creation:

- The `portmap` plugin has a time-of-check-time-of-use race where it checks for existence of the chain but the operation isn't atomic.
- The `firewall` plugin doesn't check for existing chains and just returns an error.

This commit makes both operations idempotent by creating the chain and then discarding the error if it's caused by the chain already existing. I've added a test for idempotent creation in the `firewall` plugin but the `portmap` plugin already had one so I've left that in place.

### DIFF
--- a/pkg/utils/iptables.go
+++ b/pkg/utils/iptables.go
@@ -1,0 +1,64 @@
+// Copyright 2017 CNI authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package utils
+
+import (
+	"errors"
+	"fmt"
+
+	"github.com/coreos/go-iptables/iptables"
+)
+
+const statusChainExists = 1
+
+// EnsureChain idempotently creates the iptables chain. It does not
+// return an error if the chain already exists.
+func EnsureChain(ipt *iptables.IPTables, table, chain string) error {
+	if ipt == nil {
+		return errors.New("failed to ensure iptable chain: IPTables was nil")
+	}
+	exists, err := ChainExists(ipt, table, chain)
+	if err != nil {
+		return fmt.Errorf("failed to list iptables chains: %v", err)
+	}
+	if !exists {
+		err = ipt.NewChain(table, chain)
+		if err != nil {
+			eerr, eok := err.(*iptables.Error)
+			if eok && eerr.ExitStatus() != statusChainExists {
+				return err
+			}
+		}
+	}
+	return nil
+}
+
+// ChainExists checks whether an iptables chain exists.
+func ChainExists(ipt *iptables.IPTables, table, chain string) (bool, error) {
+	if ipt == nil {
+		return false, errors.New("failed to check iptable chain: IPTables was nil")
+	}
+	chains, err := ipt.ListChains(table)
+	if err != nil {
+		return false, err
+	}
+
+	for _, ch := range chains {
+		if ch == chain {
+			return true, nil
+		}
+	}
+	return false, nil
+}
--- a/pkg/utils/iptables_test.go
+++ b/pkg/utils/iptables_test.go
@@ -1,0 +1,78 @@
+// Copyright 2017-2018 CNI authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package utils
+
+import (
+	"fmt"
+	"math/rand"
+	"runtime"
+
+	"github.com/containernetworking/plugins/pkg/ns"
+	"github.com/containernetworking/plugins/pkg/testutils"
+	"github.com/coreos/go-iptables/iptables"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+const TABLE = "filter" // We'll monkey around here
+
+var _ = Describe("chain tests", func() {
+	var testChain string
+	var ipt *iptables.IPTables
+	var cleanup func()
+
+	BeforeEach(func() {
+
+		// Save a reference to the original namespace,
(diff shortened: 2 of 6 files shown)


tgross:
While it wasn't the intent of this PR, this may fix https://github.com/containernetworking/plugins/issues/370, which is bubbling up a similar error condition as we saw in https://github.com/hashicorp/nomad/issues/6567

plugins/meta/portmap/chain.go
```diff
@@ -35,16 +35,17 @@ type chain struct {
 
 // setup idempotently creates the chain. It will not error if the chain exists.
 func (c *chain) setup(ipt *iptables.IPTables) error {
-	// create the chain
-	exists, err := chainExists(ipt, c.table, c.name)
-	if err != nil {
+
+	err := ipt.NewChain(c.table, c.name)
+	eerr, eok := err.(*iptables.Error)
```
mars1024: Type cast should be done when `err != nil`, right?

plugins/meta/portmap/chain.go
```diff
@@ -35,16 +35,17 @@ type chain struct {
 
 // setup idempotently creates the chain. It will not error if the chain exists.
 func (c *chain) setup(ipt *iptables.IPTables) error {
-	// create the chain
-	exists, err := chainExists(ipt, c.table, c.name)
```
mars1024: I think `get` is cheaper than `create`, so maybe it's better to keep these lines...

plugins/meta/firewall/iptables.go
```diff
@@ -43,7 +43,16 @@ func ensureChain(ipt *iptables.IPTables, table, chain string) error {
 		}
 	}
 
-	return ipt.NewChain(table, chain)
+	err = ipt.NewChain(table, chain)
```
mars1024: LGTM for your intention and implementation, but IMO, idempotency should be done in `iptables` package because decoupling is important for us...  cc @squeed  waiting for your suggestion...

plugins/meta/portmap/chain.go
tgross: We don't strictly need the check (ref https://play.golang.org/p/uSRCkdUHKni) but if you think it'd read nicer if I did, I can do that.

plugins/meta/portmap/chain.go
tgross: Sure, will do.

plugins/meta/firewall/iptables.go
tgross: It could be done there, sure. But the `iptables` package is a fairly bare-bones wrapper around the command line tools. They also don't have any documentation promising idempotency, whereas `containernetworking/plugins` does.

plugins/meta/portmap/chain.go
tgross: I've updated these sections to break early on nil errors and get rid of the switch.

plugins/meta/portmap/chain.go
tgross: Done

plugins/meta/firewall/iptables.go
squeed: Well, I know the maintainer :-).

plugins/meta/portmap/chain.go
```diff
@@ -35,14 +35,19 @@ type chain struct {
...
+
 	if !exists {
-		if err := ipt.NewChain(c.table, c.name); err != nil {
-			return err
+		err := ipt.NewChain(c.table, c.name)
+		if err != nil {
+			eerr, eok := err.(*iptables.Error)
+			if eok && eerr.ExitStatus() != 1 {
+				return err
+			}
```
bboreham: Can we wrap this up in a function so we don't repeat the lines twice, and also so the function name would be descriptive on what we're checking and why it's ok to ignore the error?

dcbw: Maybe put this function into a new pkg/util/iptables.go file and then we can use it from both plugins easily, since they both need to do the same thing?

mars1024: And it's better to make `1` a constant, seems more meaningful..

tgross: Sounds good.

There's probably a few other places where iptables usage overlaps between these two plugins as well as `ipmasq` and `bridge`. But maybe for sake of keeping the changeset reasonable, I can do pull just these out to `pkg/util/iptables.go` and then I can come back in a later PR to do that sort of refactoring?

tgross:
I think at this point I've resolved the items from code review, but I think there's still the open question to @squeed about whether this should be handled upstream in go-iptables instead.

pkg/utils/iptables.go
```diff
@@ -0,0 +1,57 @@
...
+
+	"github.com/coreos/go-iptables/iptables"
+)
+
+const statusChainExists = 1
+
+// EnsureChain idempotently creates the iptables chain. It does not
+// return an error if the chain already exists.
+func EnsureChain(ipt *iptables.IPTables, table, chain string) error {
+	exists, err := ChainExists(ipt, table, chain)
```
mars1024: Should we add a pointer check here to avoid unexpected np error?

pkg/utils/iptables_test.go
```diff
@@ -0,0 +1,76 @@
...
+		testChain = fmt.Sprintf("cni-test-%d", rand.Intn(10000000))
+
+		ipt, err = iptables.NewWithProtocol(iptables.ProtocolIPv4)
+		Expect(err).NotTo(HaveOccurred())
+
+		runtime.LockOSThread()
+		err = testNs.Set()
+		Expect(err).NotTo(HaveOccurred())
+
+		cleanup = func() {
```
mars1024: Better but not necessary to put `cleanup ` in `AfterEach`

plugins/meta/firewall/iptables.go
```diff
@@ -33,6 +34,7 @@ func getPrivChainRules(ip string) [][]string {
 }
 
 func ensureChain(ipt *iptables.IPTables, table, chain string) error {
```
mars1024: If we switch to the utils package, this function should be deprecated.

pkg/utils/iptables.go
```diff
@@ -0,0 +1,57 @@
...
+				return err
+			}
+		}
+	}
+	return nil
+}
+
+// ChainExists checks whether an iptables chain exists.
+func ChainExists(ipt *iptables.IPTables, table, chain string) (bool, error) {
+	chains, err := ipt.ListChains(table)
```
mars1024: ditto

pkg/utils/iptables_test.go
tgross: Done.

pkg/utils/iptables.go
tgross: Done.

plugins/meta/firewall/iptables.go
tgross: Oops! Fixed.

pkg/utils/iptables.go
tgross: Done

plugins/meta/portmap/chain.go
tgross: Done.

squeed:
lgtm!

PR #409, opened 2019-11-11
integration: fix ip address collision in integration tests

squeed:


mars1024:
After reviewing, merged already :)  late lgtm

PR #414, opened 2019-11-20
bring in a MAC address generator for CNI plugins

### LINKED ISSUE #407: Do we need a unified MAC generator for container interface?
Now I'm working on a tool to defend MAC flooding attack from container, I want to implement this based on ebtables on linux bridge.
But because of irregular MAC addresses generated by kernel, there will be a large mount of ebtables rules.
So I think maybe we need a unified MAC generator for our L2 CNI plugins(macvlan, vlan, ipvlan, etc) to generate regular MAC addresses(with same prefix) based on a certain rule, this will help us to recognize/locate the MACs of the container interfaces with a "regular" expression.

mars1024:
The first PR to resolve #407  , two things have been done in this PR
1. bring a special OUI for CNI
2. deprecate old MAC generator and create a new one
    - support dual stack
    - support CNI OUI

If this PR is merged, I'll try to introduce this generator to plugins, like macvlan, ptp, bridge ...

### DIFF
--- a/pkg/ip/link_linux.go
+++ b/pkg/ip/link_linux.go
@@ -220,28 +220,21 @@ func DelLinkByNameAddr(ifName string) ([]*net.IPNet, error) {
 	return out, nil
 }
 
-func SetHWAddrByIP(ifName string, ip4 net.IP, ip6 net.IP) error {
+// SetHardwareAddress will set the interface to a specified MAC address,
+// if the address is not assigned, we will generate a random CNI-OUI MAC
+// address for the interface.
+func SetHardwareAddress(ifName string, hwAddr net.HardwareAddr) error {
 	iface, err := netlink.LinkByName(ifName)
 	if err != nil {
 		return fmt.Errorf("failed to lookup %q: %v", ifName, err)
 	}
 
-	switch {
-	case ip4 == nil && ip6 == nil:
-		return fmt.Errorf("neither ip4 or ip6 specified")
+	if hwAddr == nil {
+		hwAddr = hwaddr.GenerateMAC()
+	}
 
-	case ip4 != nil:
-		{
-			hwAddr, err := hwaddr.GenerateHardwareAddr4(ip4, hwaddr.PrivateMACPrefix)
-			if err != nil {
-				return fmt.Errorf("failed to generate hardware addr: %v", err)
-			}
-			if err = netlink.LinkSetHardwareAddr(iface, hwAddr); err != nil {
-				return fmt.Errorf("failed to add hardware addr to %q: %v", ifName, err)
-			}
-		}
-	case ip6 != nil:
-		// TODO: IPv6
+	if err = netlink.LinkSetHardwareAddr(iface, hwAddr); err != nil {
+		return fmt.Errorf("failed to set hardware address for %s: %v", ifName, err)
 	}
 
 	return nil
--- a/pkg/ip/link_linux.go
+++ b/pkg/ip/link_linux.go
@@ -230,7 +230,10 @@ func SetHardwareAddress(ifName string, hwAddr net.HardwareAddr) error {
 	}
 
 	if hwAddr == nil {
-		hwAddr = hwaddr.GenerateMAC()
+		hwAddr, err = hwaddr.GenerateMAC()
+		if err != nil {
+			return fmt.Errorf("failed to generate random CNI-OUI MAC address: %v", err)
+		}
 	}
 
 	if err = netlink.LinkSetHardwareAddr(iface, hwAddr); err != nil {
--- a/pkg/utils/hwaddr/hwaddr.go
+++ b/pkg/utils/hwaddr/hwaddr.go
@@ -15,49 +15,27 @@
 package hwaddr
 
 import (
-	"fmt"
+	"crypto/rand"
 	"net"
 )
 
-const (
-	ipRelevantByteLen      = 4
-	PrivateMACPrefixString = "0a:58"
-)
+// The first 24 bits of the MAC represent the Organizationally Unique Identifier (OUI),
+// and the first byte of the MAC address has two special bits.
+// 1. The least-significant bit: 0 for unicast and 1 for multicast
+// 2. The second-least-significant bit: 0 for globally unique and 1 for locally administered
+// Since MAC address for container interfaces is locally administered and unicast, so we can
+// fix the two LSb of MSB to 10 and use any values for other bits. This will avoid any conflicts
+// with public OUIs.
+// We use 02:58:00 for CNI official OUI here.
+var cniOUI = []byte{0x02, 0x58, 0x00}
 
-var (
-	// private mac prefix safe to use
-	PrivateMACPrefix = []byte{0x0a, 0x58}
-)
-
-type SupportIp4OnlyErr struct{ msg string }
-
-func (e SupportIp4OnlyErr) Error() string { return e.msg }
-
-type MacParseErr struct{ msg string }
-
-func (e MacParseErr) Error() string { return e.msg }
-
-type InvalidPrefixLengthErr struct{ msg string }
-
-func (e InvalidPrefixLengthErr) Error() string { return e.msg }
-
-// GenerateHardwareAddr4 generates 48 bit virtual mac addresses based on the IP4 input.
-func GenerateHardwareAddr4(ip net.IP, prefix []byte) (net.HardwareAddr, error) {
-	switch {
-
-	case ip.To4() == nil:
-		return nil, SupportIp4OnlyErr{msg: "GenerateHardwareAddr4 only supports valid IPv4 address as input"}
-
-	case len(prefix) != len(PrivateMACPrefix):
-		return nil, InvalidPrefixLengthErr{msg: fmt.Sprintf(
-			"Prefix has length %d instead  of %d", len(prefix), len(PrivateMACPrefix)),
-		}
-	}
-
-	ipByteLen := len(ip)
-	return (net.HardwareAddr)(
-		append(
-			prefix,
-			ip[ipByteLen-ipRelevantByteLen:ipByteLen]...),
-	), nil
+// GenerateMAC will generate MAC addresses with fixed first 24 bits (CNI OUI), and the last 24 bits
+// will be random, so there are at most 16777216(2^24) different addresses.
+// To avoid MAC address collision, we suggest to use this function when length of subnet masking is
+// greater or equal than 20 in IPAM.
+func GenerateMAC() net.HardwareAddr {
+	hw := make(net.HardwareAddr, 6)
+	copy(hw[:3], cniOUI)
+	_, _ = rand.Read(hw[3:])
+	return hw
 }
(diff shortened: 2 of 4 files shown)


bboreham:
I have no objection to the change; having a consistent prefix seems mildly useful.
However this PR does remove some public functions `GenerateHardwareAddr4` and `SetHWAddrByIP` so we should note this in the release notes for next release.

mars1024:
> I have no objection to the change; having a consistent prefix seems mildly useful.
> However this PR does remove some public functions `GenerateHardwareAddr4` and `SetHWAddrByIP` so we should note this in the release notes for next release.

Of course, the release notes should include this if this merged.

pkg/ip/link_linux.go
```diff
@@ -220,28 +220,21 @@ func DelLinkByNameAddr(ifName string) ([]*net.IPNet, error) {
...
-			if err != nil {
-				return fmt.Errorf("failed to generate hardware addr: %v", err)
-			}
-			if err = netlink.LinkSetHardwareAddr(iface, hwAddr); err != nil {
-				return fmt.Errorf("failed to add hardware addr to %q: %v", ifName, err)
-			}
-		}
-	case ip6 != nil:
-		// TODO: IPv6
+	if err = netlink.LinkSetHardwareAddr(iface, hwAddr); err != nil {
```
squeed: If the mac is randomly generated, can you retry if there's a duplicate? This is going to be sort of tough, since I don't think the kernel stops you from creating devices with duplicate MACs and you might need to scan every namespace.

squeed:
Hey @dcbw do you know if the kernel's MAC generator has any sort of duplicate detection?

pkg/ip/link_linux.go
mars1024: Unfortunately I don't think there is a good way to check MAC collision.. And all the plugins are generating MACs randomly now, so the collision risk is actually existing.
After the first three bytes are fixed, the probability of collision becomes larger. As my comments above, I did some tests on this and gave a **safe** usage suggestion.

pkg/utils/hwaddr/hwaddr.go
```diff
@@ -15,49 +15,27 @@
...
-		append(
-			prefix,
-			ip[ipByteLen-ipRelevantByteLen:ipByteLen]...),
-	), nil
+// The first 24 bits of the MAC represent the Organizationally Unique Identifier (OUI),
+// and the first byte of the MAC address has two special bits.
+// 1. The least-significant bit: 0 for unicast and 1 for multicast
+// 2. The second-least-significant bit: 0 for globally unique and 1 for locally administered
+// Since MAC address for container interfaces is locally administered and unicast, so we can
+// fix the two LSb of MSB to 10 and use any values for other bits. This will avoid any conflicts
```
jellonek: Minor typo: LSB

dcbw:
> Hey @dcbw do you know if the kernel's MAC generator has any sort of duplicate detection?

There is none. All kernel drivers use:

```
/**
 * eth_random_addr - Generate software assigned random Ethernet address
 * @addr: Pointer to a six-byte array containing the Ethernet address
 *
 * Generate a random Ethernet address (MAC) that is not multicast
 * and has the local assigned bit set.
 */
static inline void eth_random_addr(u8 *addr)
{
        get_random_bytes(addr, ETH_ALEN);
        addr[0] &= 0xfe;        /* clear multicast bit */
        addr[0] |= 0x02;        /* set local assignment bit (IEEE802) */
}
```

pkg/ip/link_linux.go
dcbw: The kernel generates random addresses and does not check for collisions.

bboreham: Applying "Birthday Problem" analysis, 24 bits of randomness means that with 4096 containers we'd have a 50% chance of a collision.  This seems too low to be comfortable.

The current situation, letting Linux create random addresses, has >40 bits of randomness so far lower probability.

bboreham:
I think the point about potential collisions means we could not make this behaviour the default for commonly-used plugins (bridge, etc).
It could be added as an option, in which case maybe the OUI could be specified in the option.

pkg/utils/hwaddr/hwaddr.go
mars1024: emmm, actually LSb here means least-significant bit, not byte @jellonek

jellonek: Ough. Right!

mars1024:
Sorry for the late reply,  @bboreham , here is a [birthday problem test](https://play.golang.org/p/IzD4sDweGrH), we can see the results as below
```
subnet prefix 24 birthday collision potential 0.001944 
subnet prefix 23 birthday collision potential 0.007767 
subnet prefix 22 birthday collision potential 0.030737 
subnet prefix 21 birthday collision potential 0.117449 
subnet prefix 20 birthday collision potential 0.393395 
subnet prefix 19 birthday collision potential 0.864632 
subnet prefix 18 birthday collision potential 0.999664 
subnet prefix 17 birthday collision potential 1.000000 
subnet prefix 16 birthday collision potential 1.000000 
```
So I think maybe **22** is a safe value for real usage, this means that in terms of probability we can have 32 subnets (prefix==22) without MAC collision.
Of course, just as your suggestion, this feature will not be used as a default choice for our common plugins, it should be optional.

I've force-pushed to update the recommended safe subnet prefix to 22, do you think there is something else to do before merging this, @bboreham ?

pkg/utils/hwaddr/hwaddr.go
```diff
@@ -15,49 +15,27 @@
 package hwaddr
 
 import (
-	"fmt"
+	"math/rand"
```
aojea: do you foresee any problem with the deterministic behavior of "math/rand/ vs "crypt/rand"?
https://flaviocopes.com/go-random/

pkg/ip/link_linux.go
aojea: mac collision only impacts if the containers are in the same layer 2 segments, right?
IIRC the max number of interfaces per linux bridge is 1024, another thing is stretched layer2 domains, but I don't think this applies here

pkg/utils/hwaddr/hwaddr.go
```diff
@@ -15,49 +15,27 @@
...
+// Since MAC address for container interfaces is locally administered and unicast, so we can
+// fix the two LSb of MSB to 10 and use any values for other bits. This will avoid any conflicts
+// with public OUIs.
+// We use 02:58:00 for CNI official OUI here.
+var cniOUI = []byte{0x02, 0x58, 0x00}
+
+// GenerateMAC will generate MAC addresses with fixed first 24 bits (CNI OUI), and the last 24 bits
+// will be random, so there are at most 16777216(2^24) different addresses.
+// To avoid MAC address collision, we suggest to use this function when length of subnet masking is
+// greater or equal than 22 in IPAM.
```
aojea: Nit, I think that this comment is a bit misleading because the underlying problem is the number of interfaces, and that number is directed related to the number of ip addresses. 
The number of IP addresses depends on the subnet size, but is not a 1to1 relation, I mean a /16 doesn't have to have all the addresses used.
Also, if we are using a local bridge the limit is imposed by the kernel to 1024, funny thing is that matches the /22 requirements :smile:

pkg/utils/hwaddr/hwaddr.go
```diff
@@ -15,49 +15,27 @@
...
+func GenerateMAC() net.HardwareAddr {
+	hw := make(net.HardwareAddr, 6)
+	copy(hw[:3], cniOUI)
+	_, _ = rand.Read(hw[3:])
```
s1061123: rand.Seed() is not invoked before rand.Read(). Without 'rand.Seed()' call, rand.Read() uses 'rand.Seed(1)' as you see in [document](https://golang.org/pkg/math/rand/#Seed), hence it is not random and cracker could infer the next MAC address. Call rand.Seed() in plugin side can be the 
possible solution, but it may be missed, so how about to add rand.Seed() in init()?

mars1024: Thanks for the good catch! And @aojea has provided another way above, switching `math/rand` to `crypto/rand`, which will make MAC addresses more encrypted and secure. More important, not rely on `Seed` anymore.

pkg/utils/hwaddr/hwaddr.go
mars1024: Thanks! `crypto/rand` seems better here.

pkg/ip/link_linux.go
mars1024: Emmm... This is public utility package, so maybe this will be called in other scenarios without 1024 max interfaces limit. But because of the limit, this is safe for CNI common plugins, right?

pkg/utils/hwaddr/hwaddr.go
mars1024: done!

pkg/ip/link_linux.go
aojea: yeah, I just wanted to point out that despite mac collision is a real problem, there are other more important limitations like this max number of interfaces per bridge or BUM traffic.
A layer 2 network with 4096 devices must be unmanageable, I actually don't know if all HW devices will be able to have a mac-table size that is able to host such amount of macs.

mars1024: yeap, got your point, the safe subnet prefix (22) is just for limiting cases, but in common cases, we don't need to care much about this.

pkg/utils/hwaddr/hwaddr.go
mars1024: updated!

aojea:
LGTM

mars1024:
CI failure fixed, I think this is ready to go :)

pkg/utils/hwaddr/hwaddr.go
```diff
@@ -15,49 +15,29 @@
...
+// Since MAC address for container interfaces is locally administered and unicast, so we can
+// fix the two LSb of MSB to 10 and use any values for other bits. This will avoid any conflicts
+// with public OUIs.
+// We use 02:58:00 for CNI official OUI here.
+var cniOUI = []byte{0x02, 0x58, 0x00}
+
+// GenerateMAC will generate MAC addresses with fixed first 24 bits (CNI OUI), and the last 24 bits
+// will be random, so there are at most 16777216(2^24) different addresses.
+// To avoid MAC address collision as much as possible, this function is suggested to be used within
+// the limitation that each subnet should contain no more 1024(2^10) using IP addresses.
```
bboreham, line 35: I'm a little lost why this is talking about subnets.  Surely the requirement is that there be no collision in an L2 domain?

matthewdupre:
I don't think these collision probabilities are acceptable.  Even a user running only 200 containers has a 0.1% chance of a collision.

I suspect we need at least 32 bits here.

mars1024:
> I don't think these collision probabilities are acceptable. Even a user running only 200 containers has a 0.1% chance of a collision.
> 
> I suspect we need at least 32 bits here.

But OUI will take 24 bits as definition, so only 24 bits left, I think if the collision probabilities are not acceptable, I'm sorry to close this PR... 

cc @bboreham @squeed @dcbw , more comments?

pkg/utils/hwaddr/hwaddr.go
mars1024: yeah, subnet here means a single L2 domain, do you have a more proper word to be put here?

PR #415, opened 2019-11-20
install jq in vagrantfile

zjj2wry:


zjj2wry:
https://github.com/containernetworking/cni/blob/master/scripts/exec-plugins.sh#L23

This script uses the jq command, which can be installed in advance in the vagrantfile to help complete the cni example demo.

PR #416, opened 2019-11-22
support IPs runtime config in host-local

### LINKED ISSUE #411: Support 'ips' runtimeConfig with format <ip>[/<prefix>]
Now host-local can support IP assignment through CNI_ARGS and args in network config, but runtime config seems more flexible than the two ways above.
So I think we need to introduce 'ips' runtimeConfig to host-local plugin.

mars1024:
try to fix #411

### DIFF
--- a/plugins/ipam/host-local/backend/allocator/config.go
+++ b/plugins/ipam/host-local/backend/allocator/config.go
@@ -29,8 +29,10 @@ type Net struct {
 	Name          string      `json:"name"`
 	CNIVersion    string      `json:"cniVersion"`
 	IPAM          *IPAMConfig `json:"ipam"`
-	RuntimeConfig struct {    // The capability arg
+	RuntimeConfig struct {
+		// The capability args
 		IPRanges []RangeSet `json:"ipRanges,omitempty"`
+		IPs      []net.IP   `json:"ips,omitempty"`
 	} `json:"runtimeConfig,omitempty"`
 	Args *struct {
 		A *IPAMArgs `json:"cni"`
@@ -48,7 +50,7 @@ type IPAMConfig struct {
 	DataDir    string         `json:"dataDir"`
 	ResolvConf string         `json:"resolvConf"`
 	Ranges     []RangeSet     `json:"ranges"`
-	IPArgs     []net.IP       `json:"-"` // Requested IPs from CNI_ARGS and args
+	IPArgs     []net.IP       `json:"-"` // Requested IPs from CNI_ARGS, args and capabilities
 }
 
 type IPAMEnvArgs struct {
@@ -80,7 +82,8 @@ func LoadIPAMConfig(bytes []byte, envArgs string) (*IPAMConfig, string, error) {
 		return nil, "", fmt.Errorf("IPAM config missing 'ipam' key")
 	}
 
-	// Parse custom IP from both env args *and* the top-level args config
+	// Parse custom IP from env args, the top-level args config and capabilities
+	// in runtime configuration
 	if envArgs != "" {
 		e := IPAMEnvArgs{}
 		err := types.LoadArgs(envArgs, &e)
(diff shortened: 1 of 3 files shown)


mars1024:
Actually all the tests have passed, but the progress is still pending... something is wrong with travis maybe...

plugins/ipam/host-local/backend/allocator/config.go
```diff
@@ -48,7 +50,7 @@ type IPAMConfig struct {
 	DataDir    string         `json:"dataDir"`
 	ResolvConf string         `json:"resolvConf"`
 	Ranges     []RangeSet     `json:"ranges"`
-	IPArgs     []net.IP       `json:"-"` // Requested IPs from CNI_ARGS and args
+	IPArgs     []net.IP       `json:"-"` // Requested IPs from CNI_ARGS, args and capabilities
```
matthewdupre: It looks like the ips convention describes itself as being a list of CIDRs (don't like the name!), which can't go in net.IPs.  It might be that the convention is written wrongly, but something needs to change.

mars1024: Good question! Thanks @matthewdupre , I'll check this later ~

mars1024: @matthewdupre  Sorry for the late reply, I checked CNI convention and found that, for CNI_ARGS and args in network config, "ips" follow the spec as
```
Spec:
"ips": ["<ip>[/<prefix>]", ...]
Examples:
"ips": ["10.2.2.42/24", "2001:db8::5"]
The plugin may require the IP address to include a prefix length.
```
So I think ips can be filled with both IP and CIDR format.
But the spec of ips capability in runtime config only show a CIDR format, so maybe we should update the spec after merging this.

dcbw: @matthewdupre conventions say prefix is optional; but I really think host-local should accept prefix as well (and validate against ranges) for consistency with the static plugin.

dcbw:
@mars1024 FWIW I had started working on this a long time ago. Maybe my old branch will be useful. https://github.com/dcbw/plugins/tree/ip-prefix

Code attempts to validate that the given CNI_ARGS/args IP + prefix is contained within one of the configured ranges.

plugins/ipam/host-local/backend/allocator/config.go
mars1024: @matthewdupre Waiting for your reply, then I'll update this PR.

mars1024:
> @mars1024 FWIW I had started working on this a long time ago. Maybe my old branch will be useful. https://github.com/dcbw/plugins/tree/ip-prefix
> 
> Code attempts to validate that the given CNI_ARGS/args IP + prefix is contained within one of the configured ranges.

@dcbw  Thanks, BTW, do you remember why we choose to use IP format instead of CIDR format in `host-local` plugin?

plugins/ipam/host-local/backend/allocator/config.go
squeed: Yeah, I agree with @dcbw; host-local should accept a host+prefix (and validate that it matches an existing range), so that it matches static-ip.

mars1024: OK, I will do this soon, thanks for all your suggestions.

mars1024: I read about the related lines in `static` plugin, it seems that `static` plugin only support CIDR format (ip with prefix) as the line [https://github.com/containernetworking/plugins/blob/655116585396cc26ca973db2d5daef66d821ad90/plugins/ipam/static/main.go#L216](https://github.com/containernetworking/plugins/blob/655116585396cc26ca973db2d5daef66d821ad90/plugins/ipam/static/main.go#L216), so I'm wondering whether we should support both IP and CIDR format in `host-local`? @squeed @dcbw 

The `ips` in `static` is more like a whole **IP config** including ip, subnet and gateway, but `ips` in `host-local` is just an **IP** because the other configs (subnet and gateway) can come from ranges.
So I think maybe the two `ips` are different, should we separate them in convention?

mars1024: conclusion: support both IP and CIDR format in `host-local` plugin

squeed: I think it's probably okay to use the same convention for both. You are right, it's slightly awkward. But I think the higher-level intent is "I want this address please" which is the same for both.

dcbw:
@mars1024 is this worth revisiting?

mars1024:
> @mars1024 is this worth revisiting? (I think it is...)

Yeah, of course, I'll come back to this soon...

mars1024 (commented):
Update: I'm trying to implement a new public type `IP` to support both IP with or without prefix.

Issue #418, opened 2019-12-05
Portmap plugin is not idempotent if executed in parallel


aojea:
KIND is running e2e kubernetes jobs in parallel, creating a test environment with high load creating and deleting pods and services constantly. It uses containerd as CRI

It brought my attention that despite the tests are passing there are a considerable number of errors like this (happening in ipv4 and ipv6 jobs, it's independent of the cluster ip family):

```
Dec 04 10:30:42 kind-worker containerd[79]: time="2019-12-04T10:30:42.632259521Z" level=error msg="StopPodSandbox for \"4474f785df637e833b3c4331e24888ba8d50a34cd798cdd041920e09c52bfa1e\" failed" error="failed to destroy network for sandbox \"4474f785df637e833b3c4331e24888ba8d50a34cd798cdd041920e09c52bfa1e\": could not teardown ipv4 dnat: running [/usr/sbin/iptables -t nat -X CNI-DN-2cbdc0366bc65971858cc --wait]: exit status 1: iptables: No chain/target/match by that name.\n"
```

However, my understanding from the portmap plugin code is that this operations should be idempotent and not fail

https://github.com/containernetworking/plugins/blob/655116585396cc26ca973db2d5daef66d821ad90/plugins/meta/portmap/portmap.go#L345

After investigating during some time, I think that the problem maybe caused because there are several several `teardowns`operations in parallel, don't know if this is the expected behaviour in containerd (cc: @Random-Liu ):

```
ec 04 10:17:51 kind-worker containerd[79]: time="2019-12-04T10:17:51.066858228Z" level=info msg="StopPodSandbox for \"2f503a1c0253dda279b863a5a53dfc904eee346800a248440cfe625e89e74b30\""
Dec 04 10:17:51 kind-worker containerd[79]: time="2019-12-04T10:17:51.066934330Z" level=info msg="Container to stop \"361103b7742246501e02e71b1b7d08f08ccd2289e6a11216888398efcd337e87\" must be in running or unknown state, current state \"CONTAINER_EXITED\""
Dec 04 10:17:51 kind-worker containerd[79]: time="2019-12-04T10:17:51.067582275Z" level=info msg="StopPodSandbox for \"2f503a1c0253dda279b863a5a53dfc904eee346800a248440cfe625e89e74b30\""
[...]

PR #420, opened 2019-12-09
idempotent portmap teardown 

aojea:
It turns out that the portmap plugin is not idempotent if its executed in parallel.
The errors are caused due to a race of different iptables executions deleting the chains.
This patch adds a mutex based in a file lock that solves the concurrency problems.

xref: https://github.com/containernetworking/plugins/issues/418

### DIFF
--- a/plugins/meta/portmap/chain.go
+++ b/plugins/meta/portmap/chain.go
@@ -17,11 +17,15 @@ package main
 import (
 	"fmt"
 	"strings"
+	"time"
 
 	"github.com/containernetworking/plugins/pkg/utils"
 	"github.com/coreos/go-iptables/iptables"
+	"github.com/juju/fslock"
 	"github.com/mattn/go-shellwords"
 )
+
+const lockfile = "/tmp/portmap.lock"
 
 type chain struct {
 	table       string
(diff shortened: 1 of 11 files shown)


plugins/meta/portmap/chain.go
```diff
@@ -17,12 +17,16 @@ package main
...
 	"strings"
+	"time"
 
 	"github.com/containernetworking/plugins/pkg/utils"
 	"github.com/coreos/go-iptables/iptables"
+	"github.com/juju/fslock"
 	"github.com/mattn/go-shellwords"
 )
 
+const lockfile = "/tmp/portmap.lock"
```
aojea: is this the right approach?

mars1024:
Related PR #408

aojea:
> Related PR #408

I've added a test to the PR that verifies the idempotency if there are concurrent executions.
I've tried the approach of the mentioned PR but was still flaky and failed constantly without adding a sleep to the test between goroutines. 
The external lock guarantees the atomicity of the operation, as you can see with 10 goroutines there are no errors tearing down the portmap rules

mars1024:
> > Related PR #408
> 
> I've added a test to the PR that verifies the idempotency if there are concurrent executions.
> I've tried the approach of the mentioned PR but was still flaky and failed constantly without adding a sleep to the test between goroutines.
> The external lock guarantees the atomicity of the operation, as you can see with 10 goroutines there are no errors tearing down the portmap rules

Do you have the old PR branch which I can take a look at?
If we want to make a function idempotent, we must make sure every step of this function is idempotent, just like #408, maybe more error handlers are needed.

IMO, global locker should be our last choice, but I believe it will work.

aojea:
@mars1024 I've create a new PR so you can compare https://github.com/containernetworking/plugins/pull/421

PR #421, opened 2019-12-11
Portmap doesn't fail if chain doesn't exist

aojea:
It turns out that the portmap plugin is not idempotent if its executed in parallel.
The errors are caused due to a race of different instantiations deleting the chains.
This patch does that the portmap plugin doesn't fail if the errors are because the chain doesn't exist on teardown.

xref: https://github.com/containernetworking/plugins/issues/418

Signed-off-by: Antonio Ojea <antonio.ojea.garcia@gmail.com>

### DIFF
--- a/plugins/meta/portmap/chain.go
+++ b/plugins/meta/portmap/chain.go
@@ -70,15 +70,8 @@ func (c *chain) teardown(ipt *iptables.IPTables) error {
 	// flush the chain
 	// This will succeed *and create the chain* if it does not exist.
 	// If the chain doesn't exist, the next checks will fail.
-	if err := ipt.ClearChain(c.table, c.name); err != nil {
-		eerr, eok := err.(*iptables.Error)
-		switch {
-		case eok && eerr.IsNotExist():
-			// swallow here, the chain was already deleted
-			return nil
-		default:
-			return err
-		}
+	if err := utils.ClearChain(ipt, c.table, c.name); err != nil {
+		return err
 	}
 
 	for _, entryChain := range c.entryChains {
@@ -97,32 +90,15 @@ func (c *chain) teardown(ipt *iptables.IPTables) error {
 				}
 				chainParts = chainParts[2:] // List results always include an -A CHAINNAME
 
-				if err := ipt.Delete(c.table, entryChain, chainParts...); err != nil {
-					eerr, eok := err.(*iptables.Error)
-					switch {
-					case eok && eerr.IsNotExist():
-						// swallow here, the chain was already deleted
-						continue
-					case eok && eerr.ExitStatus() == 2:
-						// swallow here, invalid command line parameter because the referring rule is missing
-						continue
-					default:
-						return fmt.Errorf("Failed to delete referring rule %s %s: %v", c.table, entryChainRule, err)
-					}
+				if err := utils.DeleteRule(ipt, c.table, entryChain, chainParts...); err != nil {
+					return err
 				}
+
 			}
 		}
 	}
 
-	err := ipt.DeleteChain(c.table, c.name)
-	eerr, eok := err.(*iptables.Error)
-	switch {
-	case eok && eerr.IsNotExist():
-		// swallow here, the chain was already deleted
-		return nil
-	default:
-		return err
-	}
+	return utils.DeleteChain(ipt, c.table, c.name)
 }
 
 // insertUnique will add a rule to a chain if it does not already exist.
--- a/plugins/meta/portmap/chain.go
+++ b/plugins/meta/portmap/chain.go
@@ -71,7 +71,14 @@ func (c *chain) teardown(ipt *iptables.IPTables) error {
 	// This will succeed *and create the chain* if it does not exist.
 	// If the chain doesn't exist, the next checks will fail.
 	if err := ipt.ClearChain(c.table, c.name); err != nil {
-		return err
+		eerr, eok := err.(*iptables.Error)
+		switch {
+		case eok && eerr.IsNotExist():
+			// swallow here, the chain was already deleted
+			return nil
+		default:
+			return err
+		}
 	}
 
 	for _, entryChain := range c.entryChains {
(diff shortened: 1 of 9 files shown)


plugins/meta/portmap/chain.go
```diff
@@ -70,7 +70,10 @@ func (c *chain) teardown(ipt *iptables.IPTables) error {
 	// flush the chain
 	// This will succeed *and create the chain* if it does not exist.
 	// If the chain doesn't exist, the next checks will fail.
-	if err := ipt.ClearChain(c.table, c.name); err != nil {
+	err := ipt.ClearChain(c.table, c.name)
+	eerr, eok := err.(*iptables.Error)
```
mars1024: what if eok is false?

plugins/meta/portmap/chain_test.go
```diff
@@ -195,4 +198,48 @@ var _ = Describe("chain tests", func() {
...
+		Expect(err).NotTo(HaveOccurred())
+		errCh := make(chan error, N)
+		for i := 0; i < N; i++ {
+			wg.Add(1)
+			go func() {
+				defer wg.Done()
+				// Use the testing namespace
+				runtime.LockOSThread()
+				defer runtime.UnlockOSThread()
+				err = testNs.Set()
```
mars1024: Test has already been in test namespace after `BeforeEach`.

aojea: yeah, and it goes back to it because of L67.
The problem is that the new goroutines use the root namespace

plugins/meta/portmap/chain.go
aojea: the if below checks if err!=nil, can eok be false if err=nil?
maybe should I change the if and use eok instead or I'm misunderstanding how eok and err works?

plugins/meta/portmap/chain_test.go
```diff
@@ -195,4 +198,48 @@ var _ = Describe("chain tests", func() {
...
+				Expect(err).NotTo(HaveOccurred())
+				// teardown chain
+				errCh <- testChain.teardown(ipt)
+			}()
+			// wait 1 ms until next execution
+			time.Sleep(1 * time.Millisecond)
```
aojea: if I remove this sleep it fails constantly, with the lock it does not

aojea:
> Line 93 in chain.go `if err := ipt.Delete(c.table, entryChain, chainParts...); err != nil {` should be idempotent.

Do you mean that I should check if there is an error of non existent chain and swallow it?

plugins/meta/portmap/chain.go
mars1024: I mean `err` maybe a normal error instead of an iptables Error, this may happen in [https://github.com/coreos/go-iptables/blob/master/iptables/iptables.go#L453](https://github.com/coreos/go-iptables/blob/master/iptables/iptables.go#L453) .

plugins/meta/portmap/chain_test.go
mars1024: L67 in which file? I'm sorry I didn't find it, could you help to paste a link here?

mars1024:
> > Line 93 in chain.go `if err := ipt.Delete(c.table, entryChain, chainParts...); err != nil {` should be idempotent.
> 
> Do you mean that I should check if there is an error of non existent chain and swallow it?

Yes, that's what I want, and if you could print errs in test, I believe we will locate the root cause exactly.

plugins/meta/portmap/chain.go
aojea: got it, thanks

aojea:
> > > Line 93 in chain.go `if err := ipt.Delete(c.table, entryChain, chainParts...); err != nil {` should be idempotent.
> > 
> > 
> > Do you mean that I should check if there is an error of non existent chain and swallow it?
> 
> Yes, that's what I want, and if you could print errs in test, I believe we will locate the root cause exactly.

I've already found the root cause in #418, and It's easy to reproduce https://github.com/containernetworking/plugins/issues/418#issuecomment-562974220

You can see that it always fail in those line, however, you are right and I've found a small percentage of time that fails in the line you mention

plugins/meta/portmap/chain_test.go
aojea: in L67 of this file there is a `runtime.LockOSThread()` so the test is always in that namespace, but I think that your question is why I have to set again the namespace inside the goroutine.
I'm doing it because is the only way I could make that the goroutine uses the testing namespace, otherwise they always use the root namespace, don't know if there is a better way though

aojea:
@mars1024 PTAL, I added your suggestions and it's working now, I can't reproduce the problem
Thank you for the guidance

plugins/meta/portmap/chain_test.go
squeed: Yes, as of go 1.10 new goroutines always spawn in the host netns. However, this should be a testNs.Do( closure ) to ensure we *leave* the testNS!

aojea:
@squeed please take look to current PR
the failure is because of this error
```      running [/sbin/iptables -t filter -X cni-test-6142911 --wait]: exit status 1: Another app is currently holding the xtables lock; waiting (1s) for it to exit...
      iptables: No chain/target/match by that name.
```
the xtables lock message makes that 
https://github.com/coreos/go-iptables/blob/af017ce15ce0adc4beab89d270abfeeb4d262cf1/iptables/iptables.go#L50-L54

fails to parse correctly the error message, thus it doesn't swallow the error

Should I submit a new patch to go-iptables to consider this scenario? add it to current PR? or is it ok submitting the PR as it is now?

aojea:
@mars1024 @squeed please take a look, I've rebased over the commit that fixes the flaky test and I think that is ready to go

mars1024:
Thanks for your efforts, almost LGTM, and I think it's better to squash some lines into some idempotent utility functions in `pkg/utils/iptables.go` just like #408 .

squeed:
> Should I submit a new patch to go-iptables to consider this scenario? add it to current PR? or is it ok submitting the PR as it is now?

Yes, please! Though, I'm surprised that's failing. It should wait indefinitely. (I have merge rights in go-iptables too).

aojea:
@squeed @mars1024 I think that I've addressed all the remaining comments, can you please take a look again?

squeed:
@aojea I've released go-iptables, can you pick that up in this PR?

aojea:
@squeed here you go, go-iptables upgraded to 0.4.4 with the last changes
by the way, I had a fun time with go1.13 removing some dependencies from the vendor folder :-)

squeed:
lgtm!

aojea:
/hold @squeed @mars1024 there is a bug in the go-iptables code I've introduced, my apologies, we need to bump the iptables code again
I've submitted the fix https://github.com/coreos/go-iptables/pull/72

aojea:
/hold cancel

updated the go-iptables dependency to 0.4.5 with the fix 
@squeed seems we are ready to go :smile:

PR #427, opened 2019-12-18
do DelegateDel if delegateAdd failed

### LINKED ISSUE #426: flannel /var/lib/cni/flannel/cid  not be removed if cni run failed
/var/lib/cni/flannel/xxx will leak if allocate ip failed.
1.https://github.com/containernetworking/plugins/blob/master/plugins/meta/flannel/flannel.go#L168 allocate ip failed,   
2. https://github.com/containernetworking/plugins/blob/master/plugins/meta/flannel/flannel.go#L164 will leak

maybe add `consumeScratchNetConf`  if  `invoke.DelegateAdd`  failed.

hongli-my:
fix #426

### DIFF
--- a/plugins/meta/flannel/flannel_linux.go
+++ b/plugins/meta/flannel/flannel_linux.go
@@ -28,7 +28,7 @@ import (
 	"os"
 )
 
-func doCmdAdd(args *skel.CmdArgs, n *NetConf, fenv *subnetEnv) error {
+func doCmdAdd(args *skel.CmdArgs, n *NetConf, fenv *subnetEnv) (err error) {
 	n.Delegate["name"] = n.Name
 
 	if !hasKey(n.Delegate, "type") {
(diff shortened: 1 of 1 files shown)


plugins/meta/flannel/flannel_linux.go
```diff
@@ -65,6 +65,12 @@ func doCmdAdd(args *skel.CmdArgs, n *NetConf, fenv *subnetEnv) error {
 		},
 	}
 
+	defer func() {
+		if err != nil {
+			doCmdDel(args, n)
```
mars1024: I think this line will never be hit. We need a `defer` in `delegateAdd`.

squeed:
can you add a test for this?

bboreham:
I don't understand why a delegated Del is the right answer here.
Why doesn't the Add side clean up before returning an error?

plugins/meta/flannel/flannel_linux.go
hongli-my: here: https://github.com/containernetworking/plugins/blob/master/plugins/meta/flannel/flannel.go#L170 ?

add here need check `saveScratchNetConf`, `DelegateAdd`, `return result.Print()`  three cases.

mars1024: no, actually we only need to check `err`

dcbw:
Traditionally the runtime (thing calling CNI) is responsible for cleaning up when ADD fails, eg https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/dockershim/docker_sandbox.go#L177 because there may be additional things the runtime does that need to be cleaned up outside of networking.

PR #434, opened 2020-01-09
Fix for the case where kernel doesn't have CONFIG_BRIDGE_VLAN_FILTERING

carlosedp:
If the Linux kernel is not built with the parameter CONFIG_BRIDGE_VLAN_FILTERING, passing `vlanFiltering` in the Bridge struct returns an error creating the bridge interface. This happens even when no parameter is set on `Vlan` in the CNI config.

This change fixes the case where no `Vlan` parameter is configured on CNI config file so the flag doesn't need to be included in the struct.

Addresses problems related in issue #370 seen on the Raspberry Pi kernel.

Cc. @alexellis

### DIFF
--- a/plugins/main/bridge/bridge.go
+++ b/plugins/main/bridge/bridge.go
@@ -224,7 +224,9 @@ func ensureBridge(brName string, mtu int, promiscMode, vlanFiltering bool) (*net
 			// default packet limit
 			TxQLen: -1,
 		},
-		VlanFiltering: &vlanFiltering,
+	}
+	if vlanFiltering {
+		br.VlanFiltering = &vlanFiltering
 	}
 
 	err := netlink.LinkAdd(br)


alexellis:
The forced push was to update the commit message. Thank you for looking at this quickly for us.

bboreham:
> This change fixes the case where no Vlan parameter is configured on CNI config file

Nitpick: also if Vlan is configured as `0`.  Which is fine AFAIK, I just like to be precise.

alexellis:
Thanks @bboreham 

What needs to happen to get a patch release made up with binaries? We can keep maintaining a fork, but it would be nice to switch over to upstream again.

jellonek:
We plan to do release "quickly" but I can not guarantee if that will be in a week or two.

PR #449, opened 2020-02-04
flannel: remove net conf file after DEL succeed

mars1024:
`consumeScratchNetConf` will remove net conf file in defer, so if `invoke.DelegateDel` fails and CNI DEL is called again, it will always be successful because net conf file has been removed already.

So I think the better way is to remove net conf file after `invoke.DelegateDel` succeed, this will make sure the CNI DEL retry will go through `invoke.DelegateDel` every time.

### DIFF
--- a/plugins/meta/flannel/flannel.go
+++ b/plugins/meta/flannel/flannel.go
@@ -146,12 +146,19 @@ func saveScratchNetConf(containerID, dataDir string, netconf []byte) error {
 	return ioutil.WriteFile(path, netconf, 0600)
 }
 
-func consumeScratchNetConf(containerID, dataDir string) ([]byte, error) {
+func consumeScratchNetConf(containerID, dataDir string) (func(error), []byte, error) {
 	path := filepath.Join(dataDir, containerID)
-	// Ignore errors when removing - Per spec safe to continue during DEL
-	defer os.Remove(path)
 
-	return ioutil.ReadFile(path)
+	// cleanup will do clean job when no error happens in consuming/using process
+	cleanup := func(err error) {
+		if err == nil {
+			// Ignore errors when removing - Per spec safe to continue during DEL
+			_ = os.Remove(path)
+		}
+	}
+	netConfBytes, err := ioutil.ReadFile(path)
+
+	return cleanup, netConfBytes, err
 }
 
 func delegateAdd(cid, dataDir string, netconf map[string]interface{}) error {
--- a/plugins/meta/flannel/flannel_linux.go
+++ b/plugins/meta/flannel/flannel_linux.go
@@ -68,8 +68,8 @@ func doCmdAdd(args *skel.CmdArgs, n *NetConf, fenv *subnetEnv) error {
 	return delegateAdd(args.ContainerID, n.DataDir, n.Delegate)
 }
 
-func doCmdDel(args *skel.CmdArgs, n *NetConf) error {
-	netconfBytes, err := consumeScratchNetConf(args.ContainerID, n.DataDir)
+func doCmdDel(args *skel.CmdArgs, n *NetConf) (err error) {
+	cleanup, netConfBytes, err := consumeScratchNetConf(args.ContainerID, n.DataDir)
 	if err != nil {
 		if os.IsNotExist(err) {
 			// Per spec should ignore error if resources are missing / already removed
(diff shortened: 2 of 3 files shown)


plugins/meta/flannel/flannel_linux.go
```diff
@@ -78,10 +78,18 @@ func doCmdDel(args *skel.CmdArgs, n *NetConf) error {
...
+
 	nc := &types.NetConf{}
-	if err = json.Unmarshal(netconfBytes, nc); err != nil {
+	if err = json.Unmarshal(netConfBytes, nc); err != nil {
 		return fmt.Errorf("failed to parse netconf: %v", err)
 	}
 
-	return invoke.DelegateDel(context.TODO(), nc.Type, netconfBytes, nil)
+	err = invoke.DelegateDel(context.TODO(), nc.Type, netConfBytes, nil)
+	return err
```
jellonek: Why this is splitted?

plugins/meta/flannel/flannel_windows.go
```diff
@@ -65,10 +65,18 @@ func doCmdDel(args *skel.CmdArgs, n *NetConf) error {
...
+
 	nc := &types.NetConf{}
-	if err = json.Unmarshal(netconfBytes, nc); err != nil {
+	if err = json.Unmarshal(netConfBytes, nc); err != nil {
 		return fmt.Errorf("failed to parse netconf: %v", err)
 	}
 
-	return invoke.DelegateDel(context.TODO(), nc.Type, netconfBytes, nil)
+	err = invoke.DelegateDel(context.TODO(), nc.Type, netConfBytes, nil)
+	return err
```
jellonek: ditto

plugins/meta/flannel/flannel_linux.go
```diff
@@ -78,10 +78,18 @@ func doCmdDel(args *skel.CmdArgs, n *NetConf) error {
 		return err
 	}
 
+	defer func() {
+		// remove net conf after DEL succeed
+		if err == nil {
```
mars1024: To make the return of `invoke.DelegateDel` can be judged here, @jellonek

bboreham: You could avoid this by having `(err error)` as the return, i.e. a named return variable, and changing `err :=` to `err =`.  Both are pretty subtle points of Go behaviour.

plugins/meta/flannel/flannel.go
```diff
@@ -146,12 +146,17 @@ func saveScratchNetConf(containerID, dataDir string, netconf []byte) error {
...
 
-func consumeScratchNetConf(containerID, dataDir string) ([]byte, error) {
+func consumeScratchNetConf(containerID, dataDir string) (func(), []byte, error) {
 	path := filepath.Join(dataDir, containerID)
-	// Ignore errors when removing - Per spec safe to continue during DEL
-	defer os.Remove(path)
 
-	return ioutil.ReadFile(path)
+	// cleanup should be called after DEL succeed
+	cleanup := func() {
```
bboreham: if you set this to a no-op function on error then the caller can do `cleanup()` every time; doesn't have to know the coupling between error and cleanup.

mars1024: got it, thanks!

plugins/meta/flannel/flannel_linux.go
mars1024: thanks, will update this

mars1024:
Although the DEL retry mechanism of kubelet is disturbing something, but I still think the idempotency of each retry is important in CNI plugins. To avoid leaking or some wrong error messages.

plugins/meta/flannel/flannel_windows.go
```diff
@@ -65,10 +65,13 @@ func doCmdDel(args *skel.CmdArgs, n *NetConf) error {
 		return err
 	}
 
+	// cleanup will work when no error happens
+	defer cleanup(err)
```
bboreham: I think this doesn't work because it gets the value of `err` at the time control passes through this line.
See https://play.golang.org/p/3W5_yxRJJ_6 vs https://play.golang.org/p/-9c9Z3pvVFO

dcbw: The alternative is to make your cleanup function take a *pointer* to the error, and then `defer cleanup(&err)` and then it'll work :)

mars1024: updated, thanks!

mars1024:
updated, PTAL, thanks ~

PR #451, opened 2020-02-05
Add route-override plugin

s1061123:
This changes to add 'route-overide' meta CNI plugin, which
overrides (i.e. add/del) ip route for the interface.

### DIFF
--- a/plugins/meta/route-override/README.md
+++ b/plugins/meta/route-override/README.md
@@ -1,0 +1,64 @@
+# route-override: Meta CNI plugin for overriding IP route
+
+## Overview
+ route-override IPAM works as meta CNI plugin to override IP route given by previous CNI plugins.
+It is useful in multiple interface case with [network-attachment-definition](https://github.com/K8sNetworkPlumbingWG/multi-net-spec) to change the destination to specific interface, not default route.
+
+## Example Configuration
+
+```
+{
+    "cniVersion": "0.3.0",
+    "name" : "mymacvlan",
+    "plugins": [
+    {
+        "type": "macvlan",
+        "master": "eth1",
+        "mode": "bridge",
+        "ipam": {
+            ...
+        }
+    },
+    {
+        "type" : "route-override",
+        "del": [
+        {
+            "dst": "192.168.0.0/24"
+        }],
+        "add": [
+        {
+            "dst": "192.168.0.0/24",
+            "gw": "10.1.254.254"
+        }]
+    }
+    ]
+}
+```
+
+## Configuration Reference
+
+* `type`: (string, required): "routing-override"
+* `flushExisting`: (bool, optional): set it `true` if you flush all routes.
+* `flushDefaultGateway`: (bool, optional): set it `true` if you flush default route (gateway).
+* `del`: (object, optional): list of routes to delete from the container namespace. Each route is a dictionary with "dst" and optional "gw" fields. If "gw" is omitted, "gateway" in the previous plugin results will be used.
+* `add`: (object, optional): list of routes to add to the container namespace. Each route is a dictionary with "dst" and optional "gw" fields. If "gw" is omitted, "gateway" in the previous plugin results will be used.
+* `skipcheck`: (bool, optional): set it to `true` if there will be any changes in routes between `add`/`del` calls of this plugin.
+
+## Process Sequence
+
+`route-override` will manipulate the routes as following sequences:
+
+1. flush routes if `flushExisting` is enabled.
+1. flush gateway if `flushDefaultGateway` is enabled.
+1. if `del` is non empty - attempt to delete all listed in this variable routes. Missing routes are ignored.
+1. add all routes listed in `add`
+
+## Supported Arguments
+
+The following ["args" conventions](https://github.com/containernetworking/cni/blob/master/CONVENTIONS.md#args-in-network-config) are supported:
+All arguments is the same as configuration reference above.
+
+* `flushExisting`: (bool, optional)
+* `flushDefaultGateway`: (bool, optional)
+* `del`: (object, optional)
+* `add`: (object, optional)
--- a/plugins/meta/route-override/route-override.go
+++ b/plugins/meta/route-override/route-override.go
@@ -1,0 +1,418 @@
+// Copyright 2019 CNI authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// This is a "meta-plugin". It reads in its own netconf, it does not create
+// any network interface but just changes route information given from
+// previous cni plugins
+
+package main
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"net"
+
+	"github.com/containernetworking/cni/pkg/skel"
+	"github.com/containernetworking/cni/pkg/types"
+	"github.com/containernetworking/cni/pkg/types/current"
+	"github.com/containernetworking/cni/pkg/version"
+	"github.com/containernetworking/plugins/pkg/ns"
+
+	"github.com/vishvananda/netlink"
+)
+
+// RouteOverrideConfig represents the network route-override configuration
+type RouteOverrideConfig struct {
+	types.NetConf
+
+	FlushRoutes  bool           `json:"flushExisting,omitempty"`
+	FlushGateway bool           `json:"flushDefaultGateway,omitempty"`
+	DelRoutes    []*types.Route `json:"del"`
+	AddRoutes    []*types.Route `json:"add"`
+	SkipCheck    bool           `json:"skipcheck,omitempty"`
+
+	Args *struct {
+		A *IPAMArgs `json:"cni"`
+	} `json:"args"`
+}
+
+// IPAMArgs represents CNI argument conventions for the plugin
+type IPAMArgs struct {
+	FlushRoutes  *bool          `json:"flushExisting,omitempty"`
+	FlushGateway *bool          `json:"flushDefaultGateway,omitempty"`
+	DelRoutes    []*types.Route `json:"del,omitempty"`
+	AddRoutes    []*types.Route `json:"add,omitempty"`
+	SkipCheck    *bool          `json:"skipcheck,omitempty"`
+}
+
+func parseConf(data []byte, envArgs string) (*RouteOverrideConfig, error) {
+	conf := RouteOverrideConfig{FlushRoutes: false}
+
+	if err := json.Unmarshal(data, &conf); err != nil {
+		return nil, fmt.Errorf("failed to load netconf: %v", err)
+	}
+
+	// override values by args
+	if conf.Args != nil {
+		if conf.Args.A.FlushRoutes != nil {
+			conf.FlushRoutes = *conf.Args.A.FlushRoutes
+		}
+
+		if conf.Args.A.FlushGateway != nil {
+			conf.FlushGateway = *conf.Args.A.FlushGateway
+		}
+
+		if conf.Args.A.DelRoutes != nil {
+			conf.DelRoutes = conf.Args.A.DelRoutes
+		}
+
+		if conf.Args.A.AddRoutes != nil {
+			conf.AddRoutes = conf.Args.A.AddRoutes
+		}
+
+		if conf.Args.A.SkipCheck != nil {
+			conf.SkipCheck = *conf.Args.A.SkipCheck
+		}
+
+	}
+
+	// Parse previous result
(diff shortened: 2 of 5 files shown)


bboreham:
How come `vendor` has lots of changes but `go.mod` didn't change?

plugins/meta/route-override/README.md
```diff
@@ -0,0 +1,64 @@
+# route-override: Meta CNI plugin for overriding IP route
+
+## Overview
+ route-override IPAM works as meta CNI plugin to override IP route given by previous CNI plugins.
+It is useful in a case with [network-attachment-definition](https://github.com/K8sNetworkPlumbingWG/multi-net-spec).
```
matthewdupre: What case?

plugins/meta/route-override/README.md
```diff
@@ -0,0 +1,64 @@
...
+    ]
+}
+```
+
+## Configuration Reference
+
+* `type`: (string, required): "routing-override"
+* `flushroutes`: (bool, optional): true if you flush all routes.
+* `flushgateway`: (bool, optional): true if you flush default route (gateway).
+* `delroutes`: (object, optional): list of routes add to the container namespace. Each route is a dictionary with "dst" and optional "gw" fields. If "gw" is omitted, value of "gateway" will be used.
```
matthewdupre: s/add/delete/.  Routes to add (also for addroutes) would be better as well.

mccv1r0:
Does cniCheck need to be supported?

s1061123:
@bboreham thank you for your comments. It looks only the version/package list sorting as far as I looked. Of course I could remove it from diff because I don't introduce any pkgs.

s1061123:
@mccv1r0 Thanks! Now, CHECK command is already supported. But it is optional (see `skipcheck` option).

plugins/meta/route-override/README.md
s1061123: When a container has multiple interface, user (i.e. container process) want to send the traffic to specific destination (i.e. VPN traffic goes to `tun0`, not `eth0`). In such case, we need to modify routing table of the network namespace and the plugin does. Clarified?

plugins/meta/route-override/README.md
s1061123: Fixed. Thank you for your comments!

plugins/meta/route-override/README.md
s1061123: I added more description in README

plugins/meta/route-override/route-override.go
```diff
@@ -0,0 +1,411 @@
...
+
+// Todo:
+// + only checko route/dst
+//go build ./cmd/route-override/
+
+// RouteOverrideConfig represents the network route-override configuration
+type RouteOverrideConfig struct {
+	types.NetConf
+
+	PrevResult *current.Result `json:"-"`
```
mars1024: `PrevResult` is already in `types.NetConf`, why do we need a new one?

s1061123: Fixed.

plugins/meta/route-override/route-override.go
```diff
@@ -0,0 +1,400 @@
...
+				ips.Gateway = net.IPv6zero
+			} else {
+				ips.Gateway = net.IPv4zero
+			}
+		}
+	}
+
+	newRoutes := []*types.Route{}
+	err = netns.Do(func(_ ns.NetNS) error {
+		// Flush route if required
```
jellonek: s/route/routes/

plugins/meta/route-override/route-override.go
```diff
@@ -0,0 +1,400 @@
...
+							fmt.Fprintf(os.Stderr, "failed to delte route %v: %v", delroute, err)
+						}
+						continue NEXT
+					}
+
+				}
+				newRoutes = append(newRoutes, route)
+			}
+		} else {
+			deleteAllRoutes(res)
```
jellonek: Do not ignore errors returned there - they should be at least logged as a warning.

plugins/meta/route-override/route-override.go
```diff
@@ -0,0 +1,400 @@
...
+		}
+
+		if conf.FlushGateway {
+			deleteGWRoute(res)
```
jellonek: ditto

plugins/meta/route-override/route-override.go
```diff
@@ -0,0 +1,400 @@
...
+			}
+		}
+		// Add route
+		dev, _ := netlink.LinkByName(containerIFName)
+		for _, route := range conf.AddRoutes {
+			newRoutes = append(newRoutes, route)
+			if err := addRoute(dev, route); err != nil {
+				fmt.Fprintf(os.Stderr, "failed to add route: %v: %v", route, err)
+			}
+		}
```
jellonek: Above procedure is incorrect.
You should validate on which already configured interface is configured a subnet, in which the particular gw can be found. Using first found interface will lead to issues.
This also has to be covered with an integration test.

plugins/meta/route-override/route-override.go
```diff
@@ -0,0 +1,400 @@
...
+		return fmt.Errorf("failed to override routes: %v", err)
+	}
+
+	return types.PrintResult(newResult, overrideConf.CNIVersion)
+}
+
+func cmdDel(args *skel.CmdArgs) error {
+	// TODO: the settings are not reverted to the previous values. Reverting the
+	// settings is not useful when the whole container goes away but it could be
+	// useful in scenarios where plugins are added and removed at runtime.
```
jellonek: This can be a really big issue as we can not expect that next plugin in the chain will just ignore that configuration is different from what it left after it was called for `ADD` command.
There can be (i'm not sure but probably calico is an example) plugins "in the wild" which will find that undesirable.

plugins/meta/route-override/route-override.go
```diff
@@ -0,0 +1,400 @@
...
+	}
+
+	return err
+}
+
+func deleteRoute(route *types.Route, res *current.Result) error {
+	var err error
+	// fallback to eth0 if there is no interface in result
+	if res.Interfaces == nil {
+		link, _ := netlink.LinkByName("eth0")
```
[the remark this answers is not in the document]
s1061123: This code does not delete default route and it just deletes explicit route in `route`, hence not changed. Please let me know if you have clear reason why we remove default route when other route is removed.

plugins/meta/route-override/route-override.go
s1061123: I'm not clear about it. 

As far as I checked, [calico CNI plugin](https://github.com/projectcalico/cni-plugin) does not touch routes in cmdDel() as following:

https://github.com/projectcalico/cni-plugin/blob/master/internal/pkg/utils/network_linux.go#L437-L478

s1061123:
>IMO the biggest no-go for this in current version is lack of DEL command support.
>This will be tricky, as plugin has to store info about what it's removing, then it should readd everything in a proper order.

I am not clear that why we need to revert the routes.

1. No interface plugins does not remove routes at DEL command
  I cannot find all interface CNI plugins in github.com/containernetworking/plugins, that manipulate the routes in DEL command. This means the interface related routes are removed by linux kernel at the interface deletion (i.e. netlink.LinkDel). So usually none does not verify the routes at DEL.

2. Routes are changed by others hence we cannot identify which routes should be valid
  Linux network does not guarantee that routes is not changed once setup, even though none changed the routes.
  In IPv6 case, Router Advertisement may inject some routes for the interface after ADD command.

3. If we found inconsistent routes at DEL command, nothing can be done
  CNI spec mentions `Plugins should generally complete a DEL action without error even if some resources are missing.`
  So even though we found some inconsistent routes at DEL command, we just returns success, so we don't have to revert
  it.

Could you let me know some valid use case that route needs to be restored to initial state at DEL?

plugins/meta/route-override/route-override.go
jellonek: It was only an example (as You pointed - missed one).
Also afair multus can work (or is planned to work) in "thick plugin" mode which will allow dynamic calling add/remove on plugins - in which case this will also lead to issues.

jellonek:
> 1. No interface plugins does not remove routes at DEL command

because there is no need for that if you are removing an interface (routes connected with it are automatically tear down)

ad. 2&3 - valid point with Router Advertisement. IMO we should at least try to recreate situation which was on the stack before ADD was called and if that will fail, return a success.

Anyway - after rethinking that back - I can live with keeping that as TODO, so we can just wait for any issue with that and if it will appear - get back to the story again.

plugins/meta/route-override/route-override.go
s1061123: Currently from multus side, none working for "thick plugin" because of less usecase/requirements, so I suppose your concern is valid but target to future. How about keep it in TODO (or issue) to track it?

plugins/meta/route-override/route-override.go
```diff
@@ -0,0 +1,420 @@
...
+		if err != nil {
+			return nil, fmt.Errorf("could not convert result to current version: %v", err)
+		}
+	}
+
+	return &conf, nil
+}
+
+func deleteAllRoutes(res *current.Result) error {
+	var err error
```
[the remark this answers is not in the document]
s1061123: I cannot remove it because 

- golang requires type declaration (as you may know about it, of course) (1)
- using short variable declarations, :=, we may remove it in some case, but this case is not because of scope of variables. (2)

(1)
```
[tohayash@tohayash-srv plugins-s1061123]$ ./build_linux.sh
Building plugins
  bandwidth
  firewall
  flannel
  portmap
  route-override
# github.com/containernetworking/plugins/plugins/meta/route-override
plugins/meta/route-override/route-override.go:124:8: undefined: err
plugins/meta/route-override/route-override.go:125:11: undefined: err
plugins/meta/route-override/route-override.go:126:16: undefined: err
plugins/meta/route-override/route-override.go:130:7: undefined: err
plugins/meta/route-override/route-override.go:131:10: undefined: err
plugins/meta/route-override/route-override.go:132:15: undefined: err
plugins/meta/route-override/route-override.go:140:9: undefined: err
```

(2)
```
# github.com/containernetworking/plugins/plugins/meta/route-override
plugins/meta/route-override/route-override.go:140:9: undefined: err
```

plugins/meta/route-override/route-override.go
s1061123: Let me double check that we should do check the gateway is already configured or not in the plugin?
As far as I know, the linux kernel network stack will take care of the validation, hence we do not do that in the plugin side. I also output error message from `netlink.RouteAdd()`.

s1061123:
@jellonek I've taken care of your comments, could you please take a look into it to double check?

plugins/meta/route-override/route-override.go
```diff
@@ -0,0 +1,419 @@
...
+	newResult, err := processRoutes(args.Netns, overrideConf)
+	if err != nil {
+		return fmt.Errorf("failed to override routes: %v", err)
+	}
+
+	return types.PrintResult(newResult, overrideConf.CNIVersion)
+}
+
+func cmdDel(args *skel.CmdArgs) error {
+	// TODO: the settings are not reverted to the previous values. Reverting the
```
dcbw: @s1061123 I would say just remove the TODO since I don't think it's useful or feasible to clear out the settings.

plugins/meta/route-override/route-override.go
```diff
@@ -0,0 +1,419 @@
...
+		// Flush routes if required
+		if !conf.FlushRoutes {
+		NEXT:
+			for _, route := range res.Routes {
+				for _, delroute := range conf.DelRoutes {
+					if route.Dst.IP.Equal(delroute.Dst.IP) &&
+						bytes.Equal(route.Dst.Mask, delroute.Dst.Mask) {
+						err = deleteRoute(delroute, res)
+						if err != nil {
+							fmt.Fprintf(os.Stderr, "failed to delte route %v: %v", delroute, err)
```
dcbw: @s1061123 "delte" -> "delete"

plugins/meta/route-override/route-override.go
```diff
@@ -0,0 +1,419 @@
...
+				containerIFName = i.Name
+				break
+			}
+		}
+		// Add route
+		dev, _ := netlink.LinkByName(containerIFName)
+		for _, route := range conf.AddRoutes {
+			newRoutes = append(newRoutes, route)
+			if err := addRoute(dev, route); err != nil {
+				fmt.Fprintf(os.Stderr, "failed to add route: %v: %v", route, err)
```
dcbw: @s1061123 should this be a hard failure? If the user asked to add a route, and it can't be added, it seems like it should fail. Thoughts?

plugins/meta/route-override/route-override.go
s1061123: changed.

plugins/meta/route-override/route-override.go
s1061123: fixed.

plugins/meta/route-override/route-override.go
s1061123: hmm... At first, sometimes adding routes are failed because some plugins may add conflicting routes implicitly hence I accept the failure to add route. But it should be hard failure, so changed them (add/del/delAll).

Elegant996:
Would it be possible to incorporate support for routing tables as an extra field? Default would be table 0 if omitted. Thanks!

Issue #320, opened 2019-05-17
host-device plugin doesn't work with virtio interfaces


pperiyasamy:
The existing host-device CNI plugin isn't made to work with virtio interfaces inside a VM (Kubernetes worker node).  The following error is seen while creating k8s POD.

`  Warning  FailedCreatePodSandBox  37s               kubelet, my-centos  Failed create pod sandbox: rpc error: code = Unknown desc = [failed to set up sandbox container "a9635a3fad602b0d6172a52010b1e01695e99f1372d1054413edecfb73c5b82e" network for pod "testpod1": NetworkPlugin cni failed to set up pod "testpod1_default" network: Multus: Err adding pod to network "host-device-network": Multus: error in invoke Delegate add - "host-device": failed to find host device: no net directory under pci device 0000:00:14.0: "lstat /sys/bus/pci/devices/0000:00:14.0/net: no such file or directory", failed to clean up sandbox container "a9635a3fad602b0d6172a52010b1e01695e99f1372d1054413edecfb73c5b82e" network for pod "testpod1": NetworkPlugin cni failed to teardown pod "testpod1_default" network: Multus: error in invoke Delegate del - "host-device": failed to find "net2": Link not found / Multus: error in invoke Delegate del - "host-device": failed to find "net1": Link not found]`

It is looking for net directory at /sys/bus/pci/devices/0000:00:14.0, but for virtio device (0000:00:14.0) resides under /sys/bus/pci/devices/0000:00:14.0/virtio<<vf_id>> directory.

pperiyasamy:
Fixed the issue with #453

VamshiKrishnaNemalikonda:
I see similar issue - can someone suggest. I'm using Multus + host-device. Thank you

[...]

PR #453, opened 2020-02-11
Make host-device to work with virtio net device

pperiyasamy:
In case pciBusID contains pci address of the virtio device, then lookup `ip link` under `{sysBusPCIDir}/virtio{id}/net` directory.

Issue: https://github.com/containernetworking/plugins/issues/320

Signed-off-by: Periyasamy Palanisamy <periyasamy.palanisamy@est.tech>

### DIFF
--- a/plugins/main/host-device/host-device.go
+++ b/plugins/main/host-device/host-device.go
@@ -296,7 +296,12 @@ func getLink(devname, hwaddr, kernelpath, pciaddr string) (netlink.Link, error) 
 	} else if len(pciaddr) > 0 {
 		netDir := filepath.Join(sysBusPCI, pciaddr, "net")
 		if _, err := os.Lstat(netDir); err != nil {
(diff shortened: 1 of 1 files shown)


dcbw:
lgtm I guess...

pperiyasamy:
> Looks solid to me. Can we test that somehow?

@jellonek and @dcbw thanks for the feedback.
@jellonek  how do you want to test it ? I've tested the fix via [device plugin](https://github.com/intel/sriov-network-device-plugin) and [multus](https://github.com/intel/multus-cni) which were configured to pass virtio interface PCI address as `pciBusID` into `host-device` plugin.

jellonek:
I think that the only way to test that would be to turn sysBusPCI from const to a variable (with possibility to change it's value from test package, so by turning it into exported variable or by adding a helper fuc to set its value), which then could be mocked in test passing a manually prepared directory with expected entries.
That would be quite extensive change so IMO we can postpone it as a TODO item.

PR #455, opened 2020-02-20
Unlock OS thread after netns is restored

booxter:
The current ns package code is very careful about not leaving the calling
thread with the overridden namespace set, for example when origns.Set() fails.
This is achieved by starting a new green thread, locking its OS thread, and
never unlocking it. Which makes golang runtime to scrap the OS thread backing
the green thread after the go routine exits.

While this works, it's probably not as optimal: stopping and starting a new OS
thread is expensive and may be avoided if we unlock the thread after resetting
network namespace to the original. On the other hand, if resetting fails, it's
better to leave the thread locked and die.

While it won't work in all cases, we can still make an attempt to reuse the OS
thread when resetting the namespace succeeds. This can be achieved by unlocking
the thread conditionally to the namespace reset success.

Signed-off-by: Ihar Hrachyshka <ihrachys@redhat.com>

### DIFF
--- a/pkg/ns/ns_linux.go
+++ b/pkg/ns/ns_linux.go
@@ -178,7 +178,16 @@ func (ns *netNS) Do(toRun func(NetNS) error) error {
 		if err = ns.Set(); err != nil {
 			return fmt.Errorf("error switching to ns %v: %v", ns.file.Name(), err)
 		}
-		defer threadNS.Set() // switch back
+		defer func() {
+			err := threadNS.Set() // switch back
+			if err == nil {
+				// Unlock the current thread only when we successfully switched back
+				// to the original namespace; otherwise leave the thread locked which
+				// will force the runtime to scrap the current thread, that is maybe
+				// not as optimal but at least always safe to do.
+				runtime.UnlockOSThread()
+			}
+		}()
 
 		return toRun(hostNS)
 	}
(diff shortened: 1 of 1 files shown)


booxter:
@rmohr @davidvossel

jellonek:
We don't have this noted anywhere specifically (and probably this should be changed) that we are trying to be compliant only with last 3 stable versions of golang, so right now with go >=1.11.

Our tests are running only against https://github.com/containernetworking/plugins/blob/master/.travis.yml#L5 these versions.

rmohr:
> We don't have this noted anywhere specifically (and probably this should be changed) that we are trying to be compliant only with last 3 stable versions of golang, so right now with go >=1.11.

Then the PR should be good. Thanks.

mars1024:
@jellonek @rmohr A small question, why should we force-unlock the OS thread? I think the lock will be released after the goroutine exits.

rmohr:
> @jellonek @rmohr A small question, why should we force-unlock the OS thread? I think the lock will be released after the goroutine exits.

Instead of unlocking the thread, the thread gets reaped when the routine exits when it is not unlocked: https://golang.org/src/runtime/proc.go?s=102346:102367#L3574.

We will get new threads if needed, but it makes the operation definitely heavier.

jellonek:
@mars1024 To understand that you need to know about long standing bug in golang which was leading to reusing threads with changed netns by other goroutines, while they were expecting that they are running in the original one netns.
PTAL on https://www.weave.works/blog/linux-namespaces-golang-followup

jellonek:
Let's wait for next maintainers meeting (02.26) with merge.

bboreham:
Historical note: here is the statement that "`UnlockOSThread` is not safe for use in any library code"
https://github.com/golang/go/issues/20458.
This was fixed in Go 1.10 as noted.

PR #456, opened 2020-02-22
modify the error url of windowscontainer

timyinshi:
the error url is: 
https://docs.microsoft.com/en-us/virtualization/windowscontainers/manage-containers/container-networking

the right url is:
https://docs.microsoft.com/en-us/virtualization/windowscontainers/container-networking/architecture


dcbw:
@timyinshi can you add your DCO to the commit? `git commit --amend -s` and then force-push the commit to your branch should be good enough. Thanks!

timyinshi:
@dcbw thanks,i will have a new pr

Issue #457, opened 2020-02-23
replace juju/errors because of CNCF license scan report


mars1024:
Just as the newest [license](https://lfscanning.org/reports/cncf/cni-2020-02-05-8f096c05-3cfa-424c-8431-8c04719143c3.html) scan report from CNCF, `juju/errors` will not be suitable to be used in our repos.
We should find another way to replace it.

mars1024:
Err in juju/errors is just like below
```
func (e *Err) Error() string {
	// We want to walk up the stack of errors showing the annotations
	// as long as the cause is the same.
	err := e.previous
	if !sameError(Cause(err), e.cause) && e.cause != nil {
		err = e.cause
	}
	switch {
	case err == nil:
		return e.message
	case e.message == "":
		return err.Error()
	}
	return fmt.Sprintf("%s: %v", e.message, err)
}
```

So I will replace it with `fmt.Errorf` ..

PR #458, opened 2020-02-26
replace juju/errors because of CNCF license scan

mars1024:
ref to #457


bboreham:
The basic idea is OK, but now [Go stdlib](https://golang.org/pkg/errors/) has `%w` I think we should use that.
Since we don't want to force everyone to upgrade to Go 1.13, this could be a wrapper function inside a utility package here.  It could even be named `errors.Annotate` to reduce diffs.

mars1024:
> The basic idea is OK, but now [Go stdlib](https://golang.org/pkg/errors/) has `%w` I think we should use that.
> Since we don't want to force everyone to upgrade to Go 1.13, this could be a wrapper function inside a utility package here. It could even be named `errors.Annotate` to reduce diffs.

Thanks, added an error utility package as wrapper to reduce diffs.

squeed:
lgtm!

Issue #461, opened 2020-03-10
ERROR: table `nat' is incompatible, use 'nft' tool.


greenpau:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind feature

**Description**

I use nftables; when starting a container I get: 

```
ERRO[0000] Error adding network: failed to list chains: running [/usr/sbin/iptables -t nat -S --wait]: exit status 1: iptables v1.8.2 (nf_tables): table `nat' is incompatible, use 'nft' tool.

ERRO[0000] Error while adding pod to CNI network "podman": failed to list chains: running [/usr/sbin/iptables -t nat -S --wait]: exit status 1: iptables v1.8.2 (nf_tables): table `nat' is incompatible, use 'nft' tool.

Error: error configuring network namespace for container 51f6adbaed7d674fb4b48d501eb7ce0605d09e003ac09f6588b98dea7230ca9f: failed to list chains: running [/usr/sbin/iptables -t nat -S --wait]: exit status 1: iptables v1.8.2 (nf_tables): table `nat' is incompatible, use 'nft' tool.
```

**Steps to reproduce the issue:**

1. Create network configuration:

```
cat >/etc/cni/net.d/podman.conflist <<EOF
{
  "cniVersion": "0.4.0",
  "name": "podman",
  "plugins": [
    {
      "type": "bridge",
      "bridge": "cni-podman0",
      "isGateway": true,
      "ipMasq": true,
      "ipam": {
        "type": "host-local",
        "routes": [
          {
            "dst": "0.0.0.0/0"
          }
        ],
        "ranges": [
          [
            {
              "subnet": "192.168.124.0/24",
              "gateway": "192.168.124.1"
            }
          ]
        ]
      }
    },
    {
      "type": "portmap",
      "capabilities": {
        "portMappings": true
      }
    },
    {
      "type": "firewall",
      "backend": "nftables"
    }
  ]
}
EOF
```

2. Pull fedora image and start a container:

```
podman pull fedora:latest
podman run -it fedora bash
```

3.

**Describe the results you received:**

```
ERRO[0000] Error adding network: failed to list chains: running [/usr/sbin/iptables -t nat -S --wait]: exit status 1: iptables v1.8.2 (nf_tables): table `nat' is incompatible, use 'nft' tool.

[...]

PR #462, opened 2020-03-15
WIP: firewall: add nftables backend

greenpau:
Resolves: #461


greenpau:
@mars1024 , could you please take a look at this PR? 🙏  It is almost ready (pending adding tests), but I did update `README.md`.

Also, currently, the `nftables` backend appears to be working with NAT table too. However, I noticed orphan chains there https://github.com/containernetworking/plugins/issues/461#issuecomment-599290204

mars1024:
> @mars1024 , could you please take a look at this PR? 🙏 It is almost ready (pending adding tests), but I did update `README.md`.

Of course, and you should add some tests later.

mars1024:
BTW, cc firewall plugin author @dcbw

greenpau:
@mars1024 , qq, this is first time I am writing tests with `ginko` framework.

How I can test the code locally? I am trying invoking it in the following fashion. I do the same for existing test and it does not work either. This leads me to believe there is some other way to invoke testing.

```
$ go test -v plugins/meta/firewall/firewall_nftables_test.go
# command-line-arguments [command-line-arguments.test]
plugins/meta/firewall/firewall_nftables_test.go:168:12: undefined: cmdAdd
plugins/meta/firewall/firewall_nftables_test.go:176:12: undefined: cmdCheck
FAIL    command-line-arguments [build failed]
FAIL
```

greenpau:
@mars1024 , I tried ` ./test_linux.sh`, but it tests all plugins, not just `firewall`.

mars1024:
`ginko` can also work with `go test`, please try `go test github.com/containernetworking/plugins/plugins/meta/firewall`

greenpau:
> ginko can also work with go test, please try go test github.com/containernetworking/plugins/plugins/meta/firewall

@mars1024 , thank you! It worked!

plugins/meta/firewall/README.md
```diff
@@ -133,3 +133,99 @@ of the container as shown:
...
+
+```
+$ nft list chain ip filter FORWARD -a
+table ip filter {
+        chain FORWARD { # handle 2
+                type filter hook forward priority 0; policy drop;
+                oifname "cni-podman0" ip daddr 192.168.124.0/24 ct state established,related counter packets 100 bytes 113413 accept # handle 71
+                iifname "cni-podman0" ip saddr 192.168.124.0/24 counter packets 124 bytes 12996 accept # handle 72
+                iifname "cni-podman0" oifname "cni-podman0" counter packets 0 bytes 0 accept # handle 73
+                oifname "virbr0" ip daddr 192.168.122.0/24 ct state established,related counter packets 0 bytes 0 accept # handle 51
```
mars1024: How about move these rules into a new chain of filter table? just like `iptables` backend

plugins/meta/firewall/nftables.go
```diff
@@ -0,0 +1,350 @@
...
+		r.verdict = "accept"
+	}
+	if strings.HasSuffix(r.text, " drop") {
+		r.verdict = "drop"
+	}
+	return r, nil
+}
+
+func (r *nftRule) parseText() error {
+	offset := strings.LastIndex(r.text, "# handle ")
```
mars1024: `strings.Split` seems more efficient here

plugins/meta/firewall/nftables.go
```diff
@@ -0,0 +1,350 @@
...
+	if offset > 0 {
```
mars1024: should this return error if offset < 0?

plugins/meta/firewall/nftables.go
```diff
@@ -0,0 +1,350 @@
...
+func (nb *nftBackend) isValidInput(result *current.Result) error {
+	if len(result.Interfaces) == 0 {
+		return fmt.Errorf("the data passed to firewall plugin did not contain network interfaces")
+	}
+
+	if result.Interfaces[0].Name == "" {
+		return fmt.Errorf("the data passed to firewall plugin has no bridge name, e.g. cnibr0")
+	}
+
+	nb.targetInterface = result.Interfaces[0].Name
```
mars1024: I don't think target interface is always at position 0, what we want here is the peer of container interface.

plugins/meta/firewall/nftables.go
```diff
@@ -0,0 +1,350 @@
...
+
+	if len(result.IPs) == 0 {
+		return fmt.Errorf("the data passed to firewall plugin has no IP addresses")
+	}
+
+	if len(result.IPs) != 1 {
+		return fmt.Errorf("the data passed to firewall plugin has more than one IP address")
+	}
+
+	addr := result.IPs[0].Address
```
mars1024: Actually, we support multiple-IP allocation in a single cni process

plugins/meta/firewall/nftables.go
```diff
@@ -0,0 +1,350 @@
...
+	if addr.IP.To4() == nil {
+		return fmt.Errorf("the data passed to firewall plugin has non-IPv4 address")
+	}
+
+	_, netAddr, err := net.ParseCIDR(addr.String())
+	if err != nil {
+		return fmt.Errorf("the data passed to firewall plugin has invalid IPv4 address")
+	}
+
+	nb.targetNetwork = netAddr
```
mars1024: targetNetwork here should be an IP, not a CIDR

mars1024:
We need more tests @greenpau

greenpau:
> We need more tests @greenpau

@mars1024 , working on ti. Thank you for the review!

plugins/meta/firewall/README.md
greenpau: > How about move these rules into a new chain of filter table? just like iptables backend

@mars1024 , the challenge with doing that is you end up having a lot of chain that you would need to pull. It is like pulling the entire config. In case of `iptables` backend, there is a good library behind it, i.e. `go-iptables`. In the case of `nftables`, there is `google/nftables`, but it is not as extensible as it comes to rule. see https://github.com/sbezverk/nftableslib/issues/56#issuecomment-599162875

Here, we list a single chain, find the rules with relevant interfaces, and add/remove what is no longer necessary.

plugins/meta/firewall/nftables.go
greenpau: @mars1024 , will change it to use `Split`.

plugins/meta/firewall/nftables.go
greenpau: > should this return error if offset < 0?

@mars1024 , not an issue as we pivot to `Split()`. I guess my thought was that the relevant rule will not start with `#`.

plugins/meta/firewall/nftables.go
greenpau: > Actually, we support multiple-IP allocation in a single cni process

@mars1024 , will amend to support multiple IPv4 interfaces.

plugins/meta/firewall/nftables.go
greenpau: > targetNetwork here should be an IP, not a CIDR

@mars1024 , in this case all you are doing is NAT overloading, why do you need to create rules for each individual IP? It is less intensive to do a calculation for 1 CIDR as opposed to 100s IP addresses.

greenpau:
> How about move these rules into a new chain of filter table? just like iptables backend

@mars1024, after thinking about it a bit, it may require its own chain .. for consistency.

What would the logic behind naming the individual chains? For example, I see IP masquarade chaing names in the following fashion. I guess, what is the logic behind `586108da8523562bc2ca795b `?

```
table ip nat { # handle 27

        chain POSTROUTING { # handle 3
                ip saddr 192.168.124.145  counter packets 0 bytes 0 jump CNI-586108da8523562bc2ca795b # handle 562

        chain CNI-586108da8523562bc2ca795b { # handle 559
                ip daddr 192.168.124.0/24  counter packets 0 bytes 0 accept # handle 560
                ip daddr != 224.0.0.0/4  counter packets 0 bytes 0 masquerade  # handle 561
        }
```

greenpau:
> @mars1024, after thinking about it a bit, it may require its own chain .. for consistency.

@mars1024 , running into an issue implementing it as a chain. how many chain would you create for the below rules?

```
nft insert rule filter FORWARD position 51 oifname "cni-podman0" ip daddr 192.168.124.0/24 ct state established,related counter packets 0 bytes 0 accept
nft insert rule filter FORWARD position 51 iifname "cni-podman0" ip saddr 192.168.124.0/24 counter packets 0 bytes 0 accept
```

Do you create a chain for interface? e.g. `CNI-FORWARD-PODMAN0`? them sub-chain for each IP address?

I am still think it could be proper to do it with a subnet match.

greenpau:
Back on this PR.

greenpau:
Pivoting from invoking `nft` via shell to using [`github.com/google/nftables`](https://github.com/google/nftables).

Found decent blog posts on the subject:

* https://tomdnetto.net/post/advanced_nftables_with_go
* https://tomdnetto.net/post/linux_networking_from_go_nftables

plugins/meta/firewall/README.md
greenpau: > How about move these rules into a new chain of filter table? just like iptables backend

@mars1024 , creating separate chains with a jump in the FORWARD chain.

greenpau:
Currently waiting on the resolution of the issue related to Chain information being missing from verdicts. https://github.com/google/nftables/pull/106

greenpau:
> Currently waiting on the resolution of the issue related to Chain information being missing from verdicts. google/nftables#106

This is resolved and merged.

greenpau:
Pending clarity as to `vendor/` directory question https://github.com/containernetworking/plugins/issues/518

greenpau:
@mars1024 , @dcbw , please review this PR. I think I addressed some of the initial concerns. The build is failing b/c of #518, but that does not change the overall functionality.

dcbw:
@erig0 can you review this one too? thanks!

greenpau:
> @erig0 can you review this one too? thanks!

@mars1024 , @dcbw , @erig0, it might actually not work due to the fact the `containernetworking/plugins` using older version of Go. Please see this comment: https://github.com/google/nftables/issues/110#issuecomment-668386381. Would love your input as to how to resolve this.

dcbw:
@greenpau we can certainly bump the required Go version of the CNI plugins repo. Our general policy is Go versions N and N-1, which means we should support Go 1.13 and Go 1.14. If it's possible to support 1.12 that's great, but not required. Anything 1.11 and lower is not supported.

greenpau:
> we can certainly bump the required Go version of the CNI plugins repo. 

@dcbw , that would be great!

erig0:
> Pivoting from invoking nft via shell to using github.com/google/nftables.

Another option is to use libnftables. This what `nft` uses. firewalld also uses the JSON support from libnftables.

plugins/meta/firewall/README.md
```diff
@@ -133,3 +133,207 @@ of the container as shown:
...
+        chain FORWARD {
+                type filter hook forward priority filter; policy drop;
+                jump cnins-3-4026550857-dummy0
+        }
+
+        chain cnins-3-4026550857-dummy0 {
+                oifname "dummy0" ip6 daddr 2001:db8:100:100::1 ct state established,related counter packets 0 bytes 0 accept
+                iifname "dummy0" ip6 saddr 2001:db8:100:100::1 counter packets 0 bytes 0 accept
+                iifname "dummy0" oifname "dummy0" counter packets 0 bytes 0 accept
+        }
```
erig0, line 337: Ideally these would be combined into one chain in a table of family `inet`. I don't think there is value in splitting them up. It can probably avoid some duplicate rules.

greenpau: > Ideally these would be combined into one chain in a table of family inet. I don't think there is value in splitting them up. It can probably avoid some duplicate rules.

@erig0, please see @mars1024  comment that he wanted a jump.

greenpau: > I don't think there is value in splitting them up. It can probably avoid some duplicate rules.

@erig0, it has a value in a sense that it is easier to delete the rule when you have a jump like this. It would be very hard to parse each rule.

plugins/meta/firewall/README.md
```diff
@@ -133,3 +133,207 @@ of the container as shown:
...
+      {
+         "type": "firewall",
+         "backend": "nftables"
+      }
+   ]
+}
+```
+
+Prior to the invocation of CNI `firewall` plugin, the `FORWARD` chain in `filter`
+table might be configured be as follows:
```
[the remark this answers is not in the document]
greenpau: > This is not guaranteed to exist. It will likely exist on any machine that uses the iptables-nft variant of iptables as it implicitly create this table/chain. I would not assume the table/chain exists.

@erig0, I can create a chain. that is I do it when testing changes.

@dcbw , @mars1024 , agreed?

greenpau:
From failed tests:

```
  Unexpected error:
      <*errors.errorString | 0xc000085fa0>: {
          s: "nftBackend.Add() error: failed parsing netfilter tables: nftables table filter not found",
      }
      nftBackend.Add() error: failed parsing netfilter tables: nftables table filter not found
  occurred
```

Now most of the tests pass. Noticed the issue with `table filter not found`. Per comment by @erig0 , need to add the chain creation.

plugins/meta/firewall/README.md
erig0: I get the jump to a dedicated chain. That 100% makes sense. I was saying I don't think you need to split the "ip" and "ip6" families. You could use "inet" which covers both IPv4 and IPv6.

greenpau: > I get the jump to a dedicated chain. That 100% makes sense. I was saying I don't think you need to split the "ip" and "ip6" families. You could use "inet" which covers both IPv4 and IPv6.

@erig0, that I could do. 👍 It makes sense because IPv4/6 are all a part of the same construct.

greenpau:
just noticed an issue with `chanName`. I forgot to add `ContainerID` in the mix.

greenpau:
another issue is with the interface name ... I take the interface name of the container. It should be the interface name of the bridge ... the very first interface of in the list.

greenpau:
Added working version of nftables plugin in https://github.com/greenpau/cni-plugins#cni-nftables-firewall. Still testing it.

PavelSosin-320:
Why firewall configuration is needed when Podman runs inside a VM container,  like WSL2 ? A firewall is always supplied by the Host node. The communication between the Podman network and interface Eth0 provided dynamically by the Virtual switch doesn't need any filtering. In the overlay network, the entire local network can be protected by the router if it runs OpenWRT or Free Balena.

06kellyjac:
> Why firewall configuration is needed when Podman runs inside a VM container, like WSL2 ? A firewall is always supplied by the Host node. The communication between the Podman network and interface Eth0 provided dynamically by the Virtual switch doesn't need any filtering. In the overlay network, the entire local network can be protected by the router if it runs OpenWRT or Free Balena.
@PavelSosin-320 

In this context the firewall plugin is how networking is set up for the containers. This is separate to what you refer to as the "overlay network".
You can use a "fully featured" firewall like `firewalld` or, as most people use containers, it'll actually be using `iptables` (a packet filter).
This PR is to add `nftables` as an option (a different packet filter).

You can read up on the CNI firewall plugin here: https://github.com/containernetworking/plugins/blob/master/plugins/meta/firewall/README.md


You can already use podman with iptables in WSL2 without this PR, as long as your linux distro still uses iptables.

---

If I got anything wrong here let me know

dcbw:
> another issue is with the interface name ... I take the interface name of the container. It should be the interface name of the bridge ... the very first interface of in the list.

Note that this won't always be the first interface... we've struggled with this for other plugins (eg bandwidth) too, and don't have a great answer yet because some plugins might not have a host-side network interface, like VMs that use tap. But in any case, take a look at what the bandwidth plugin does; I *think* it takes the first non-sandbox interface in the list. If you just look for the first interface, it may be the container interface, so you have to look at the "Sandbox" key.

greenpau:
> Note that this won't always be the first interface... 

@dcbw , I reworked the plugin completely ... please see https://github.com/greenpau/cni-plugins/tree/main/pkg/firewall

I was thinking of linking from here to https://github.com/greenpau/cni-plugins/blob/6b7a838e15f159e8695df59b4f4f722ca3f49118/pkg/firewall/cmd.go#L12. Any objections?

Also, on a separate note, I found out that `portmap` will not work on RHEL 7, just RHEL 8.

cjschroeder:
Has there been any movement on this front?  Nftables support (not firewalld) is a necessary need for those using the nftables service on Fedora, for instance.

SilverBut:
Not sure what's going on there since its now closed. @greenpau Any idea?

greenpau:
> Not sure what's going on there since its now closed.

@SilverBut , closed it since I implemented it here: https://github.com/greenpau/cni-plugins

3nprob:
@dcbw @squeed @mars1024 @erig0  Considering that the PR author abandoned getting this merged here and are content with their own fork - what's missing for you (apart from rebasing) to get this PR to a mergable state?

nftables support for firewall is a pretty important feature, and getting moreso as more and more migrate to nft.

rtgiskard:
I'm using nftables with customized rules, and do not want any part of the system to alter the rules itself, so that all the firewall rules can be managed in one place and easy to check.

Is it possible to have an option to disable iptables just like [docker](https://docs.docker.com/network/iptables/#prevent-docker-from-manipulating-iptables)?

rtgiskard:
> @rtgiskard
> 
> > Is it possible to have an option to disable iptables just like [docker](https://docs.docker.com/network/iptables/#prevent-docker-from-manipulating-iptables)?
> 
> I suggest you create a new issue.

thank you, here is the [new issue](https://github.com/containernetworking/plugins/issues/762)

unknowndevQwQ:
> > Not sure what's going on there since its now closed.
> 
> @SilverBut , closed it since I implemented it here: https://github.com/greenpau/cni-plugins

@greenpau What's stopping you from merging this beautiful new feature into Upstream?

3nprob:
@unknowndevQwQ the same thing stopping you, presumably :p
It'd have to get merged by a maintainer.

PR #468, opened 2020-03-26
check bridge's port state

### LINKED ISSUE #463: GratuitousArp not work,  bridge can not receive arp package
when create pod,  GratuitousArp  can only receive by veth on host, bridge can not get arp package by tcpdump
https://github.com/containernetworking/plugins/blob/master/plugins/main/bridge/bridge.go#L472

if add sleep here, will work well
```
		// Send a gratuitous arp
		for _, ipc := range result.IPs {
			if ipc.Version == "4" {
				time.Sleep(1 * time.Second)
				_ = arping.GratuitousArpOverIface(ipc.Address.IP, *contVeth)
			}
		}
```

hongli-my:
fix #463
link host veth pair to bridge, the Initial state
of port is BR_STATE_DISABLED and change to
BR_STATE_FORWARDING async.

### DIFF
--- a/plugins/main/bridge/bridge.go
+++ b/plugins/main/bridge/bridge.go
@@ -22,6 +22,7 @@ import (
 	"net"
 	"runtime"
 	"syscall"
+	"time"
 
 	"github.com/j-keck/arping"
 	"github.com/vishvananda/netlink"
@@ -444,11 +445,6 @@ func cmdAdd(args *skel.CmdArgs) error {
 
 		// Configure the container hardware address and IP address(es)
 		if err := netns.Do(func(_ ns.NetNS) error {
-			contVeth, err := net.InterfaceByName(args.IfName)
-			if err != nil {
-				return err
-			}
-
 			// Disable IPv6 DAD just in case hairpin mode is enabled on the
 			// bridge. Hairpin mode causes echos of neighbor solicitation
 			// packets, which causes DAD failures.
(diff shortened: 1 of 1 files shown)


plugins/main/bridge/bridge.go
```diff
@@ -465,8 +461,35 @@ func cmdAdd(args *skel.CmdArgs) error {
 			if err := ipam.ConfigureIface(args.IfName, result); err != nil {
 				return err
 			}
+			return nil
+		}); err != nil {
+			return err
+		}
+
+		// check bridge port state
```
squeed, line 469: "wait for bridge port state to be Up"

plugins/main/bridge/bridge.go
```diff
@@ -465,8 +461,35 @@ func cmdAdd(args *skel.CmdArgs) error {
...
+		for i := 0; i < 3; i++ {
```
squeed, line 470: It would be nice if this had some backoff - sleeping 50, then, 500, then 1000 ms.

A more idiomatic way of writing this might be
```go
retries := []int{50, 500, 1000, 1000}
for idx, sleep := range retries {
  (check host)
  if idx == len(retries) - 1) {
    fail
  }
  time.Sleep(sleep + time.Millisecond)
}
```

hongli-my: @squeed    have been modified.

squeed:
lgtm

plugins/main/bridge/bridge.go
```diff
@@ -465,8 +461,36 @@ func cmdAdd(args *skel.CmdArgs) error {
...
+		for idx, sleep := range retries {
+			hostVeth, err := netlink.LinkByName(hostInterface.Name)
+			if err != nil {
+				return err
+			}
+			if hostVeth.Attrs().OperState == netlink.OperUp {
+				break
+			}
+
+			if idx == len(retries)-1 {
```
mars1024, line 480: If we try to escape from here, we only try 4 times and need 3 intervals, so I think we should sleep in the beginning of the loop and let `retries := []int{0,50,500,1000,1000}` , so we will make full use of `retries`.

plugins/main/bridge/bridge.go
mars1024: nit: should we make this section into to a utility function?

plugins/main/bridge/bridge.go
hongli-my: @mars1024   add it.

mars1024: Sorry, I think sleep should be the first step of the loop... just like
```golang
retries := []int{0, 50, 500, 1000, 1000}
for idx, sleep := range retries {
  time.Sleep(sleep * time.Millisecond)

  (check host)

  if idx == (len(retries) - 1) {
    fail
  }
}
```

hongli-my: sleep 0 before   do something ?

mars1024: yes, 0 means no sleep in the first time, then we will have 5 tries and 4 non-zero intervals as expectation.

hongli-my: @mars1024 ,  have changed.

dcbw:
lgtm

PR #469, opened 2020-03-31
portmap: Apply the DNAT hairpin to the whole subnet

AlbanBedel:
The DNAT hairpin rule only allow the container itself to access the ports it is exposing thru the host IP. Other containers in the same subnet might also want to access this service via the host IP, so apply this rule to the whole subnet instead of just for the container.

This is particularly useful with setups using a reverse proxy for https. With such a setup connections between containers (for ex. oauth2) have to downgrade to http, or need complex dns setup to make
use of the internal IP of the reverse proxy. On the other hand going thru the host IP is easy as that is probably what the service name already resolve to.

### DIFF
--- a/plugins/meta/portmap/portmap.go
+++ b/plugins/meta/portmap/portmap.go
@@ -48,9 +48,9 @@ const MarkMasqChainName = "CNI-HOSTPORT-MASQ"
 const OldTopLevelSNATChainName = "CNI-HOSTPORT-SNAT"
 
 // forwardPorts establishes port forwarding to a given container IP.
-// containerIP can be either v4 or v6.
-func forwardPorts(config *PortMapConf, containerIP net.IP) error {
-	isV6 := (containerIP.To4() == nil)
+// containerNet.IP can be either v4 or v6.
+func forwardPorts(config *PortMapConf, containerNet net.IPNet) error {
+	isV6 := (containerNet.IP.To4() == nil)
 
 	var ipt *iptables.IPTables
 	var err error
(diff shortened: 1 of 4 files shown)


squeed:
Oh hey, this is a really good point.
Can you look in to the test failures?

squeed:
Sorry, one last change I'd like to see: Can you update the portmap plugin README? It includes the list of expected iptables rules, which have now changed.

Once that's in, we can merge.

plugins/meta/portmap/portmap.go
```diff
@@ -249,7 +249,7 @@ func fillDnatRules(c *chain, config *PortMapConf, containerIP net.IP) {
 			copy(hpRule, ruleBase)
 
 			hpRule = append(hpRule,
-				"-s", containerIP.String(),
+				"-s", containerNet.String(),
```
dcbw, line 252: @AlbanBedel Brian and I think it's unclear and wrong to pass the unmasked IP (eg 172.30.16.2/24) to iptables, since iptables will simply mask off the host bits and actually use 172.30.16.0/24. So instead here on line 252, could you do something ilke:

```
ipnet := net.IPNet{IP: containerNet.IP.Mask(containerNet.Mask), Mask: containerNet.Mask}
hpRule = append(hpRule,
        "-s", ipnet.String(),
```

And then update the readme example iptables rules to use the network (172.16.30.0/24) instead.

AlbanBedel: It might be unusual for some but it is definitely not wrong. It is not luck or due to some implementation quirk that iptables mask the low bits, that's exactly how a network address must be interpreted. What you suggest just add useless code.

mccv1r0: FWIW, I ran this PR locally and confirmed that iptables rules show just the prefix/mask (i.e. host bits are all 0) in the iptables chain.  So the rules show e.g: `192.168.0.0/16` and not `192.168.0.2/16`

AlbanBedel: Then I'll fix the README to match what iptables report for the rules we add.

dcbw: Yes it's valid and I think clarity is not a useful thing, but I'm OK with this.

PR #472, opened 2020-04-06
Reset the route flag before moving the rule

ahenan:


ahenan:
Sometimes when the plugin is chained with other plugins e.g. sriov, the existing dynamically created routing rule will have the flags field set and that will cause the route replace to fail. Resetting the flag should sort out the mysterious "invalid argument" error reported from the kernel.

squeed:
Paging @plwhite, the author of this plugin. Does this change make sense to you?

plwhite:
Can I just check I understand @ahenan ?

Suppose you have two devices, A and B, and you are going to move B's routes to the new table. All the routes in table B that have been dynamically changed have to be moved to the new table, but have to lose their dynamic flag and become static because as a user you cannot add dynamic routes. Presumably they can still be overridden by further dynamic changes. Is that right? And does that specifically happen with SR-IOV devices?

I'm comfortable with the change, but I think this is subtle enough that a short comment above the change saying why it is necessary would make it clearer for future users.

ahenan:
@plwhite I think we are on the same page, so you are right.

I cannot really say it happens only with sriov but since that is the case, I guess we may see it with some other configurations/setups.
Agreed, it is subtle enough to be worth a brief explanation, I have added few lines regarding this.

plwhite:
Thanks - I am comfortable with that change. It seems completely safe, and I agree is necessary in at least some cases.

plwhite:
I don't have permission to merge, so will wait for a maintainer to step in.

ahenan:
No worries, thanks.

ahenan:
@maintainers any plans to close this PR?

squeed:
lgtm

dcbw:
lgtm

PR #475, opened 2020-04-15
win-bridge: add support for portMappings capability

vboulineau:
If the pluging receives portMappings in runtimeConfig, the pluing will add a NAT policy for each port mapping on the generated endpoints.

It enables HostPort usage on Windows with win-bridge (currently silently ignored).
Requires Windows >=1903 for containers to reach their own node through NodeIP + HostPort.

### DIFF
--- a/pkg/hns/netconf_windows.go
+++ b/pkg/hns/netconf_windows.go
@@ -20,10 +20,11 @@ import (
 	"fmt"
 	"net"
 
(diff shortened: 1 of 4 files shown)


squeed:
Seems reasonable... @nagiesek @daschott care to take a look?

nagiesek:
Looks like goodness to me

Issue #450, opened 2020-02-04
DHCP ignores containers mac-address when sending DHCPREQUEST to server


leier69:
Hi! 

I'm opening this because I'm trying to make a Podman container to join a PFsense LAN network on a bridge using a static DHCP lease. I thought that when podman released 1.7 (supporting setting static mac-addresses) this would be an easy task. I was wrong.

So currently im using the following network config list:
```
[root@leierpc leier]# cat /etc/cni/net.d/podnet.conflist
{
   "cniVersion": "0.4.0",
   "name": "podnet",
   "plugins": [
      {
         "type": "bridge",
         "bridge": "br0",
         "ipam": {
            "type": "dhcp"
         }
      },
      {
        "type": "tuning",
        "capabilities": {
          "mac": true
      }
      }
   ]
}
```

And this works from a Podman perspective. It asks for a IP through the dhcp.sock and the mac-address sat is indeed static. From Podmans side of things.

```
[root@leierpc leier]# podman run -d --net podnet --name="testMac"  --mac-address=3e:84:ea:ab:7d:52 docker.io/library/debian tail -f /dev/null
2020/02/04 19:58:25 5a0841b3f0347378d14ce3f7673da0380a720cf1873bc9a768f98aacc42c9d39/podnet/eth0: acquiring lease
2020/02/04 19:58:26 5a0841b3f0347378d14ce3f7673da0380a720cf1873bc9a768f98aacc42c9d39/podnet/eth0: lease acquired, expiration is 2020-02-04 21:58:26.330024245 +0100 CET m=+13799.752185010
5a0841b3f0347378d14ce3f7673da0380a720cf1873bc9a768f98aacc42c9d39

[root@leierpc leier]# podman exec testMac ip link show 
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN mode DEFAULT group default qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
3: eth0@if38: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue state UP mode DEFAULT group default 
    link/ether 3e:84:ea:ab:7d:52 brd ff:ff:ff:ff:ff:ff link-netnsid 0
```
But the DHCP plugin (`/usr/libexec/cni/dhcp daemon &`) does not respect the containers mac-address and presents the DHCP server with a seemingly random mac-address.

```
[...]

PR #480, opened 2020-04-30
macvlan: set mac address from args

clinta:
This change sets the mac address if specified during the creation of the
macvlan interface. This is superior to setting it via the tuning plugin
because this ensures the mac address is set before an IP is set,
allowing a container to get a reserved IP address from DHCP.

Related #450

### DIFF
--- a/plugins/main/macvlan/macvlan.go
+++ b/plugins/main/macvlan/macvlan.go
@@ -45,6 +45,17 @@ type NetConf struct {
 	Master string `json:"master"`
 	Mode   string `json:"mode"`
 	MTU    int    `json:"mtu"`
+	Mac    string `json:"mac,omitempty"`
+
+	RuntimeConfig struct {
+		Mac string `json:"mac,omitempty"`
+	} `json:"runtimeConfig,omitempty"`
+}
+
+// MacEnvArgs represents CNI_ARG
+type MacEnvArgs struct {
+	types.CommonArgs
+	MAC types.UnmarshallableString `json:"mac,omitempty"`
 }
 
 func init() {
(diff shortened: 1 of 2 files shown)


dcbw:
@clinta looks mostly good, but can you update the testcases too with a case to set the MAC on creation?

clinta:
A test has been added to test setting the mac address.

squeed:
Sorry for the run-around, but I'm going to ask for one more small change :-). Then we can get this merged.

Can you add support for the mac capability arg? See https://github.com/containernetworking/cni/blob/master/CONVENTIONS.md#well-known-capabilities

Thanks!

clinta:
I can do that. I'll look at the tuning plugin some more to see how it's done. I'm a bit new to CNI and added this for my own needs using podman, and I don't quite understand how the mac address in the network config is useful. Doesn't that result in every container on a network getting the same mac address?

squeed:
@clinta thanks! Let me know if you have any questions.

CNI has 3 kinds of dynamic configuration: args, CNI_ARGS, and capability arguments. The latter is newest, and allows for a sort of negotiation between the runtime and the plugins around optional parameters. It's more structured CNI_ARGS, but provides essentially the same functionality.

You don't write the mac parameter to the CNI configuration file. Rather, you tell the runtime (and libcni) where to substitute in the mac address at runtime.

Make sense?

clinta:
Thanks, that cleared it up. I wasn't seeing at first how RuntimeConfig was set by the container runtime. I've added this functionality, and another test for it.

squeed:
Looks great, thanks!

squeed:
Now, do you want to add this parameter to the bridge and p2p plugins :-) ?

PR #486, opened 2020-05-17
host-device: Bring interfaces down before moving.

cswindle:
When trying to move a master and slave interface into a container it is not possible without first bringing the interfaces down. This change ensures that the interface is set to down prior to trying to move the interface into the container. This matches the behaviour on moving an interface out of the container.

### DIFF
--- a/plugins/main/host-device/host-device.go
+++ b/plugins/main/host-device/host-device.go
@@ -193,6 +193,10 @@ func moveLinkIn(hostDev netlink.Link, containerNs ns.NetNS, ifName string) (netl
 		if err != nil {
 			return fmt.Errorf("failed to find %q: %v", hostDev.Attrs().Name, err)
 		}
+		// Devices can be renamed only when down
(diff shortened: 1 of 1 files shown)


squeed:
Interesting! Clearly the plugin is relying on the usual behavior where moving a device in to a netns causes it to go down. I tried to reproduce this with veth and vlan devices, and they all went down when moving namespaces.

What kind of device are you trying to move?

dcbw:
Also, the kernel explicitly sets the interface down when moving the netnamespace in `dev_change_net_namespace()` by calling `dev_close()`, so is something bringing the interface back up during the change operation? That would be pretty odd...

cswindle:
I am trying to move a Mellanox VF on Azure into a pod, along with the netvsc device that links to it. Where the netvsc device is the master for the VF.

squeed:
Weird. Huh. LGTM

PR #487, opened 2020-05-21
portmap: don't use unspecified address as iptables rule destination 

aojea:
It may happen that you want to map a port only in one IP family.
It can be achieved using the unspecified IP address of the
corresponding IP family as HostIP i.e.:

podman run --rm --name some-nginx -d -p 0.0.0.0:8080:80 nginx

The problem is that current implementation considers the
unspecified address valid and appends it to the iptables rule:

-A CNI-DN-60380cb3197c5457ed6ba -s 10.88.0.0/16
-d 0.0.0.0/32 -p tcp -m tcp --dport 8080 -j CNI-HOSTPORT-SETMARK

This rule is not forwarding the traffic to the corresponding port.

We should use the unspecified address only to discriminate the IP
family of the port mapping, but not use it to filter the dst.

### DIFF
--- a/plugins/meta/portmap/portmap.go
+++ b/plugins/meta/portmap/portmap.go
@@ -225,6 +225,8 @@ func fillDnatRules(c *chain, config *PortMapConf, containerNet net.IPNet) {
 	c.rules = make([][]string, 0, 3*len(entries))
 	for _, entry := range entries {
 		// If a HostIP is given, only process the entry if host and container address families match
+		// and append it to the iptables rules
+		addRuleBaseDst := false
 		if entry.HostIP != "" {
 			hostIP := net.ParseIP(entry.HostIP)
 			isHostV6 := (hostIP.To4() == nil)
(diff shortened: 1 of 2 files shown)


aojea:
/assign @dcbw @danwinship @squeed

This breaks KIND when using podman provider, because we are using unspecified addresses to do the portmapping only in the corresponding IP family

/cc @bentheelder @amwat

danwinship:
makes sense to me

plugins/meta/portmap/portmap.go
```diff
@@ -225,19 +225,27 @@ func fillDnatRules(c *chain, config *PortMapConf, containerNet net.IPNet) {
...
 		if entry.HostIP != "" {
 			hostIP := net.ParseIP(entry.HostIP)
 			isHostV6 := (hostIP.To4() == nil)
 
 			if isV6 != isHostV6 {
 				continue
 			}
+
+			// Unspecified addresses can not be used as destination
+			// TODO: check if we need to filter additional addresses
```
bboreham, line 239: We don't really want to merge a TODO - can you make a decision here?

aojea: I think that is fine to filter only unspecified addresses, I was thinking about link-local or multicast but who knows 🤷‍♂️ , unspecified are the only ones that does not work for sure.

Issue #488, opened 2020-05-22
bridge hangs when configuring coredns POD IP address


tanxiaoning007:
When I used kubeadm to setup the environment on an new system, and then apply the flannel.yaml, I find that the bridge hangs at Time.Sleep(). Although we can delete CoreDNS Pod multiple times to ensure that the pod works properly, But it is not automatic.

``` log

bridge ipam.ExecAdd result IP4:{IP:{IP:10.244.0.2 Mask:ffffff00}
Gateway:10.244.0.1 Routes:[{Dst:{IP:10.244.0.0 Mask:ffff0000} GW:<nil>}]},
DNS:{Nameservers:[] Domain: Search:[] Options:[]}

bridge ipadm.ConfigureIface result Interfaces:[{Name:cni0 Mac:4e:19:41:e9:b4:a7
Sandbox:} {Name:veth5c51018d Mac:6a:a4:9d:71:05:96 Sandbox:} {Name:eth0 Mac:
Sandbox:/proc/15863/ns/net}], IP:[{Version:4 Interface:0xc42000f2c0
Address:{IP:10.244.0.2 Mask:ffffff00} Gateway:10.244.0.1}],
Routes:[{Dst:{IP:10.244.0.0 Mask:ffff0000} GW:10.244.0.1} {Dst:{IP:0.0.0.0
Mask:00000000000000000000ffff00000000} GW:10.244.0.1}], DNS:{Nameservers:[]
Domain: Search:[] Options:[]}


kubectl exec -it coredns-78fcdf6894-8dctr -n kube-system -- /bin/sh
/ # route -n
Kernel IP routing table
Destination     Gateway         Genmask         Flags Metric Ref    Use Iface
10.244.0.0      0.0.0.0         255.255.255.0   U     0      0        0 eth0
/ #
```

``` golang
pkg/ip/addr_linux.go
	for {
		addrs, err := netlink.AddrList(link, netlink.FAMILY_ALL)
		if err != nil {
			return fmt.Errorf("could not list addresses: %v", err)
		}

		if len(addrs) == 0 {
			return nil
		}

		ok := true
		for _, addr := range addrs {
			if addr.Flags&(syscall.IFA_F_TENTATIVE|syscall.IFA_F_DADFAILED) > 0 {
				ok = false
				break // Break out of the `range addrs`, not the `for`
			}
		}

		if ok {
			return nil
		}
		if time.Now().After(deadline) {
[...]

PR #491, opened 2020-05-22
Use another thread to process time.Sleep, so even if time.Sleep hangs…

tanxiaoning007:
…, we can also use context.timeout to ensure that the program runs normally

Signed-off-by: Xiaoning Tan <txiaoning@casachina.com.cn>

### DIFF
--- a/pkg/ip/addr_linux.go
+++ b/pkg/ip/addr_linux.go
@@ -15,6 +15,7 @@
 package ip
 
 import (
+	"context"
 	"fmt"
 	"syscall"
 	"time"
@@ -23,6 +24,8 @@ import (
 )
 
 const SETTLE_INTERVAL = 50 * time.Millisecond
+
+type readOnlyTimeChan <-chan time.Time
 
 // SettleAddresses waits for all addresses on a link to leave tentative state.
 // This is particularly useful for ipv6, where all addresses need to do DAD.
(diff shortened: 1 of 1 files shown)


tanxiaoning007:
#488

tanxiaoning007:
Closes #488

tanxiaoning007:
Fixes #488

bboreham:
`SETTLE_INTERVAL` is 50ms - do you have any insight into how adding a 10 second timeout makes a discernable difference?

pkg/ip/addr_linux.go
```diff
@@ -57,12 +63,17 @@ func SettleAddresses(ifName string, timeout int) error {
 		if ok {
 			return nil
 		}
-		if time.Now().After(deadline) {
+
+		go func() {
+			timeEventChan <- time.After(SETTLE_INTERVAL)
```
squeed, line 68: time.After returns a channel... immediately. This isn't actually waiting.

tanxiaoning007: sorry, too late to reply, I still try to figure out the root cause, and I sometimes encounter this problem after applying this fix on a new k8s environment.

+++ time.After returns a channel... immediately. This isn't actually waiting.  
yes, it will be blocked by the select sentence

tanxiaoning007:
> `SETTLE_INTERVAL` is 50ms - do you have any insight into how adding a 10 second timeout makes a discernable difference?

I just suspect the time.Sleep function is blocked, so I made this change by using context.Timeout, but there is still a problem with this fix. I will continue to find out the root cause

PR #495, opened 2020-05-30
Update golang's image version in release.sh

### LINKED ISSUE #494: Update golang's image version in release.sh
Golang's image has already update to [golang:1.14-alpine](https://hub.docker.com/layers/golang/library/golang/1.14-alpine/images/sha256-911ebd34ce76d69beac233711215ece09b81f9000bb0fc4615ef3ee732a1c495?context=explore).
But we still use [golang:1.12-alpine](https://github.com/containernetworking/plugins/blob/b76fdd7c03f434a3a481bfaee0a0e48e70979486/scripts/release.sh#L19).
Should we also update it?

huchengze:
fix #494

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,9 +3,9 @@ sudo: required
 dist: xenial
 
 go:
-  - 1.11.x
   - 1.12.x
   - 1.13.x
+  - 1.14.x
 
 env:
   global:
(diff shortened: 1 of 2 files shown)


squeed:
Can you bump the travis go version as well? Thanks!

bboreham:
Travis version is here: https://github.com/containernetworking/plugins/blob/master/.travis.yml#L6-L8

We should add 1.14 and drop 1.11 from the list tested.

bboreham:
If you're feeling extra-keen you could do the other repo too: https://github.com/containernetworking/cni/blob/master/.travis.yml

PR #499, opened 2020-06-17
host-local: still allocate empty file to avoid IP leak

### LINKED ISSUE #498: host-local IP may leak
On one of our online servers, we found that there were many empty IP files created by host-local and the net range is exhausted.  
 
It may happen when host-local creates an empty file successfully, but fails to write string into it and further fails to remove it. When kubelet reties and retries to ensure a new container, this process is repeated so dozens of empty files are created. 
https://github.com/containernetworking/plugins/blob/1fb9793607361bc8ae032a12964c67e7f2b92079/plugins/ipam/host-local/backend/disk/backend.go#L62-L73

We should check if IP file is empty to determine whether an IP can be allocated. This not only prevents exhausting all free IPs, but also helps to recover leaked IP files.

chendotjs:
Fix #498

### DIFF
--- a/plugins/ipam/host-local/backend/disk/backend.go
+++ b/plugins/ipam/host-local/backend/disk/backend.go
@@ -59,9 +59,24 @@ func New(network, dataDir string) (*Store, error) {
 func (s *Store) Reserve(id string, ifname string, ip net.IP, rangeID string) (bool, error) {
 	fname := GetEscapedPath(s.dataDir, ip.String())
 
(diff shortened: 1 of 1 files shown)


squeed:
While you're at this, can you fix the root bug, and create the reserved file atomically? It would involve writing to a temporary file, then renaming that file to the final filename.

squeed:
It looks like we'll have to implement our own rename that calls `renameat2(... RENAME_NOREPLACE)`

PR #505, opened 2020-06-24
Update firewall README.md

### LINKED ISSUE #504: Document the CNI-ADMIN chain in firewall plugin README
## Background
We'd like to leverage the `firewall` plugin with the `iptables` backend and add a rule to block traffic from the containers to a specific IP. The firewall plugin creates the `CNI-ADMIN` chain. 

Is that chain meant for extension by us ? Could we safely use it to add a REJECT rule which all containers would be subject to ?

We also noticed the `iptablesAdminChainName`  in [code](https://github.com/containernetworking/plugins/blob/master/plugins/meta/firewall/firewall.go#L41-L43),  which isn't documented the README but seems to suggest that its meant for consumers of CNI to leverage.

If confirmed, I can add a PR to update the README.

xtreme-sameer-vohra:
Document `CNI-ADMIN` chain usage as well as `iptablesAdminChainName`

closes #504

### DIFF
--- a/plugins/meta/firewall/README.md
+++ b/plugins/meta/firewall/README.md
@@ -133,3 +133,17 @@ of the container as shown:
 - `-s 10.88.0.2 -m conntrack --ctstate RELATED,ESTABLISHED -j CNI-FORWARD`
 - `-d 10.88.0.2 -j CNI-FORWARD`
 
+A new chain, CNI-ADMIN is added to the CNI-FORWARD chain. CNI-ADMIN is intended as a user-controlled chain that all traffic passes through prior to the rules managed by the `firewall` plugin.
+
+`CNI-FORWARD` chain:
+- `-j CNI-ADMIN`
+
+The chain name `CNI-ADMIN` can be overridden by specifying `iptablesAdminChainName` in 
+
+```
+      {
+        "type": "firewall",
+	"backend": "iptables",
+	"iptablesAdminChainName": "SOME-OTHER-CHAIN-NAME",
+      }
+```
--- a/plugins/meta/firewall/README.md
+++ b/plugins/meta/firewall/README.md
@@ -133,12 +133,12 @@ of the container as shown:
 - `-s 10.88.0.2 -m conntrack --ctstate RELATED,ESTABLISHED -j CNI-FORWARD`
 - `-d 10.88.0.2 -j CNI-FORWARD`
 
-A new chain, CNI-ADMIN is added to the CNI-FORWARD chain. CNI-ADMIN is intended as a user-controlled chain that all traffic passes through prior to the rules managed by the `firewall` plugin.
+The `CNI-FORWARD` chain first sends all traffic to `CNI-ADMIN` chain, which is intended as an user-controlled chain for custom rules that run prior to rules managed by the `firewall` plugin. The `firewall` plugin does not add, delete or modify rules in the `CNI-ADMIN` chain.
 
 `CNI-FORWARD` chain:
 - `-j CNI-ADMIN`
 
-The chain name `CNI-ADMIN` can be overridden by specifying `iptablesAdminChainName` in 
+The chain name `CNI-ADMIN` can be overridden by specifying `iptablesAdminChainName` in the plugin configuration
 
 ```
       {


plugins/meta/firewall/README.md
```diff
@@ -133,3 +133,17 @@ of the container as shown:
 - `-s 10.88.0.2 -m conntrack --ctstate RELATED,ESTABLISHED -j CNI-FORWARD`
 - `-d 10.88.0.2 -j CNI-FORWARD`
 
+A new chain, CNI-ADMIN is added to the CNI-FORWARD chain. CNI-ADMIN is intended as a user-controlled chain that all traffic passes through prior to the rules managed by the `firewall` plugin.
```
dcbw, line 136: @xtreme-sameer-vohra thanks for the PR! Just a couple comments...

The CNI-ADMIN chain and the CNI-FORWARD chain are actually orthogonal. What I think you really mean is that CNI-ADMIN is added to the FILTER table with precedence over CNI-FORWARD. So perhaps:

`A new chain CNI-ADMIn is added to the FILTER table with precedence over any other CNI firewall plugin chains. CNI-ADMIN is intended ... (the rest of what you have here)`

plugins/meta/firewall/README.md
```diff
@@ -133,3 +133,17 @@ of the container as shown:
...
+
+`CNI-FORWARD` chain:
```
dcbw, line 138: I dont' think this section is required, becuase the forward chain doesn't acutally jump to the admin chain. Simply that when a packet hits the FILTER table, it gets sent to CNI-ADMIN for evaluation before being sent to CNI-FORWARD.

plugins/meta/firewall/README.md
```diff
@@ -133,3 +133,17 @@ of the container as shown:
...
+- `-j CNI-ADMIN`
+
+The chain name `CNI-ADMIN` can be overridden by specifying `iptablesAdminChainName` in
```
dcbw, line 141: .. in the plugin configuration:

plugins/meta/firewall/README.md
xtreme-sameer-vohra: I might have misunderstood, but based on [this section](https://github.com/containernetworking/plugins/blob/master/plugins/meta/firewall/iptables.go#L168-L169) it looks like `CNI-ADMIN` is added to the `CNI-FORWARD` chain

xtreme-sameer-vohra: Although, it would effectively have the same impact as having precedence over any other rule in the CNI-FORWARD chain

dcbw: @xtreme-sameer-vohra chains aren't added to other chains. iptables chains are completely independent, but you can jump from chain to chain. But that's not what's happening here. It's simply that packets *first* get sent to the CNI-ADMIN chain (becuase it's ordered first in the table) and if no rule in that chain matches, then they get sent to the CNI-FORWARD chain.

So all I'd like changed here is the wording so that it accurately represents what happens in iptables' logic. Thoughts? Thanks!

xtreme-sameer-vohra: Thanks for the clarification. I updated it to reflect the entry for `CNI-FORWARD`, referencing the `FORWARD` chain rather than the `FILTER` table. Does that make sense ?

bboreham:
Still seems to be one comment outstanding?

plugins/meta/firewall/README.md
xtreme-sameer-vohra: hey @dcbw  does the comment make sense with the changes ?

xtreme-sameer-vohra:
> Still seems to be one comment outstanding?

I am assuming the remaining comment is not an issue anymore but will clarify

plugins/meta/firewall/README.md
bboreham: Actually looking at the code @dcbw's statement "the forward chain doesn't actually jump to the admin chain" seems to be false.

xtreme-sameer-vohra: Hey @bboreham 
What do you suggest ?
IMHO documenting that the CNI-ADMIN chain is available to utilize by users is the most valuable

dcbw: @xtreme-sameer-vohra I still think my comment applies but I should word it better... 

1) a chain is not added to any other chain, they jump between each other. Semantics mostly but let's be correct here
2) CNI-ADMIN and FORWARD chain do not directly interact. Instead, FORWARD chain jumps to CNI-FORWARD chain which in turn jumps to CNI-ADMIN

dcbw: How about this wording instead:

The CNI-FORWARD chain first sends all traffic to the CNI-ADMIN chain, which is intended as a user-controlled chain for custom rules that run prior to the rules managed by the `firewall` plugin. The `firewall` plugin does not add, delete, or modify rules in the CNI-ADMIN chain.

xtreme-sameer-vohra: > How about this wording instead:
> 
> The CNI-FORWARD chain first sends all traffic to the CNI-ADMIN chain, which is intended as a user-controlled chain for custom rules that run prior to the rules managed by the `firewall` plugin. The `firewall` plugin does not add, delete, or modify rules in the CNI-ADMIN chain.

Thanks, updated the PR.

dcbw:
/lgtm
thanks!

PR #515, opened 2020-07-31
bridge: add non-CNI-managed interfaces to a CNI bridge

MarkSpencerTan:
We needed a way to set non-CNI-managed interfaces as ports of the CNI Bridge. This will allow a user to specify a spec like this: 

*given that mybridge-vxlan is created already*

```yaml
  config: '{
    "cniVersion": "0.3.1",
    "name": "mynet",
    "bridge": "mybridge",
    "ports": ["mybridge-vxlan"],
    "type": "bridge",
    "ipam": {}
}'
```

### DIFF
--- a/plugins/main/bridge/bridge.go
+++ b/plugins/main/bridge/bridge.go
@@ -46,16 +46,16 @@ const defaultBrName = "cni0"
 
 type NetConf struct {
 	types.NetConf
-	BrName       string `json:"bridge"`
-	Master       string `json:"master"`
-	IsGW         bool   `json:"isGateway"`
-	IsDefaultGW  bool   `json:"isDefaultGateway"`
-	ForceAddress bool   `json:"forceAddress"`
-	IPMasq       bool   `json:"ipMasq"`
-	MTU          int    `json:"mtu"`
-	HairpinMode  bool   `json:"hairpinMode"`
-	PromiscMode  bool   `json:"promiscMode"`
-	Vlan         int    `json:"vlan"`
+	BrName       string   `json:"bridge"`
+	Ports        []string `json:"ports"`
+	IsGW         bool     `json:"isGateway"`
+	IsDefaultGW  bool     `json:"isDefaultGateway"`
+	ForceAddress bool     `json:"forceAddress"`
+	IPMasq       bool     `json:"ipMasq"`
+	MTU          int      `json:"mtu"`
+	HairpinMode  bool     `json:"hairpinMode"`
+	PromiscMode  bool     `json:"promiscMode"`
+	Vlan         int      `json:"vlan"`
 }
 
 type gwInfo struct {
@@ -216,7 +216,7 @@ func bridgeByName(name string) (*netlink.Bridge, error) {
 	return br, nil
 }
 
-func ensureBridge(brName string, mtu int, promiscMode, vlanFiltering bool, master string) (*netlink.Bridge, error) {
+func ensureBridge(brName string, mtu int, promiscMode, vlanFiltering bool, ports []string) (*netlink.Bridge, error) {
 	br := &netlink.Bridge{
 		LinkAttrs: netlink.LinkAttrs{
 			Name: brName,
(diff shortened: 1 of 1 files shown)


plugins/main/bridge/bridge.go
```diff
@@ -236,6 +238,17 @@ func ensureBridge(brName string, mtu int, promiscMode, vlanFiltering bool) (*net
 		return nil, fmt.Errorf("could not add %q: %v", brName, err)
 	}
 
+	// check if master was specified for the bridge
+	if master != "" {
+		m, err := netlink.LinkByName(master)
+		if err != nil {
+			return nil, fmt.Errorf("failed to lookup master %q: %v", master, err)
```
jellonek, line 245: There is missing teardown.
If during this particular run of plugin bridge interface was created (err != nil && err != syscall.EEXIST) - the bridge interface should be removed.
It's not a big issue because... later in this func the same teardown procedure is also missing in case of any error...

squeed:
It seems like the naming here is exactly backwards :-). `LinkSetMaster(m, br)` sets m's master to br, not the other way around.

It seems like you want the plugin to move one (or more...) links in to the bridge when you create it. Given that we will adopt an existing bridge, why do you need the plugin to do it? You could set the bridge up yourself and move the interfaces.

If that won't work, or will be difficult, can you explain your use case?

MarkSpencerTan:
> It seems like the naming here is exactly backwards :-). `LinkSetMaster(m, br)` sets m's master to br, not the other way around.
> 
> It seems like you want the plugin to move one (or more...) links in to the bridge when you create it. Given that we will adopt an existing bridge, why do you need the plugin to do it? You could set the bridge up yourself and move the interfaces.
> 
> If that won't work, or will be difficult, can you explain your use case?

Thanks for pointing that out @squeed ! I was mistaken in the description of my pull request. I do want to set this bridge that we're creating as a master of another interface. In our case, it's a vxlan interface that we want to add to the bridge.

I got the order mixed up on my description but I think the code will stay the same as this is exactly what we need: "sets m's master to br".

I'll update the description to match that

plugins/main/bridge/bridge.go
MarkSpencerTan: Thanks @jellonek for that recommendation! I added the teardown for the block and also for the others lower into the function and verified that the bridge does indeed get deleted when an error is thrown

MarkSpencerTan:
@jellonek would you happen to know why this change broke the travis CI test? The error seems to be vague

dcbw:
@MarkSpencerTan the windows CI failure is a flake. I can see the utility of something like this, but the wording of the PR is confusing. What you want to do is "add non-CNI-managed interfaces to a CNI bridge".

A few comments:

1. we shouldn't be adding new config with terms like `master` since that's not actually what's happening here. The bridge is the "parent" (for lack of a better term) of the interface you're adding, or (same thing) the interface you're adding is a port of the bridge.
2. other plugins (host-device) have a device description that is either ifname, sysfs path, or PCI address and we should be consistent here; you can't always use the ifname on multiple systems
3. why is this limited to one interface?

MarkSpencerTan:
> @MarkSpencerTan the windows CI failure is a flake. I can see the utility of something like this, but the wording of the PR is confusing. What you want to do is "add non-CNI-managed interfaces to a CNI bridge".
> 
> A few comments:
> 
> 1. we shouldn't be adding new config with terms like `master` since that's not actually what's happening here. The bridge is the "parent" (for lack of a better term) of the interface you're adding, or (same thing) the interface you're adding is a port of the bridge.
> 2. other plugins (host-device) have a device description that is either ifname, sysfs path, or PCI address and we should be consistent here; you can't always use the ifname on multiple systems
> 3. why is this limited to one interface?

Thanks so much for the feedback @dcbw ! I've made the changes to make the PR more clear, I agree that calling it as `Ports` instead of Master would make it more clear as to what the feature is doing.
I also like the idea of supporting multiple Ports so I've made the changes to accomodate that. It will accept a []string instead of just a string.

I'm still looking up changes to accomodate #2, but thanks for the heads up on that one

MikeZappa87:
@MarkSpencerTan are you still working on this? I am creating a vxlan plugin and this would be helpful since I need a way to attach the vxlan interface to the bridge

PR #521, opened 2020-08-05
Bump Go version to 1.13 and 1.14

dcbw:
@containernetworking/cni-maintainers


squeed:
lgtm, feel free to merge on green.

dcbw:
I don't understand the windows failures, it's complaining that "you may only call It inside Describe, Context, or When" yet that's exactly where the testcase calls it from...

PR #523, opened 2020-08-21
Fix race condition in GetCurrentNS

### LINKED ISSUE #524: GetCurrentNS doesn't always return the "current" netns
We use `ns.WithNetNSPath` and `ip.SetupVethWithName` to set up container networking in CNI plugin [Antrea](https://github.com/vmware-tanzu/antrea). Recently we found a bug https://github.com/vmware-tanzu/antrea/issues/1113 that some Pods' network interfaces were not in the host netns after restarting Worker Nodes. The network interfaces were finally found in other Pods's netns, and we think it's because of a race condition in `GetCurrentNS`. @antoninbas had a good description about how it happened: https://github.com/vmware-tanzu/antrea/issues/1113#issuecomment-676611494.

The fact that It doesn't happen to other CNIs might be because Antrea has a long running daemon processing all CNI requests and executing the above steps. So if the race condition happens, the "hostNS" might have been another container's netns.

We use this library widely in our project and would appreciate this could be fixed from upstream side. This is a potential fix for it: https://github.com/containernetworking/plugins/pull/523.

tnqn:
In GetCurrentNS, If there is a context-switch between
getCurrentThreadNetNSPath and GetNS, another goroutine may execute in
the original thread and change its network namespace, then the original
goroutine would get the updated network namespace, which could lead to
unexpected behavior, especially when GetCurrentNS is used to get the
host network namespace in netNS.Do.

The added test has a chance to reproduce it with "-count=50".

The patch fixes it by locking the thread in GetCurrentNS.

Fixes #524

### DIFF
--- a/pkg/ns/ns_linux.go
+++ b/pkg/ns/ns_linux.go
@@ -26,6 +26,11 @@ import (
 
 // Returns an object representing the current OS thread's network namespace
 func GetCurrentNS() (NetNS, error) {
+	// Lock the thread in case other goroutine executes in it and changes its
+	// network namespace after getCurrentThreadNetNSPath(), otherwise it might
+	// return an unexpected network namespace.
+	runtime.LockOSThread()
+	defer runtime.UnlockOSThread()
 	return GetNS(getCurrentThreadNetNSPath())
 }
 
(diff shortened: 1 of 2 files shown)


squeed:
Agreed. test failures are windows

PR #525, opened 2020-09-04
Introduce a new VRF CNI meta plugin (also bump netlink package to v1.1.0)

fedepaol:
This plugin allows to create a [VRF](https://www.kernel.org/doc/Documentation/networking/vrf.txt) with the given name (or use the existing one if any) in the target namespace, and to allocate the interface to it.

VRFs make it possible to use multiple routing tables on the same namespace and allows isolation among interfaces within the same namespace. On top of that, this allows different interfaces to have overlapping CIDRs (or even addresses).

This is only useful in addition to other plugins.

The configuration is pretty simple and looks like:
```
{
    "type": "vrf",
    "vrfname": "blue"
}
```

## Note: 

In order to limit the size of the PR, additional tests will follow in a separate PR depending on this one.
Plus, another PR will be submitted to add an optional routingTableID parameter that will allow the user to specify the desired routing table ID.

Finally, a separate commit is added to upgrade the netlink version, which is required to be able to set VRFs as master (with the current version, that is possible only for bridges interfaces).


s1061123:
You may need to add vrf in https://github.com/containernetworking/plugins/blob/master/plugins/linux_only.txt if vrf is only for linux.

plugins/meta/vrf/main.go
```diff
@@ -0,0 +1,164 @@
+package main
```
s1061123, line 1: Could you please add copyright section here?

plugins/meta/vrf/vrf.go
```diff
@@ -0,0 +1,105 @@
...
+		return 0, fmt.Errorf("findFreeRoutingTableID: Failed to find links %v", err)
+	}
+	for _, l := range links {
+		if vrf, ok := l.(*netlink.Vrf); ok {
+			if vrf.Table > maxTable {
+				maxTable = vrf.Table
+			}
+		}
+	}
+	return (maxTable + 1), nil
```
s1061123, line 104: routing table id seems to be unsigned char, uint8 in golang. Could you please double check it?
https://github.com/torvalds/linux/blob/master/include/uapi/linux/rtnetlink.h#L227

plugins/meta/vrf/vrf.go
```diff
@@ -0,0 +1,105 @@
...
+func findFreeRoutingTableID() (uint32, error) {
+	var maxTable uint32
+	links, err := netlink.LinkList()
+	if err != nil {
+		return 0, fmt.Errorf("findFreeRoutingTableID: Failed to find links %v", err)
+	}
+	for _, l := range links {
+		if vrf, ok := l.(*netlink.Vrf); ok {
+			if vrf.Table > maxTable {
+				maxTable = vrf.Table
```
s1061123, line 100: What is happen if all routing tables are consumed and there is no free routing table?

fedepaol:
> You may need to add vrf in https://github.com/containernetworking/plugins/blob/master/plugins/linux_only.txt if vrf is only for linux.

Thanks, I would have been looking for it today :-)

plugins/meta/vrf/vrf.go
fedepaol: Sure, thanks for pointing out!

plugins/meta/vrf/vrf.go
fedepaol: Good point. Let's return an error in that case.

Now that I look back to it, probably the logic needs also to be strengthened a bit here, trying to look for the first free instead of taking the max + 1, especially considering that with a following PR users will be able to choose the tableid.

plugins/meta/vrf/vrf.go
fedepaol: @s1061123  I _think_ that's a u32.

See https://elixir.bootlin.com/linux/latest/source/drivers/net/vrf.c#L44
https://elixir.bootlin.com/linux/latest/source/drivers/net/vrf.c#L1339

Which also sticks with the type used by the netlink library: https://github.com/containernetworking/plugins/blob/53377543e1e91e984033e0af5e19bc9bb6973c65/vendor/github.com/vishvananda/netlink/link_linux.go#L2745

plugins/meta/vrf/vrf.go
fedepaol: Done

plugins/meta/vrf/main.go
fedepaol: Done

plugins/meta/vrf/README.md
```diff
@@ -0,0 +1,27 @@
+# vrf plugin
+
+## Overview
+
+This plugin creates a [VRF](https://www.kernel.org/doc/Documentation/networking/vrf.txt) in the network namespace,and assigns it the interface passed in the arguments. If the VRF is already present in the namespace, it only adds the interface to it.
```
phoracek, line 5: namespace and

phoracek (changes requested):

plugins/meta/vrf/README.md
fedepaol: Done, thanks!

plugins/meta/vrf/main.go
```diff
@@ -0,0 +1,178 @@
...
+		v, err := findVRF(conf.VRFName)
+		if err != nil {
+			return err
+		}
+
+		interfaces, err := assignedInterfaces(v)
+		if err != nil {
+			return err
+		}
+
```
jellonek, line 98: So there is no removing of interfaces from VRF because we are expecting that (happy path) a next in chain plugin will destroy the interface, what will in fact also remove it from VRF?

IMO this should also cover unhappy path, so the interface will be left to next in chain plugin for del in the shape in which next plugin left it, with master interface set to what was before.

plugins/meta/vrf/main.go
```diff
@@ -0,0 +1,178 @@
...
+		// Meaning, we are deleting the last interface assigned to the VRF
+		if len(interfaces) == 1 && interfaces[0].Attrs().Name == args.IfName {
+			err = netlink.LinkDel(v)
+			if err != nil {
```
jellonek, line 102: Please remove this `if` with its body or better ...

plugins/meta/vrf/main.go
```diff
@@ -0,0 +1,178 @@
...
+				return err
+			}
+		}
+		return err
```
jellonek, line 106: ... `return nil` explicitly there.

plugins/meta/vrf/main.go
```diff
@@ -0,0 +1,178 @@
...
+	if err != nil {
+		return err
+	}
+
+	if conf.PrevResult == nil {
+		return fmt.Errorf("missing prevResult from earlier plugin")
+	}
+
+	err = ns.WithNetNSPath(args.Netns, func(_ ns.NetNS) error {
+		v, err := findVRF(conf.VRFName)
```
jellonek, line 55: Please s/v/vrf/

plugins/meta/vrf/main.go
```diff
@@ -0,0 +1,178 @@
...
+	return types.PrintResult(result, conf.CNIVersion)
+}
+
+func cmdDel(args *skel.CmdArgs) error {
+	conf, _, err := parseConf(args.StdinData)
+	if err != nil {
+		return err
+	}
+	err = ns.WithNetNSPath(args.Netns, func(_ ns.NetNS) error {
+		v, err := findVRF(conf.VRFName)
```
jellonek, line 89: Please s/v/vrf/

plugins/meta/vrf/main.go
```diff
@@ -0,0 +1,178 @@
...
+		return err
+	}
+
+	// Ensure we have previous result.
+	if conf.PrevResult == nil {
+		return fmt.Errorf("missing prevResult from earlier plugin")
+	}
+
+	err = ns.WithNetNSPath(args.Netns, func(_ ns.NetNS) error {
+		v, err := findVRF(conf.VRFName)
```
jellonek, line 127: Please s/v/vrf/

plugins/meta/vrf/main.go
```diff
@@ -0,0 +1,178 @@
...
+		if err != nil {
+			return err
+		}
+		ii, err := assignedInterfaces(v)
```
jellonek, line 131: Please s/ii/vrfInterfaces/ or some other more verbose name than `ii`.

jellonek (commented):
Mostly small nits, but overall IMO looks good.

plugins/meta/vrf/main.go
fedepaol: Ah! You are right.
The only thing I am not sure of, is the part where I should restore the master, as the CNI has no memory.

@jellonek what if I unset the master, and let the ADD command return an error if the given interface already has one master set?

fedepaol:
@jellonek I _think_ I addressed all the comments. I went with the "no override master" policy, meaning that if the interface already has a master, the add will fail instead of overriding the current master. Please take another look.

plugins/meta/vrf/main.go
```diff
@@ -0,0 +1,178 @@
...
+	"github.com/containernetworking/cni/pkg/skel"
+	"github.com/containernetworking/cni/pkg/types"
+	"github.com/containernetworking/cni/pkg/types/current"
+	"github.com/containernetworking/cni/pkg/version"
+
+	"github.com/containernetworking/plugins/pkg/ns"
+	bv "github.com/containernetworking/plugins/pkg/utils/buildversion"
+)
+
+// VrfNetConf represents the firewall configuration.
```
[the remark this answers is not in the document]
fedepaol: !!!!! That reveals where I "took inspiration from". Sorry I did not notice, changing it.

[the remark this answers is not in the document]
fedepaol: Done

plugins/meta/vrf/main.go
jellonek: Yeah, this would be what i would expect. Good idea!

plugins/meta/vrf/main.go
```diff
@@ -0,0 +1,183 @@
...
+	conf := VRFNetConf{}
+	if err := json.Unmarshal(data, &conf); err != nil {
+		return nil, nil, fmt.Errorf("failed to load netconf: %v", err)
+	}
+
+	if conf.VRFName == "" {
+		return nil, nil, fmt.Errorf("configuration is expected to have a valid vrf name")
+	}
+
+	// Parse previous result.
```
jellonek, line 164: Hmmm... Same comment is below, so this looks a bit suspicious as real parsing is done later, while there is only checking if it was provided in raw form.

fedepaol: Ouch

fedepaol: At least I was consistent in leaving the comments from where I copied the structure :laughing:  https://github.com/containernetworking/plugins/blob/0a1421a08cc909aa67360f749e90bab60206a01f/plugins/meta/firewall/firewall.go#L75

fedepaol: Done

plugins/meta/vrf/README.md
```diff
@@ -0,0 +1,27 @@
...
+## Overview
+
+This plugin creates a [VRF](https://www.kernel.org/doc/Documentation/networking/vrf.txt) in the network namespace and assigns it the interface passed in the arguments. If the VRF is already present in the namespace, it only adds the interface to it.
+
+As a table id is mandatory, the plugin generates a new one for each different VRF that is added to the namespace.
+
+It does not create any network interfaces and therefore does not bring connectivity by itself.
+It is only useful when used in addition to other plugins.
+
+## Configuration
```
aojea, line 12: can you make more detailed README with examples on how to use it with another plugins?
i.e. https://github.com/containernetworking/plugins/tree/master/plugins/meta/firewall

plugins/meta/vrf/main.go
```diff
@@ -0,0 +1,182 @@
...
+	err = ns.WithNetNSPath(args.Netns, func(_ ns.NetNS) error {
+		vrf, err := findVRF(conf.VRFName)
+
+		if _, ok := err.(netlink.LinkNotFoundError); ok {
+			vrf, err = createVRF(conf.VRFName)
+		}
+
+		if err != nil {
+			return err
+		}
```
aojea, line 63: should we check first for the erro and then if it is a LinkNotFound error?
```
		if err != nil {
		     if _, ok := err.(netlink.LinkNotFoundError); ok {
			vrf, err = createVRF(conf.VRFName)
                      }	 else {
                         return err
                     }
             }

File: plugins/meta/vrf/README.md (hunk 14 above)
Comment (fedepaol) [reply plugins/meta/vrf/README.md:12]: Sure, will do!

File: plugins/meta/vrf/main.go (hunk 15 above)
Comment (fedepaol) [reply plugins/meta/vrf/main.go:63]: I don't like extra nesting, and the above version is behaviourally identical, as the type check will fail if err is nil.

But if you think this other version is more clear I can change.

File: plugins/meta/vrf/main.go [hunk 16]
```diff
@@ -0,0 +1,182 @@
...
+
+func cmdDel(args *skel.CmdArgs) error {
+	conf, _, err := parseConf(args.StdinData)
+	if err != nil {
+		return err
+	}
+	err = ns.WithNetNSPath(args.Netns, func(_ ns.NetNS) error {
+		vrf, err := findVRF(conf.VRFName)
+		if err != nil {
+			return err
```
aojea, line 91: @dcbw is not the DEL idempotent? should it return nil ?

plugins/meta/vrf/main.go
aojea: I was more curious about what happens if err=nil with this?
```err.(netlink.LinkNotFoundError);```
from your comments I assume is not a problem

Comment (aojea) [reply plugins/meta/vrf/main.go:63]: you are right, there is lot of nesting and is very ugly to read :)

File: plugins/meta/vrf/main.go (hunk 16 above)
Comment (fedepaol) [reply plugins/meta/vrf/main.go:91]: Ah, maybe only if err is a `LinkNotFoundError`?

Comment (aojea) [reply plugins/meta/vrf/main.go:91]: you need a more CNI guy here, but the spec says

> Plugins should generally complete a DEL action without error even if some resources are missing. For example, an IPAM plugin should generally release an IP allocation and return success even if the container network namespace no longer exists, unless that network namespace is critical for IPAM management. While DHCP may usually send a 'release' message on the container network interface, since DHCP leases have a lifetime this release action would not be considered critical and no error should be returned. For another example, the bridge plugin should delegate the DEL action to the IPAM plugin and clean up its own resources (if present) even if the container network namespace and/or container network interface no longer exist.

https://github.com/containernetworking/cni/blob/master/SPEC.md

Comment (fedepaol) [reply plugins/meta/vrf/main.go:91]: Thanks for the explanation!
Yes, I will modify the behavior so the DEL fails only if there's a real failure and not because the resources are not there.

**Comment by bboreham** [comment]
Do you know anything about the CI failures?
```
/home/travis/gopath/src/github.com/containernetworking/plugins/plugins/meta/vrf/vrf_test.go:32
  passes prevResult through unchanged [It]
  /home/travis/gopath/src/github.com/containernetworking/plugins/plugins/meta/vrf/vrf_test.go:80
  Unexpected error:
      <*errors.errorString | 0xc000085c20>: {
          s: "cmdAdd failed: could not add VRF vrf0: operation not supported",
      }
      cmdAdd failed: could not add VRF vrf0: operation not supported
  occurred
  /home/travis/gopath/src/github.com/containernetworking/plugins/plugins/meta/vrf/vrf_test.go:96
```

fedepaol:
> Do you know anything about the CI failures?
> 
> ```
> /home/travis/gopath/src/github.com/containernetworking/plugins/plugins/meta/vrf/vrf_test.go:32
>   passes prevResult through unchanged [It]
>   /home/travis/gopath/src/github.com/containernetworking/plugins/plugins/meta/vrf/vrf_test.go:80
>   Unexpected error:
>       <*errors.errorString | 0xc000085c20>: {
>           s: "cmdAdd failed: could not add VRF vrf0: operation not supported",
>       }
>       cmdAdd failed: could not add VRF vrf0: operation not supported
>   occurred
>   /home/travis/gopath/src/github.com/containernetworking/plugins/plugins/meta/vrf/vrf_test.go:96
> ```

Yep, vrf support is introduced in ubuntu bionic. For this reason I opened this other pr https://github.com/containernetworking/plugins/pull/526 (probably I should have cross-referenced them).
That introduced a couple of other issues as mentioned in the pr itself.

bboreham (approved):
LGTM, thanks for all the work

fedepaol:
> LGTM, thanks for all the work

It was a pleasure :-)

PR #526, opened 2020-09-07
Bump up ubuntu CI ubuntu version to bionic

fedepaol:
VRF support was introduced in ubuntu bionic, while it's missing in Xenial.

This also introduces a change in the behaviour of nc command.
On one hand, it requires a new line to send the buffer on the other side, on the other it hangs waiting for new input.

To address this, a timeout was introduced to avoid the tests to hang, plus the buffer sent is terminated with a new line character.

This is needed to make the tests of the new cni plugin introduced in https://github.com/containernetworking/plugins/pull/525 pass.


aojea:
it fails in windows containers :smile:

bboreham:
The Windows CI failure was fixed by #529; if you rebase those errors should go away.

fedepaol:
> The Windows CI failure was fixed by #529; if you rebase those errors should go away.

Will do, thanks!

fedepaol:
@bboreham all green

integration/integration_linux_test.go
```diff
@@ -225,23 +226,29 @@ func (n Namespace) Del() {
...
+		cmd = exec.CommandContext(ctx, "ip", "netns", "exec", netns, bin, "-v", address, strconv.Itoa(port))
 	} else {
-		cmd = exec.Command("nc", address, strconv.Itoa(port))
+		cmd = exec.CommandContext(ctx, bin, address, strconv.Itoa(port))
 	}
 	cmd.Stdin = bytes.NewBuffer([]byte(message))
 	cmd.Stderr = GinkgoWriter
 	out, err := cmd.Output()
+	Expect(err).Should(Or(Not(HaveOccurred()),
+		ContainSubstring("signal: killed"),
```
bboreham, line 247: what's going on here?  the test passes if the subprocess is killed?

fedepaol: When I changed ubuntu version, all the tests using `nc` started to hang causing a major headache. My understanding is that, as per https://unix.stackexchange.com/questions/332163/netcat-send-text-to-echo-service-read-reply-then-exit , the bump in ubuntu resulted in a different nc version, which hangs waiting for new input while the previous did not. I could have added the -q parameter, making it incompatible with the OpenBSD version.

To make this test portable, and work on both versions here I use exec.CommandContext which kills the process if the timeout on the context expires. As a side effect, having the process still running waiting for extra input but killed for timeout is not an error condition.

The success or failure of the test is validated in the following lines, where we check for the output.

fedepaol: @bboreham does it make sense? Not sure if there was a more elegant way to do so.

bboreham: Does it mean the test takes a minute longer?

fedepaol: It's been a while, but yes. If nc hangs, we wait for the timeout on the context to check the output.
I can make it way shorter though, 2 seconds should be more than enough.

bboreham: We discussed this at the maintainers meeting.  It's way more convoluted than I realised.

The test where this is used [specifically checks the time taken](https://github.com/containernetworking/plugins/blob/e78e6aa5b9fd7e3e66f0cb997152c44c2a4e43df/integration/integration_linux_test.go#L188), so any approach with a timeout is unwise.

Looking at [the nc man page](https://www.commandlinux.com/man-page/man1/nc.traditional.1.html), it should exit when the _network_ side closes. In this test the network side is an echo server which apparently does [close the connection](https://github.com/containernetworking/plugins/blob/e78e6aa5b9fd7e3e66f0cb997152c44c2a4e43df/pkg/testutils/echosvr/main.go#L52), so perhaps you could investigate why that isn't enough?

fedepaol: Sure, will do and get back.
Thanks for now!

dcbw:
@fedepaol another thought; why don't we write an echo client binary in pkg/testutils/echoclient that just sends the input data, checks the result, and then exits. We already have an echo server, so why not a client that does exactly what we want and nothing more?

fedepaol:
> @fedepaol another thought; why don't we write an echo client binary in pkg/testutils/echoclient that just sends the input data, checks the result, and then exits. We already have an echo server, so why not a client that does exactly what we want and nothing more?

I like this even more, we would have the behaviour under control (and what's more I don't need to revamp the tools for running the travis environment locally )

fedepaol:
@dcbw @bboreham PTAL.

I replaced nc with a local echoclient which is supposed to be more portable.

Now that I had to look into the echoserver implementation, I also think I found the reason why nc hanged, which is because we weren't explicitly sending the \n as part of the standard input.

pkg/testutils/echo/client/client.go
```diff
@@ -0,0 +1,45 @@
...
+	if err != nil {
+		panic(fmt.Sprintf("Failed to open connection to [%s] %v", *target, err))
+	}
+	defer conn.Close()
+
+	_, err = conn.Write([]byte(*payload))
+	if err != nil {
+		panic("Failed to send payload")
+	}
+	_, err = conn.Write([]byte("\n"))
```
bboreham, line 29: If we're adding a newline here why do we also add one at https://github.com/containernetworking/plugins/pull/526/files#diff-8466c05fa47a47331ae1d46458c3d11dR229 and https://github.com/containernetworking/plugins/pull/526/files#diff-8466c05fa47a47331ae1d46458c3d11dR229 ?

plugins/meta/portmap/portmap_suite_test.go
```diff
@@ -40,14 +41,18 @@ func TestPortmap(t *testing.T) {
...
+var echoServerBinaryPath, echoClientBinaryPath string
 
 var _ = SynchronizedBeforeSuite(func() []byte {
-	binaryPath, err := gexec.Build("github.com/containernetworking/plugins/pkg/testutils/echosvr")
+	serverBinaryPath, err := gexec.Build("github.com/containernetworking/plugins/pkg/testutils/echo/server")
 	Expect(err).NotTo(HaveOccurred())
-	return []byte(binaryPath)
+	clientBinaryPath, err := gexec.Build("github.com/containernetworking/plugins/pkg/testutils/echo/client")
+	Expect(err).NotTo(HaveOccurred())
+	return []byte(fmt.Sprintf("%s,%s", serverBinaryPath, clientBinaryPath))
```
bboreham, line 51: `strings.Join` ?

bboreham (commented):
Seems good; I have a couple of questions

plugins/meta/portmap/portmap_suite_test.go
fedepaol: Done, don't have a strong opinion on that since the list is two elements.

pkg/testutils/echo/client/client.go
fedepaol: Good catch! That's a leftover. Removed it.

fedepaol:
> Cute, but you shouldn't check the built binary in :-). Please build as part of `./test.sh`

That was a local copy I did not notice I committed.
Removed it, it's built by gexec.

fedepaol:
@bboreham @squeed I think this addressed all the comments, please take another look

bboreham (approved):
lgtm, thanks!

PR #527, opened 2020-09-08
Flannel log

dverbeir:
Logging level can be set using the "LogLevel" field of the NetConf:
```
  {
    "name": "cbr0",
    "cniVersion": "0.3.1",
    "type": "flannel",
    "logLevel": "debug",
    ...
  }
```
It can be set to "panic", "error", "verbose" or "debug", and defaults to "panic".

### DIFF
--- a/plugins/meta/flannel/flannel.go
+++ b/plugins/meta/flannel/flannel.go
@@ -29,6 +29,7 @@ import (
 	"path/filepath"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/containernetworking/cni/pkg/invoke"
 	"github.com/containernetworking/cni/pkg/skel"
@@ -41,6 +42,7 @@ import (
 const (
 	defaultSubnetFile = "/run/flannel/subnet.env"
 	defaultDataDir    = "/var/lib/cni/flannel"
+	defaultLogLevel   = "error"
 )
 
 type NetConf struct {
@@ -52,6 +54,7 @@ type NetConf struct {
 	DataDir       string                 `json:"dataDir"`
 	Delegate      map[string]interface{} `json:"delegate"`
 	RuntimeConfig map[string]interface{} `json:"runtimeConfig,omitempty"`
+	LogLevel      string                 `json:"logLevel"`
 }
 
 type subnetEnv struct {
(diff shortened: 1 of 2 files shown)


dverbeir:
I forgot to mention that, after implementing this change, I found issue https://github.com/containernetworking/cni/issues/486 which brings up the same problem and suggests a similar solution. But it had no feedback...

bboreham:
It looks like this PR is actually two PRs: one matching the title and one matching the description.

On the routes side, I wonder if it would be nicer to use the `ipam` format for specifying them, and copy that through to the `host-local` delegate.  This would save inventing a new identifier `extraRoutes`.

Either way, it feels like the new feature should be documented in [the README](https://github.com/containernetworking/plugins/blob/master/plugins/meta/flannel/README.md).

dverbeir:
> It looks like this PR is actually two PRs: one matching the title and one matching the description.
> 
> On the routes side, I wonder if it would be nicer to use the `ipam` format for specifying them, and copy that through to the `host-local` delegate. This would save inventing a new identifier `extraRoutes`.
> 
> Either way, it feels like the new feature should be documented in [the README](https://github.com/containernetworking/plugins/blob/master/plugins/meta/flannel/README.md).

Thanks Bryan for the feedback. I removed the routes change from this PR and will create another one, based on your ipam format recommendation.

jellonek (approved):
LGTM

dcbw:
Can we rename 'panic' to 'fatal' to be more consistent with the various Go logging frameworks?

dverbeir:
> Can we rename 'panic' to 'fatal' to be more consistent with the various Go logging frameworks?

Looks like a good idea indeed. I also fixed the default logLevel which didn't really defaulted to "error" as documented.

dverbeir:
Could we merge this? I know it doesn't bring a whole lot of value but it can still be useful when troubleshooting.

dverbeir:
I removed the doc update since the doc has moved to a separate repo. I'd a corresponding change there once this gets merged ;-)

bboreham:
Sorry, haven't looked at this in a while.  Needs rebase.

My $0.02: broadly this can go in: I had a reaction that it should use a logging library, but the code is so short it hardly seems worthwhile.
The core implementation should probably be pulled out to somewhere more central than `flannel`, so other plugins can take advantage, and be consistent.

MikeZappa87:
This probably should be closed since the flannel plugin is no longer in this project. When it comes to logging, I have in my queue to improve logging in the CNI for the containerd project. This might be useful as a reference.

PR #532, opened 2020-09-10
flannel: allow input ipam parameters as basis for delegate

dverbeir:
(NOTE: This work started under PR #527 which was a combination of this and the addition of logging; now split into 2 PRs)

This change allows providing an 'ipam' section as part of the input network configuration for flannel. It is then used as basis to construct the ipam parameters provided to the delegate.

All parameters from the input ipam are preserved except:
* 'subnet' which is set to the flannel host subnet
* 'routes' which is complemented by a route to the flannel network.

One use case of this feature is to allow adding back the routes to the cluster services and/or to the hosts (HostPort) when using isDefaultGateway=false. In that case, the bridge plugin does not install a default route and, as a result, only pod-to-pod connectivity would be available.

Example:
```
  {
    "name": "cbr0",
    "cniVersion": "0.3.1",
    "type": "flannel",
    "ipam": {
      "routes": [
        {
          "dst": "192.168.242.0/24"
        },
        {
          "dst": "10.96.0.0/12"
        }
      ],
      "unknown-param": "value"
    },
    "delegate": {
      "hairpinMode": true,
      "isDefaultGateway": false
    }
    ...
  }
```

This results in the following 'ipam' being provided to the delegate:
```
    {
      "routes" : [
        {
          "dst": "192.168.242.0/24"
        },
        {
          "dst": "10.96.0.0/12"
        },
        {
          "dst" : "10.1.0.0/16"
        }
      ],
      "subnet" : "10.1.17.0/24",
      "type" : "host-local"
      "unknown-param": "value"
    }
```

where "10.1.0.0/16" is the flannel network and "10.1.17.0/24" is the host flannel subnet.

Note that this also allows setting a different ipam 'type' than "host-local".

### DIFF
--- a/plugins/meta/flannel/flannel.go
+++ b/plugins/meta/flannel/flannel.go
@@ -46,6 +46,8 @@ const (
 type NetConf struct {
 	types.NetConf
 
+	// IPAM field "replaces" that of types.NetConf which is incomplete
+	IPAM          map[string]interface{} `json:"ipam,omitempty"`
 	SubnetFile    string                 `json:"subnetFile"`
 	DataDir       string                 `json:"dataDir"`
 	Delegate      map[string]interface{} `json:"delegate"`
@@ -87,6 +89,18 @@ func loadFlannelNetConf(bytes []byte) (*NetConf, error) {
 	}
 
 	return n, nil
+}
+
+func getIPAMRoutes(n *NetConf) ([]types.Route, error) {
+	rtes := []types.Route{}
+
+	if n.IPAM != nil && hasKey(n.IPAM, "routes") {
+		buf, _ := json.Marshal(n.IPAM["routes"])
+		if err := json.Unmarshal(buf, &rtes); err != nil {
+			return rtes, fmt.Errorf("failed to parse ipam.routes: %w", err)
+		}
+	}
+	return rtes, nil
 }
 
 func loadFlannelSubnetEnv(fn string) (*subnetEnv, error) {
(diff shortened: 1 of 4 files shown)


plugins/meta/flannel/flannel.go
```diff
@@ -46,6 +46,8 @@ const (
 type NetConf struct {
 	types.NetConf
 
+	// IPAM field "replaces" that of types.NetConf which is incomplete
+	IPAM          map[string]interface{} `json:"ipam,omitempty"`
```
bboreham, line 50: Why use a string map rather than a struct ?
(It came as a surprise to me that `containernetworking/cni` does not return any useful IPAM struct,  but even so, why don't you?)

bboreham (commented):
Seems broadly OK but I have one question.

plugins/meta/flannel/flannel.go
dverbeir: Even though my own use case is about the routes, I thought it would be nice to allow transparently passing any IPAM parameter that could apply to the delegate (bridge or other). In order not to limit the possibilities there, a string map looked appropriate to me (I'm not aware of any way to unmarshal into a struct while preserving and giving access to all extra fields not defined in the struct).
You will see in flannel_linux.go:getDelegateIPAM() that the incoming string map, when present, is used as basis for the delegate ipam.

dcbw:
@rajatchopra does this look sane to you? Would you use an override capability for flannel IPAM?

dverbeir:
I re-based on the latest master so the tests pass now.

How to move on with this change? Do we need the feedback from @rajatchopra?
Due to the fairly transparent way in which it's implemented, it could prove useful for various cases.

bboreham (approved):
lgtm

squeed:
seems fine

PR #535, opened 2020-09-18
VRF CNI: Additional tests

fedepaol:
This is based on https://github.com/containernetworking/plugins/pull/525, trying to limit the size of the PRs.

The new tests expand coverage, checking deletion, ip address handling,
0.4.0 compatibility, behaviour in case of multiple vrfs.

PR #536, opened 2020-09-18
VRF CNI: Add an optional table parameter.

fedepaol:
When specified from the user, the VRF will get assigned to the given tableid instead of having the CNI to choose for a free one.

This is based on https://github.com/containernetworking/plugins/pull/535 and https://github.com/containernetworking/plugins/pull/525

### DIFF
--- a/plugins/meta/vrf/README.md
+++ b/plugins/meta/vrf/README.md
@@ -4,7 +4,7 @@
 
 This plugin creates a [VRF](https://www.kernel.org/doc/Documentation/networking/vrf.txt) in the network namespace and assigns it the interface passed in the arguments. If the VRF is already present in the namespace, it only adds the interface to it.
 
-As a table id is mandatory, the plugin generates a new one for each different VRF that is added to the namespace.
+The table id is mandatory for VRF but optional in the CNI configuration. If not specified, the plugin generates a new one for each different VRF that is added to the namespace.
 
 It does not create any network interfaces and therefore does not bring connectivity by itself.
 It is only useful when used in addition to other plugins.
@@ -50,4 +50,5 @@ The only configuration is the name of the VRF, as per the following example:
 
 The following [args conventions](https://github.com/containernetworking/cni/blob/master/CONVENTIONS.md#args-in-network-config) are supported:
 
-* `vrfname` (string, optional): The name of the VRF to be created and to be set as master of the interface
+* `vrfname` (string): The name of the VRF to be created and to be set as master of the interface
+* `table` (int, optional): The route table to be associated to the created VRF.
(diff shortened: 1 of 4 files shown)


squeed:
Hey there,
Can you squash this to two commits: one that bumps netlink, and one that introduces the plugin? Thanks!

squeed:
Aha, nevermind, I didn't realize the state of the PRs.

bboreham:
Looks ok; please add something in the README explaining the new field.

fedepaol:
> Looks ok; please add something in the README explaining the new field.

Done

Elegant996:
Would it possible to add support for RuntimeConfig or should that be a separate issue/PR? Thanks!

fedepaol:
> Would it possible to add support for RuntimeConfig or should that be a separate issue/PR? Thanks!

Not sure I understood the question. What parameters would you expect to be in the runtimeConfig?

Elegant996:
> Not sure I understood the question. What parameters would you expect to be in the runtimeConfig?

Both the `vrfname` and `table` could be added to the RuntimeConfig similar to how the `mac` works for the `tuning` plugin.

https://github.com/containernetworking/plugins/blob/440dcc331b692e64bcf0893a22ceda1826ee9f61/plugins/meta/tuning/tuning.go#L48-L54

Thanks!

fedepaol:
> > Not sure I understood the question. What parameters would you expect to be in the runtimeConfig?
> 
> Both the `vrfname` and `table` could be added to the RuntimeConfig similar to how the `mac` works for the `tuning` plugin.
> 
> https://github.com/containernetworking/plugins/blob/440dcc331b692e64bcf0893a22ceda1826ee9f61/plugins/meta/tuning/tuning.go#L48-L54
> 
> Thanks!

I am pretty ignorant on the topic, but I checked about RuntimeConfig (here)[https://github.com/containernetworking/cni/blob/master/CONVENTIONS.md#dynamic-plugin-specific-fields-capabilities--runtime-configuration] and my understanding is that those well known parameters are passed to the CNI from the runtime which knows about them.

Here, both vrfName and tableID are parameters choosen by the user and the runtime does not know about them.

squeed:
Seems reasonable. Can you rebase, then we'll merge.

You can add the convention - which is often injected by meta-plugins like Multus - at a later date if there's demand.

fedepaol:
> Seems reasonable. Can you rebase, then we'll merge.
> 
> You can add the convention - which is often injected by meta-plugins like Multus - at a later date if there's demand.

Rebased

bboreham (approved):
lgtm

fedepaol:
Seems the failure is not related (and only on the arm lane)
@squeed can you merge this?

mccv1r0 (approved):

PR #537, opened 2020-09-30
Port plugins to CNI 1.0.0 and increase old verison test coverage

dcbw:


test_linux.sh
```diff
@@ -26,7 +26,7 @@ else
     echo "without coverage profile generation..."
 fi
 
-PKG=${PKG:-$(cd ${GOPATH}/src/${REPO_PATH}; go list ./... | xargs echo)}
+PKG=${PKG:-$(cd ${GOPATH}/src/${REPO_PATH}; go list ./... | grep -v dhcp | xargs echo)}
```
squeed, line 29: what's this.......?

dcbw: leftover, removed

bboreham (approved):
seems legit

dcbw:
We agreed that each plugin must at least test that 1.0.0 works for ADD.

dcbw:
@squeed @bboreham @jellonek @mccv1r0 I ran out of steam doing the "meta" plugins, got as far as flannel. What do you think now?

bboreham (approved):
Wow!

Issue #493, opened 2020-05-28
Tuning Plugin: Revert Values on Delete


killianmuldoon:
Tuning plugin sets the MTU of interfaces, but it doesn't reset to the previous value on reset. We've seen issues with this where some containers looking for the same type of interface configure MTU while other networks assume default MTU at 1500.

The CmdDel method in the tuning CNI is a TODO with:
>	// TODO: the settings are not reverted to the previous values. Reverting the
	// settings is not useful when the whole container goes away but it could be
	// useful in scenarios where plugins are added and removed at runtime.

Is there a plan to implement reverting settings with the tuning plugin? Could we implement the MTU revert by itself or would there need to be a broader reset function for other settings touched by the tuning cni?

squeed:
It's tough, because the tuning plugin wouldn't necessarily know what the MTU was at the beginning.

squeed:
Can you outline the sort of "reset" scenario you're talking about? Generally, tuning only touches the container-side of the network, so I would expect any customizations to be moot when the interface is deleted.

I suppose this could affect host-device? But that strikes me as odd that you'd want a different MTU in vs. outside the container.

killianmuldoon:
The issue we're seeing is with getting an unexpected MTU where it's not set explicitly. 

i.e. Workload 1 requires jumbo frame MTU of 9000, tuning plugin sets it, and it receives the correctly configured interface. Once it finishes its work the container is deleted. 

Workload 2 looks for the same kind of interface, but assumes a default MTU i.e. 1500. It picks up the interface, but then hits an issue because the MTU is actually at 9000.

killianmuldoon:
@squeed do you think this is worth pursuing in a PR?

dcbw:
@killianmuldoon sure, lets see what that PR looks like.

PR #540, opened 2020-10-09
tuning: revert values on delete

ipatrykx:
This PR will introduce NIC's attributes restoration as described in [Issue #493: Tuning Plugin: Revert Values on Delete](https://github.com/containernetworking/plugins/issues/493).

Currently when NIC's attributes (MAC, promiscuous mode, MTU) are changed by Tuning plugin, they remain as set even when pod is deleted and NIC is moved back to host's networking namespace. With this PR, JSON-based backup file will be created in `dataDir` (defaults to: `/var/lib/cni/tuning`), and it will be used to restore the NIC's attributes to original ones on pod deletion.

Readme and tests were updated as well to reflect those changes.


squeed:
Any reason not to restore sysctls as well?

bboreham (commented):
Tens of lines of cosmetic changes make a PR harder to review, so please take them out.
If you like, make a separate PR that is just those changes.

plugins/meta/tuning/tuning.go
```diff
@@ -165,6 +183,122 @@ func changeMtu(ifName string, mtu int) error {
...
+		if err = os.MkdirAll(backupDir, 0600); err != nil {
+			return fmt.Errorf("failed to create backup directory: %v", err)
+		}
+	}
+
+	data, err := json.MarshalIndent(config, "", " ")
+	if err != nil {
+		return fmt.Errorf("failed to marshall data for %q: %v", ifName, err)
+	}
+	if err = ioutil.WriteFile(path.Join(backupDir, ifName+".json"), data, 0600); err != nil {
```
mars1024, line 217: Is it a wise choice to cache something inside container? Or `/var/lib/cni/` on host is a better place? which is just being used by IPAM plugin `host-local`.

plugins/meta/tuning/tuning.go
```diff
@@ -258,7 +405,7 @@ func cmdCheck(args *skel.CmdArgs) error {
 
 	// Parse previous result.
 	if tuningConf.RawPrevResult == nil {
-		return fmt.Errorf("Required prevResult missing")
+		return fmt.Errorf("required prevResult missing")
```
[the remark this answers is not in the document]
ipatrykx: I believe I have deleted all the cosmetic changes.

plugins/meta/tuning/tuning.go
ipatrykx: Backup file is stored in host's filesystem (`/var/lib/cni/tunig` by default). I think that storing backup in the container wouldn't be wise as container is not exactly persistent and might be for example deleted outside of the K8s. Additionally I think that if NIC is host's resource, therefore NIC's config should be stored on host itself.

ipatrykx:
> Any reason not to restore sysctls as well?

@squeed I believe that sysctls are set per namespace so setting those in container does not affect the host's setup, therefore I think that it is not necessary to restore this settings in any way.

```
user@control-plane ~ $ kubectl exec -it tuning-pod -- cat /proc/sys/net/core/somaxconn
500
user@worker ~ $ cat /proc/sys/net/core/somaxconn
128
```

plugins/meta/tuning/tuning.go
```diff
@@ -165,6 +183,122 @@ func changeMtu(ifName string, mtu int) error {
...
+		if tuningConf.Promisc {
+			config.Promisc = new(bool)
+			*config.Promisc = (link.Attrs().Promisc != 0)
+		}
+		if tuningConf.Mtu != 0 {
+			config.Mtu = link.Attrs().MTU
+		}
+	}
+
+	backupDir := path.Join(backupPath, containerID)
```
squeed, line 205: Rather than creating a directory per container, just join container id + ifname - that's what we do in other plugins.

ipatrykx: Done :)

ipatrykx: Oh, I forgot about tests. Fixed.

plugins/meta/tuning/tuning.go
```diff
@@ -205,6 +332,11 @@ func cmdAdd(args *skel.CmdArgs) error {
 			}
 		}
 
+		isTuned := (tuningConf.Mac != "" || tuningConf.Mtu != 0 || tuningConf.Promisc)
+		if err = createBackup(args.IfName, args.ContainerID, tuningConf.DataDir, isTuned, tuningConf); err != nil {
```
bboreham, line 336: could we just skip creating a backup if nothing was tuned?

plugins/meta/tuning/tuning.go
```diff
@@ -60,6 +66,14 @@ type IPAMArgs struct {
 	Mtu     *int               `json:"mtu,omitempty"`
 }
 
+// configToRestore will contain interface attributes that should be restored on cmdDel
+type configToRestore struct {
```
bboreham, line 70: It strikes me that this struct could be a part of `TuningConf`, which would remove the need to repeat all fields.

(and maybe part of `IPAMArgs` ?)

plugins/meta/tuning/README.md
```diff
@@ -46,6 +46,8 @@ The parameters, "mac", "mtu" and "promisc", changes the interface attributes as
 }
 ```
 
+All the interface attributes set by tuning plugin will be restored to original values on container's removal.
```
bboreham, line 49: except sysctls?

bboreham (commented):
Some thoughts below

plugins/meta/tuning/README.md
ipatrykx: I've made the info more precise.

plugins/meta/tuning/tuning.go
ipatrykx: I initially though that it would be better to create a file even if nothing's changed just to avoid using the file existence check as indicator of whenever something has to be restored. However changed that as it seems that it may be more clear in general.

plugins/meta/tuning/tuning.go
ipatrykx: @bboreham I've changed the code. Renamed `configToRestore` to `ifConfig` and made it a part of `IPAMArgs`. I chose `IPAMArgs` because `Promisc` field in that structure is a pointer to bool same as in `configToRestore`. In case of json, simple bool will be omitted if it's value will be `false` and `omitempty` were specified. And I believe that in case of future development or just to avoid ambiguity it would be better to save exact value of `Promisc` in backup file whether it is true or false. 

Which leads me to another question I was wondering regarding the `Promisc bool` in `TuningConf`.  In original [code](https://github.com/containernetworking/plugins/blob/7e2c95b9f88bf274b6fe973d24bc90f3ab2cfcf6/plugins/meta/tuning/tuning.go#L354) of `cmdAdd` there is
```
if tuningConf.Promisc != false {
	if err = changePromisc(args.IfName, true); err != nil {
		return err
	}
}
```

So if user will have a NIC with promiscuous mode enabled, and would like to switch promisc off using tuning (by setting promisc to false), then tuning plugin will not disable the promiscuous mode on that NIC as the code inside if statement will not be executed. Is this an expected behaviour, e.g. linked to problems with some particular NIC or driver? If not, then `Promisc bool` could be changed to pointer, alongside MAC and MTU variables, and therefore `ifConfig` could be used in both `IPAMArgs` and `TuningConf`. I think that this shouldn't be a part of this PR anyway, so this question is just out of curiosity.

plugins/meta/tuning/tuning.go
```diff
@@ -54,10 +60,15 @@ type TuningConf struct {
 }
 
 type IPAMArgs struct {
-	SysCtl  *map[string]string `json:"sysctl"`
-	Mac     *string            `json:"mac,omitempty"`
-	Promisc *bool              `json:"promisc,omitempty"`
-	Mtu     *int               `json:"mtu,omitempty"`
+	SysCtl   *map[string]string `json:"sysctl"`
+	IfConfig *ifConfig          `json:"ifConfig,omitempty"`
```
squeed, line 64: This would be changing the configuration format, which we shouldn't do. Do you just want to embed the IfConfig struct?

ipatrykx: This was changed because of @bboreham request: https://github.com/containernetworking/plugins/pull/540#discussion_r513554021
I have explained this under that comment. If you need any more explanation, don't hesitate to ask.

Should I revert this change?

bboreham: The explanation helps, but it does not override the point that we can't just make a breaking change to the API because you think it's nicer.
I did say "maybe"; if it can't be done, so be it.

ipatrykx: @squeed @bboreham Just reverted the API changes.

plugins/meta/tuning/tuning.go
```diff
@@ -165,6 +182,116 @@ func changeMtu(ifName string, mtu int) error {
...
+
+	if len(errStr) > 0 {
+		return fmt.Errorf(strings.Join(errStr, "; "))
+	}
+
+	if err = os.Remove(filePath); err != nil {
+		return fmt.Errorf("failed to remove file %v: %v", filePath, err)
+	}
+
+	if err = delDirIfEmpty(backupPath); err != nil {
```
squeed, line 271: Now that we are using a single data directory, there's probably no point in removing the directory at all.

plugins/meta/tuning/tuning.go
```diff
@@ -37,9 +40,12 @@ import (
 	bv "github.com/containernetworking/plugins/pkg/utils/buildversion"
 )
 
+const defaultDataDir = "/var/lib/cni/tuning"
```
squeed, line 43: I know we use `/var/lib/cni` for lots of other plugins, but that was a mistake I'm trying to undo. Can you make the default directory `/run/cni/tuning` please?

squeed: By using `/run`, we don't have to worry about cleaning up stale containers on reboot.

plugins/meta/tuning/tuning.go
ipatrykx: Removed `delDirIfEmpty` and it's invocation.

plugins/meta/tuning/tuning.go
ipatrykx: Done :)

plugins/meta/tuning/tuning.go
```diff
@@ -165,6 +181,95 @@ func changeMtu(ifName string, mtu int) error {
...
+	}
+
+	file, err := ioutil.ReadFile(filePath)
+	if err != nil {
+		return fmt.Errorf("failed to open file %q: %v", filePath, err)
+	}
+
+	config := configToRestore{}
+	if err = json.Unmarshal([]byte(file), &config); err != nil {
+		return fmt.Errorf("failed to unmarshall file %q: %v", filePath, err)
```
squeed, line 233: If this fails, just log an error and return; we don't want to block deletion.

plugins/meta/tuning/tuning.go
```diff
@@ -165,6 +181,95 @@ func changeMtu(ifName string, mtu int) error {
...
+	}
+
+	var errStr []string
+
+	_, err = netlink.LinkByName(ifName)
+	if err != nil {
+		return fmt.Errorf("failed to get %q: %v", ifName, err)
```
squeed, line 240: Log here and return; the interface is clearly deleted :-)

plugins/meta/tuning/tuning.go
```diff
@@ -165,6 +181,95 @@ func changeMtu(ifName string, mtu int) error {
...
+	}
+
+	if config.Mtu != 0 {
+		if err = changeMtu(ifName, config.Mtu); err != nil {
+			err = fmt.Errorf("failed to restore MAC address: %v", err)
```
squeed, line 245: "failed to restore MTU"

plugins/meta/tuning/tuning.go
```diff
@@ -239,11 +350,18 @@ func cmdAdd(args *skel.CmdArgs) error {
...
-	// TODO: the settings are not reverted to the previous values. Reverting the
-	// settings is not useful when the whole container goes away but it could be
-	// useful in scenarios where plugins are added and removed at runtime.
-	return nil
+	tuningConf, err := parseConf(args.StdinData, args.Args)
+	if err != nil {
+		return err
+	}
+
+	err = ns.WithNetNSPath(args.Netns, func(_ ns.NetNS) error {
```
squeed, line 360: now that I think about it, do we want to block deletion at all if this fails? When would there be an error that could be recoverable on retry?

plugins/meta/tuning/tuning.go
ipatrykx: How should be an error logged?

plugins/meta/tuning/tuning.go
ipatrykx: Same as above - just returned nil for now.

plugins/meta/tuning/tuning.go
ipatrykx: Fixed :)

plugins/meta/tuning/tuning.go
ipatrykx: OK, I think you're right. Changed this to return always nil at the end, same as the 'stub' function that was here previously.

ipatrykx: Should this
```
tuningConf, err := parseConf(args.StdinData, args.Args)
	if err != nil {
		return err
	}
```

also be forced to return nil?

squeed:
looks good to me.

PR #541, opened 2020-10-15
travis: multi-arch, go1.15, and more

squeed:
- bump to go1.15, remove all other versions
- fix nasty bit where vendor dir was dirty
- add ppc64le, s390x test runs
- cross-complile in a single job to save jobs


squeed:
For some reason ppc64le is failing the sbr unit test. It seems inconsequential. If we look in to it, then we could add ppc64le as another tested arch.

squeed:
Update: Re-added ppc64le, added skips to the SBR plugin; filed #543 to fix that.

.travis.yml
```diff
@@ -1,60 +1,74 @@
...
 
 install:
   - go get github.com/onsi/ginkgo/ginkgo
   - go get github.com/containernetworking/cni/cnitool
   - go get golang.org/x/tools/cmd/cover
   - go get github.com/modocache/gover
   - go get github.com/mattn/goveralls
-  - go mod vendor
+  # installing the tools has changed our vendoring; undo that
+  - git checkout -- go.mod go.sum
```
bboreham, line 47: Ick.  Can't you put what you need in the `require` section so it will come out the same?

squeed: maybe, maybe not; there's no guarantee that these `go get` commands won't change vendor in the future.

The problem is we're using `go get` as both a library manager and a packaging system for installing CLI tools. And `go get` makes it hard to do the latter in a sane way.

squeed:
aha, s390x might be too slow for our tests to run. Bummer.

squeed:
ok, s390x is now allow_failure. This is ready for review & merge.

PR #546, opened 2020-11-03
VRF: extend supported version to 0.3.1 too

fedepaol:
The e2e tests already covers both versions, and since the plugin is meant to be used in chains, this will augment the scope of the plugins it can be used with.

### DIFF
--- a/plugins/meta/vrf/main.go
+++ b/plugins/meta/vrf/main.go
@@ -38,7 +38,7 @@ type VRFNetConf struct {
 }
 
 func main() {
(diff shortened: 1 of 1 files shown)


squeed:
seems reasonable. I checked with @fedepaol that vrf doesn't need `prevResult` on delete, so this shouldn't matter.

fedepaol:
Yep, on top of that, I have 0.3.1 tests in place for deletion:

https://github.com/containernetworking/plugins/blob/5fc309a69956b5a8747b2e8e97bdc34230683f80/plugins/meta/vrf/vrf_test.go#L357

mccv1r0 (approved):

PR #551, opened 2020-11-10
dhcp: fix example configuration

xieyanker:
The current cni config has an extra comma and cannot be parsed normally, the kubelet will report an error as follows:
"Error loading CNI config file: error parsing configuration: invalid character '}' looking for beginning of object key string"


mccv1r0 (approved):

xieyanker:
@mccv1r0  The Travis CI build failed, but it was not caused by mine, how to retest it.

Issue #123, opened 2018-02-20
portmap: delete UDP conntrack entries on teardown


squeed:
As observed in kubernetes/kubernetes#59033, a quick teardown + spinup of portmappings can cause UDP "flows" to be lost, thanks to stale conntrack entries.

From the original issue:

> 1. A server pod exposes UDP host port.
> 2. A client sends packets to the server pod thru the host port. This creates a conntrack entry.
> 3. The server pod's IP changs due to whatever reason, such as pod gets recreated.
> 4. Due to the nature of UDP and conntrack, new request from the same client to the host port will keep hitting the stale conntrack entry.
> 5. Client observes traffic black hole.

brantburnett:
I have found that this bug doesn't seem to only apply to changing pod IPs.  It can also apply if there is incoming traffic that is being dropped because there is no pod, and then a pod is added that should start receiving the traffic.

Complete details and steps to reproduce: https://github.com/projectcalico/felix/issues/1880

vmendi:
We are also affected by this.  We lose metrics whenever there's a restart of the Datadog agent. Is there any plans to fix it? any workaround available?

Thanks

Suckzoo:
Seems like this issue has been opened for a year. Is there any plan to fix this issue? Or, could you let me know how to fix this problem by my hand?

RohanKurane:
Hello,

I think I am hitting a similar issue. 

I deploy a pod with ```hostport``` and created using ```type:portmap```. 
Then when I try to delete the pod, and re-deploy the same pod with same name, I get the following error - 
```Warning  FailedCreatePodSandBox  1m    kubelet, ip-xxxxxxxxx.ec2.internal  Failed create pod sandbox: rpc error: code = Unknown desc = failed to add hostport mapping for sandbox k8s_server_default_436fb151-71b1-11e9-b2d9-128d1a3304a4_0(7993b317dc75381c0ed08a75019fde6c2fe70aaf5697af6cc1f0e85c7afddfd6): cannot open hostport 50051 for pod k8s_server_default_436fb151-71b1-11e9-b2d9-128d1a3304a4_0_: listen tcp :50051: bind: address already in use
```

Is this the same issue ? If so, is there a workaround till this issue is fixed ? 

Thanks

Simwar:
It seems like this PR aimed to fix the issue: https://github.com/kubernetes/kubernetes/pull/59286
[...]

PR #553, opened 2020-11-18
Delete stale UDP conntrack entries when adding new Portmaps to containers

aojea:
conntrack does not have any way to track UDP connections, so
it relies on timers to delete a connection.
The problem is that UDP is connectionless, so a client will keep
sending traffic despite the server has gone, thus renewing the
conntrack entries.
Pods that use portmaps to expose UDP services need to flush the existing 
conntrack entries on the port exposed when they are created, 
otherwise conntrack will keep sending the traffic to the previous IP
until the connection age (the client stops sending traffic)

Fixes: https://github.com/containernetworking/plugins/issues/123


aojea:
WIP it needs a test

pkg/utils/conntrack.go
```diff
@@ -0,0 +1,84 @@
...
+func getNetlinkFamily(isIPv6 bool) netlink.InetFamily {
+	if isIPv6 {
+		return unix.AF_INET6
+	}
+	return unix.AF_INET
+}
+
+// DeleteConntrackEntriesForDstIP delete the conntrack entries for the connections
+// specified by the given destination IP and protocol
+func DeleteConntrackEntriesForDstIP(dstIP string, protocol uint8) error {
```
aojea, line 44: not needed, but who knows if we'll need it :shrug:  :)

plugins/meta/portmap/main.go
```diff
@@ -89,12 +90,24 @@ func cmdAdd(args *skel.CmdArgs) error {
 		if err := forwardPorts(netConf, netConf.ContIPv4); err != nil {
 			return err
 		}
+		// Delete conntrack entries for UDP to avoid conntrack blackholing traffic
+		// due to stale connections. We do that after the iptables rules are set, so
+		// the new traffic picks them
+		if err := deletePortmapStaleConnections(netConf.RuntimeConfig.PortMaps, unix.AF_INET); err != nil {
+			return err
```
squeed, line 97: I know its unlikely, but we should probably just output the error to stderr and continue; if we fail to delete conntrack, it shouldn't block container startup.

bboreham (approved):
LGTM if the error return is modified

pkg/utils/conntrack.go
```diff
@@ -0,0 +1,80 @@
...
+	filter := &netlink.ConntrackFilter{}
+	filter.AddIP(netlink.ConntrackOrigDstIP, ip)
+	filter.AddProtocol(protocol)
+
+	n, err := netlink.ConntrackDeleteFilter(netlink.ConntrackTable, family, filter)
+	if err != nil {
+		return fmt.Errorf("error deleting connection tracking state for protocol: %d IP: %s, error: %v", protocol, ip, err)
+	}
+	if n == 0 {
+		log.Printf("no entries found for protocol: %d IP: %s", protocol, ip)
```
squeed, line 60: Hmm, can we get rid of this log line (and it's equivalent below)? Logging for CNI is poorly-defined - yeah, we need to fix that - and this might be confusing to administrators.

plugins/meta/portmap/main.go
```diff
@@ -89,12 +91,24 @@ func cmdAdd(args *skel.CmdArgs) error {
 		if err := forwardPorts(netConf, netConf.ContIPv4); err != nil {
 			return err
 		}
+		// Delete conntrack entries for UDP to avoid conntrack blackholing traffic
+		// due to stale connections. We do that after the iptables rules are set, so
+		// the new traffic uses them. Failures are informative only.
+		if err := deletePortmapStaleConnections(netConf.RuntimeConfig.PortMaps, unix.AF_INET); err != nil {
+			log.Printf("fail to delete conntrack connections for %s: %v", netConf.ContIPv4.IP, err)
```
squeed, line 98: (keep this log line, though - it's useful)

wording: "failed to delete stale UDP conntrack entries for "

mrueg:
@squeed would it be possible to get a new minor/patch release including this change?

aojea:
I think we should wait until we have have confirmation that it solves the original issue https://github.com/kubernetes/kubernetes/issues/95258 , just for double checking

AnishShah:
I was able to test it using the following yaml that I used to reproduce the error originally and it works.

```
apiVersion: v1
kind: Namespace
metadata:
  name: udp
---
apiVersion: v1
kind: Pod 
metadata:
  name: udp-server
  namespace: udp
spec:
    containers:
      - name: udp-server
        image: aardvarkx1/udp-server
        imagePullPolicy: Always
        ports:
          - containerPort: 10001
            protocol: UDP
            hostPort: 10001
            name: udp-test
---
apiVersion: v1
kind: Pod
metadata:
  name: udp-client
  namespace: udp
spec:
    containers:
      - name: udp-client
        image: aardvarkx1/udp-client
        imagePullPolicy: Always
        env:
          - name: SERVER_ADDRESS
            valueFrom:
              fieldRef:
                fieldPath: status.hostIP
```

AnishShah:
Can we cut a minor release now that the issue is closed?

aojea:
@bboreham @squeed what do you think? it will be nice to have a new release
 
https://github.com/containernetworking/plugins/compare/v0.8.7...master

DanTulovsky:
I just hit this issue on GKE (version: 1.19.8-gke.1000).  How do I check if this fix made it there or not?

Thank you
Dan

aojea:
Your cniplugin version has to be 0.9.0 or newer

DanTulovsky:
Yes, thank you.  Given this is all installed by GKE automatically, do you know where I could see what the plugin version being used is?

AnishShah:
Check in `/home/kubernetes/bin`

DanTulovsky:
What should I be looking for in there?

```
gke-cluster0-pool0-6e5903aa-2vmq /home/kubernetes/bin # ls -al
total 431208
drwxr-xr-x  2 root    root       4096 Mar 18 13:37 .
drwxr-xr-x 11 root    root       4096 Mar 18 13:37 ..
-rwxr-xr-x  1 root    root    4559393 Mar 18 13:37 bandwidth
-rwxr-xr-x  1 root    root    4671350 Feb 18  2020 bridge
-rwxr-xr-x  1 root    root   34485216 Mar 18 13:37 calico
-rwxr-xr-x  1 root    root   33682272 Mar 18 13:37 calico-ipam
-r-xr--r--  1 root    root     123073 Feb 23 12:22 configure-helper.sh
-rwxr-xr-x  1 root    root      26051 Feb 23 12:22 configure-kubeapiserver.sh
-rwxr-xr-x  1 root    root      29988 Mar 18 13:37 configure.sh
-rwxr-xr-x  1 chronos users  30473598 Aug 28  2020 crictl
-rwxr-xr-x  1 root    root   12120140 Feb 18  2020 dhcp
-rwxr-xr-x  1 root    root    5945760 Feb 18  2020 firewall
-rwxr-xr-x  1 root    root    3069034 Mar 18 13:37 flannel
-rwxr-xr-x  1 root    root    4629262 Feb 18  2020 gke
-rwxr-xr-x  1 root    root   35364941 Feb 23 12:22 gke-exec-auth-plugin
-rwxr-xr-x  1 root    root     627321 Feb 23 12:22 gke-exec-auth-plugin-license
-rwxr-xr-x  1 root    root      22000 Feb 23 12:22 gke-internal-configure-helper.sh
-rwxr-xr-x  1 root    root       1762 Feb 23 12:22 gke-internal-configure.sh
-rwxr-xr-x  1  344930 89939   7508876 Nov 19 00:27 health-checker
-r-xr--r--  1 root    root       4140 Feb 23 12:22 health-monitor.sh
-rwxr-xr-x  1 root    root    4153025 Feb 18  2020 host-device
-rwxr-xr-x  1 root    root    3957620 Mar 18 13:37 host-local
-rwxr-xr-x  1 root    root    4314508 Feb 18  2020 ipvlan
-rwxr-xr-x  1 root    root   42176512 Feb 23 09:24 kubectl
-rwxr-xr-x  1 root    root  115725640 Feb 23 09:24 kubelet
-rwxr-xr-x  1  344930 89939  18257656 Nov 19 00:27 log-counter
-rwxr-xr-x  1 root    root    3650379 Mar 18 13:37 loopback
-rwxr-xr-x  1 root    root    4389532 Feb 18  2020 macvlan
-rwxr-xr-x  1  344930 89939  44245360 Nov 19 00:27 node-problem-detector
-rwxr-xr-x  1 root    root    4327403 Mar 18 13:37 portmap
-rwxr-xr-x  1 root    root    4590104 Feb 18  2020 ptp
-rwxr-xr-x  1 root    root    3392736 Feb 18  2020 sbr
-rwxr-xr-x  1 root    root    2885430 Feb 18  2020 static
-rwxr-xr-x  1 root    root    3736919 Mar 18 13:37 tuning
-rwxr-xr-x  1 root    root    4314356 Feb 18  2020 vlan
```

The only somewhat relevant thing I see is:

gke-internal-configure-helper.sh:  "cniVersion": "0.3.1",

Is that the right thing to look at?  That seems like a very old version.

bboreham:
If you just run the binary `/home/kubernetes/bin/portmap` it should print its version to stdout.

DanTulovsky:
gke-cluster0-pool0-6e5903aa-2vmq / # /home/kubernetes/bin/portmap
CNI portmap plugin v0.8.0

Thank you.  I assume this is something that Google needs to fix, or does this come from somewhere else that they inherit?

DanTulovsky:
I see https://github.com/kubernetes/kubernetes/issues/95258#issuecomment-738173998 ... thank you.

PR #556, opened 2020-12-09
bridge: fix testcase to check addresses we care about

dcbw:
Instead of checking the total number of addresses, which might vary
depending on the IPv6 Privacy Address settings of the distro being
used, just check that we have the number of non-link-local addresses
we expect.

@containernetworking/cni-maintainers @squeed


bboreham (approved):
lgtm

PR #560, opened 2020-12-22
ipvlan: fix possible link leak.

chendotjs:
The change makes best effort to clean up ipvlan link if error happens. 
This would also make ipvlan consistent with the macvlan plugin.

### DIFF
--- a/plugins/main/ipvlan/ipvlan.go
+++ b/plugins/main/ipvlan/ipvlan.go
@@ -135,7 +135,7 @@ func createIpvlan(conf *NetConf, ifName string, netns ns.NetNS) (*current.Interf
 		return nil, err
 	}
 
-	mv := &netlink.IPVlan{
+	iv := &netlink.IPVlan{
 		LinkAttrs: netlink.LinkAttrs{
 			MTU:         conf.MTU,
 			Name:        tmpName,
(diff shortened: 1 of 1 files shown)


bboreham (approved):
Thanks!

plugins/main/ipvlan/ipvlan.go
```diff
@@ -145,13 +145,14 @@ func createIpvlan(conf *NetConf, ifName string, netns ns.NetNS) (*current.Interf
...
 
-	if err := netlink.LinkAdd(mv); err != nil {
+	if err := netlink.LinkAdd(iv); err != nil {
 		return nil, fmt.Errorf("failed to create ipvlan: %v", err)
 	}
 
 	err = netns.Do(func(_ ns.NetNS) error {
 		err := ip.RenameLink(tmpName, ifName)
 		if err != nil {
+			_ = netlink.LinkDel(iv)
```
mars1024, line 155: IMO, if we want to avoid link leaking, all possible errors should be walked, e.g. [https://github.com/containernetworking/plugins/blob/48a97a7ad16cb839d4871cd493faca13129737ca/plugins/main/ipvlan/ipvlan.go#L162](https://github.com/containernetworking/plugins/blob/48a97a7ad16cb839d4871cd493faca13129737ca/plugins/main/ipvlan/ipvlan.go#L162)

chendotjs: I suspect that this kind of cleanup error is difficult and not that important to handle.  Give it a logging or aggregate multiple errors then return? Any thoughts?

mars1024: Could you put link recycling in `defer`? And delete link by name?

PR #562, opened 2020-12-31
fix a bug in ifb link teardown with ip addr

gongguan:
In brandwidth plugin scenario, ifb link(bwpxxx) has no ip address, so an error `failed to get IP addresses for ...` always returned when cmdDel called and this link device won't be removed.


gongguan:
ref https://github.com/containernetworking/plugins/blob/master/pkg/ip/link_linux.go#L209-L212

gongguan:
Sorry for the noise, it was fixed by https://github.com/containernetworking/plugins/commit/449700f7ea1952e630bbbb7a05f9e4ae7e3864c4

PR #563, opened 2020-12-31
deps: bump coreos/go-iptables

### LINKED ISSUE #544: 'DEL' can not clean up all ip6tables
We execute the plugins via commandline, because everything works fine i think we are using them right - but if
we add ipv6 it can not cleanup some ip6tables with an error msg;

```
{
   "cniVersion":"0.4.0",
   "name":"hynet",
   "type":"bridge",
   "bridge":"hybr0",
   "isGateway":true,
   "ipMasq":true,
   "ipam":{
      "type":"host-local",
      "ranges":[
         [
            {
               "subnet":"fd00:101::/64",
               "gateway":"fd00:101::1"
            }
         ],
         [
            {
               "subnet":"10.10.0.0/16",
               "gateway":"10.10.0.1"
            }
         ]
      ]
   }
}
```
After "ADD" everything works as expected and looks good:
(veth, bridge on host added, eth0 on ns added, iptables created, networking fine)

`ip6tables -t nat --list`

```
Chain POSTROUTING (policy ACCEPT)
target     prot opt source               destination         
CNI-63fd59f190a7b6095f99898e  all      fd00:101::9          anywhere             /* name: "hynet" id: "testler" */

Chain CNI-63fd59f190a7b6095f99898e (1 references)
target     prot opt source               destination         
ACCEPT     all      anywhere             fd00:101::/64        /* name: "hynet" id: "testler" */
MASQUERADE  all      anywhere            !ff00::/8             /* name: "hynet" id: "testler" */

```

If we now try to "DEL":
(with prevResult added)

```
{
    "code": 999,
    "msg": "running [/usr/sbin/ip6tables -t nat -D POSTROUTING -s fd00:101::9/64 -j CNI-63fd59f190a7b6095f99898e -m comment --comment name: \"hynet\" id: \"testler\" --wait]: exit status 1: iptables: Bad rule (does a matching rule exist in that chain?).\n"
}
```
We get the error above... but the rule is in fact removed, but not the CNI-Chain.

`ip6tables -t nat --list`

```
Chain POSTROUTING (policy ACCEPT)
target     prot opt source               destination         

Chain CNI-63fd59f190a7b6095f99898e (1 references)
target     prot opt source               destination         
ACCEPT     all      anywhere             fd00:101::/64        /* name: "hynet" id: "testler" */
MASQUERADE  all      anywhere            !ff00::/8             /* name: "hynet" id: "testler" */

```

It seems that it tries to delete the same rule twice and then get this error. If we remove ipv6 everything works, ADD and DEL, with all resources and iptables cleaned up.

tlwr:
Closes #544

The above issue describes a situation where using the bridge plugin with IPv6 addresses prevented `DEL` from working correctly.

`DEL` seems to be failing in the body of `TeardownIPMasq`

This arises because:

* twice delete postrouting rules: `ipn.String()` `ipn.IP.String()` #279
* we are using a version of go-iptables which is bugged for v6

PR coreos/go-iptables/pull/74 describes why this does not work. The error message is not being checked correctly.

Using a later version of go-iptables means that
* when the second `ipt.Delete` fails (this is okay)
* we will correctly interpret this as an non-fatal error
* `TeardownIPMasq` will not prematurely exit the method
* `ipt.ClearChain` now can run
* `ipt.DeleteChain` now can run

This explains why this was working for v4 but not v6

--

Edit: as mentioned in comments below, I've updated this to vendor `go-iptables@v0.5.0`

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -8,7 +8,7 @@ require (
 	github.com/alexflint/go-filemutex v0.0.0-20171022225611-72bdc8eae2ae
 	github.com/buger/jsonparser v0.0.0-20180808090653-f4dd9f5a6b44
 	github.com/containernetworking/cni v0.8.0
-	github.com/coreos/go-iptables v0.4.5
+	github.com/coreos/go-iptables v0.5.0
 	github.com/coreos/go-systemd v0.0.0-20180511133405-39ca1b05acc7
 	github.com/d2g/dhcp4 v0.0.0-20170904100407-a1d1b6c41b1c
 	github.com/d2g/dhcp4client v1.0.0
--- a/go.sum
+++ b/go.sum
@@ -8,10 +8,6 @@ github.com/buger/jsonparser v0.0.0-20180808090653-f4dd9f5a6b44 h1:y853v6rXx+zefE
 github.com/buger/jsonparser v0.0.0-20180808090653-f4dd9f5a6b44/go.mod h1:bbYlZJ7hK1yFx9hf58LP0zeX7UjIGs20ufpu3evjr+s=
 github.com/containernetworking/cni v0.8.0 h1:BT9lpgGoH4jw3lFC7Odz2prU5ruiYKcgAjMCbgybcKI=
 github.com/containernetworking/cni v0.8.0/go.mod h1:LGwApLUm2FpoOfxTDEeq8T9ipbpZ61X79hmU3w8FmsY=
-github.com/coreos/go-iptables v0.4.5 h1:DpHb9vJrZQEFMcVLFKAAGMUVX0XoRC0ptCthinRYm38=
-github.com/coreos/go-iptables v0.4.5/go.mod h1:/mVI274lEDI2ns62jHCDnCyBF9Iwsmekav8Dbxlm1MU=
-github.com/coreos/go-iptables v0.4.6-0.20200318170312-12696f5c9108 h1:H1IVw9/BmB880YZvqWEGuRVk/EUn4OukBjD7HW+zEYQ=
-github.com/coreos/go-iptables v0.4.6-0.20200318170312-12696f5c9108/go.mod h1:/mVI274lEDI2ns62jHCDnCyBF9Iwsmekav8Dbxlm1MU=
 github.com/coreos/go-iptables v0.5.0 h1:mw6SAibtHKZcNzAsOxjoHIG0gy5YFHhypWSSNc6EjbQ=
 github.com/coreos/go-iptables v0.5.0/go.mod h1:/mVI274lEDI2ns62jHCDnCyBF9Iwsmekav8Dbxlm1MU=
 github.com/coreos/go-systemd v0.0.0-20180511133405-39ca1b05acc7 h1:u9SHYsPQNyt5tgDm3YN7+9dYrpK96E5wFilTFWIDZOM=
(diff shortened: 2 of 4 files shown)


go.mod
```diff
@@ -8,7 +8,7 @@ require (
 	github.com/alexflint/go-filemutex v0.0.0-20171022225611-72bdc8eae2ae
 	github.com/buger/jsonparser v0.0.0-20180808090653-f4dd9f5a6b44
 	github.com/containernetworking/cni v0.8.0
-	github.com/coreos/go-iptables v0.4.5
+	github.com/coreos/go-iptables v0.4.6-0.20200318170312-12696f5c9108
```
mars1024, line 11: I think this means v0.4.6 hasn't been released, maybe we need help from @squeed

mars1024 (changes requested):

go.mod
tlwr: This is go mod pseudo-version syntax for a commit that has not been versioned yet.

This [pseudo-version guidance section](https://golang.org/ref/mod#pseudo-versions) describes the three valid bits of syntax:

> vX.Y.(Z+1)-0.yyyymmddhhmmss-abcdefabcdef is used when the base version is a release version like vX.Y.Z. For example, if the base version is v1.2.3, a pseudo-version might be v1.2.4-0.20191109021931-daa7c04131f5.

In this case, the base version is `v0.4.5` and so the +1 syntax is `v0.4.6` and the [commit hash is `12696f5c9108`](https://github.com/coreos/go-iptables/commit/12696f5c91089d9f7328d84aaf0a18120b546c3e) and the timestamp of the commit is `20200318170312`

I generated this `go.mod` file using `go get github.com/coreos/go-iptables@12696f5c9108` to get the commit I wanted, which [has already been done for `github.com/vishvananda/netlink` as an example](https://github.com/containernetworking/plugins/pull/553/commits/c41c78b6002add74363d9bc055fa4df801befa00)

mars1024: @squeed is the maintainer of project [https://github.com/coreos/go-iptables](https://github.com/coreos/go-iptables), I think he will help you ~

haircommander: go-iptables was bumped today https://github.com/coreos/go-iptables/releases/tag/v0.5.0

tlwr: I've amended the commit to vendor v0.5.0 instead of the pseudo-version

go.sum
```diff
@@ -10,6 +10,10 @@ github.com/containernetworking/cni v0.8.0 h1:BT9lpgGoH4jw3lFC7Odz2prU5ruiYKcgAjM
 github.com/containernetworking/cni v0.8.0/go.mod h1:LGwApLUm2FpoOfxTDEeq8T9ipbpZ61X79hmU3w8FmsY=
 github.com/coreos/go-iptables v0.4.5 h1:DpHb9vJrZQEFMcVLFKAAGMUVX0XoRC0ptCthinRYm38=
 github.com/coreos/go-iptables v0.4.5/go.mod h1:/mVI274lEDI2ns62jHCDnCyBF9Iwsmekav8Dbxlm1MU=
+github.com/coreos/go-iptables v0.4.6-0.20200318170312-12696f5c9108 h1:H1IVw9/BmB880YZvqWEGuRVk/EUn4OukBjD7HW+zEYQ=
+github.com/coreos/go-iptables v0.4.6-0.20200318170312-12696f5c9108/go.mod h1:/mVI274lEDI2ns62jHCDnCyBF9Iwsmekav8Dbxlm1MU=
```
bboreham, line 14: Nit: we don't need these two lines.  `go mod tidy` will likely take them out.

bboreham (approved):
Thanks!

go.sum
tlwr: Pushed a new commit with tidy, let me know if you want a squash

mars1024 (approved):
/lgtm, also thanks to @haircommander for the release

Issue #470, opened 2020-04-01
DHCP Timeout option


dougbtv:
I'm looking at a [bug report](https://bugzilla.redhat.com/show_bug.cgi?id=1818786) where someone was using DHCP over a linux bridge where they enabled spanning tree, and they're experiencing timeouts.

Currently, it looks like there's a 5 second timeout @ https://github.com/containernetworking/plugins/blob/master/plugins/ipam/dhcp/lease.go#L365

Would it be reasonable to create a new option for the DHCP daemon to have a `--timeout` option to optionally increase it from 5? Or would allowing a longer timeout be OK?

PR #565, opened 2021-01-11
DHCP timeout is configurable

tlwr:
What
----

Addresses #470, albeit quite naively. Building on the existing daemon configuration code

A user would deploy the daemon (using the systemd example):

```
[Service]
ExecStart=/opt/cni/bin/dhcp daemon -timeout 15s
```

to configure the DHCP client timeout to 15s. The default timeout remains 5s unchanged.

Why
---

If a user configures their DHCP daemon, because they are in an environment where DHCP responses may take a long time, an IP address will still be allocated for their container

Why not
---

It is arguable that a simpler solution would be to arbitrarily increase the default timeout to 15s. I can redo this PR to do that instead

Further work
------------

If this is merged I will make a cni.dev PR to document the CLI option, consistent with `hostprefix` and friends

### DIFF
--- a/plugins/ipam/dhcp/daemon.go
+++ b/plugins/ipam/dhcp/daemon.go
@@ -26,6 +26,7 @@ import (
 	"path/filepath"
 	"runtime"
 	"sync"
+	"time"
 
 	"github.com/containernetworking/cni/pkg/skel"
 	"github.com/containernetworking/cni/pkg/types"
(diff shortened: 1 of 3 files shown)


mars1024 (commented):
basically LGTM

mars1024:
one nit: consider defaulting client timeout to 10s just as [https://github.com/d2g/dhcp4client/blob/ef524ad9cb076fe235ecc3c55913ef8445fa35fa/client.go#L39](https://github.com/d2g/dhcp4client/blob/ef524ad9cb076fe235ecc3c55913ef8445fa35fa/client.go#L39)

tlwr:
Thanks for your review, I've co-authored a commit which changes the default to 10s

jellonek (approved):

squeed:
awesome, thanks!

PR #566, opened 2021-01-12
vlan: fix error message text by removing ptp references

crandles:
Fixing a few error messages that the vlan plugin returns. These appear to be mistaken references to the ptp plugin.


mars1024 (approved):
/lgtm, thanks!

PR #568, opened 2021-01-14
[main/vlan] Fix error handling for delegate IPAM plugin

jellonek:
At the moment error messages generated during execution of delegate IPAM plugin are indistinguishable from main `vlan` messages leading to cryptic messages like `no plugin name provided` which IS hard to debug from a human operator level.
This PR fixes that behavior.

### DIFF
--- a/plugins/main/vlan/vlan.go
+++ b/plugins/main/vlan/vlan.go
@@ -152,7 +152,7 @@ func cmdAdd(args *skel.CmdArgs) error {
 	// run the IPAM plugin and get back the config to apply
 	r, err := ipam.ExecAdd(n.IPAM.Type, args.StdinData)
 	if err != nil {
-		return err
+		return fmt.Errorf("failed to execute IPAM delegate: %v", err)
 	}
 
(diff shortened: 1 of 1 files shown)


plugins/main/vlan/vlan.go
```diff
@@ -152,7 +152,7 @@ func cmdAdd(args *skel.CmdArgs) error {
 	// run the IPAM plugin and get back the config to apply
 	r, err := ipam.ExecAdd(n.IPAM.Type, args.StdinData)
 	if err != nil {
-		return err
+		return fmt.Errorf("failed to execute IPAM delegate: %v", err)
```
hudymi, line 155: NIT: It is a good practice to use `%w` instead of `%v` when adding additional context to errors to keep the original error type.
https://golang.org/pkg/fmt/#Errorf

jellonek: hint: check rest of the convention used in the rest of the repo.
Fill welcomed to provide such change globally ;)

mars1024 (approved):
/lgtm

PR #569, opened 2021-01-14
[main/vlan] Behave like MACVLAN for L3 and gratuitous ARP

jellonek:
Both MACVLAN and VLAN plugins have similar functionality but are behaving differently.

This PR is providing behavior of MACVLAN plugin to VLAN plugin by adding a support for L2 only configuration (with IPAM skipped) and by adding gratuitous ARP messages for all IP assigned to configured VLAN interface.


squeed:
@jellonek looks like macvlan tests are failing; can you take a look?

jellonek:
@squeed with "little delay" - done.

dcbw:
This code should run LinkSetUp() in the L2-only case, like macvlan. That is no longer done now that ConfigureIface() isn't called.

jellonek:
And that means - I need to add also a test case for the state of interface at the end. Thx, good catch!

dcbw:
@jellonek can you update for the testcase and the LinkSetUp()?

gaby:
@jellonek Any plans to update this PR?

jellonek:
Yes. I'll have time for that during next week.

MikeZappa87:
@jellonek this PR might not be required with https://github.com/containernetworking/plugins/pull/687 ?

PR #579, opened 2021-02-01
host-local: fix mismatch startRange and startIP which cause host-local endless loop

gongguan:
### Bug report
In multi ip range scenario, when lastReservedIP is last ip of range i,
RangeIter will be init as `startRange = i, cur = lastReservedIP, starIP = nil`
when `RangeIter.Next()` called, `i.cur` will be changed as first ip of range i+1,
https://github.com/containernetworking/plugins/blob/a8d1f5cd1b132f10e5d86746ab82de1cd3c2b616/plugins/ipam/host-local/backend/allocator/allocator.go#L206-L211

then `startIP` will be changed too https://github.com/containernetworking/plugins/blob/a8d1f5cd1b132f10e5d86746ab82de1cd3c2b616/plugins/ipam/host-local/backend/allocator/allocator.go#L216-L217

but `iter.startRange` still equals i.
When no ip available, allocator Get() function can't jump out this [loop](https://github.com/containernetworking/plugins/blob/a8d1f5cd1b132f10e5d86746ab82de1cd3c2b616/plugins/ipam/host-local/backend/allocator/allocator.go#L91) from here https://github.com/containernetworking/plugins/blob/a8d1f5cd1b132f10e5d86746ab82de1cd3c2b616/plugins/ipam/host-local/backend/allocator/allocator.go#L218-L220

### How to fix
I have thought two solutions:
1. In `RangeIter.Next()` when startIP changed, change startRange at the same time, Ref #583 .
2. When lastReservedIP exists, make startIP as it's next ip in RangeIter init and get rid of some logic of nil startIP in `RangeIter.Next()`, Ref #579 .

What's your suggestion?

### DIFF
--- a/plugins/ipam/host-local/backend/allocator/allocator.go
+++ b/plugins/ipam/host-local/backend/allocator/allocator.go
@@ -164,16 +164,12 @@ func (a *IPAllocator) GetIter() (*RangeIter, error) {
 		startFromLastReservedIP = a.rangeset.Contains(lastReservedIP)
 	}
 
-	// Find the range in the set with this IP
+	// lastReservedIP exist, start with lastReservedIP's next ip.
 	if startFromLastReservedIP {
 		for i, r := range *a.rangeset {
 			if r.Contains(lastReservedIP) {
-				iter.rangeIdx = i
-				iter.startRange = i
-
-				// We advance the cursor on every Next(), so the first call
-				// to next() will return lastReservedIP + 1
-				iter.cur = lastReservedIP
+				iter.rangeIdx, iter.startIP = nextIPInRangeSet(*iter.rangeset, i, lastReservedIP)
+				iter.startRange = iter.rangeIdx
 				break
 			}
 		}
@@ -190,11 +186,9 @@ func (a *IPAllocator) GetIter() (*RangeIter, error) {
 func (i *RangeIter) Next() (*net.IPNet, net.IP) {
 	r := (*i.rangeset)[i.rangeIdx]
 
-	// If this is the first time iterating and we're not starting in the middle
-	// of the range, then start at rangeStart, which is inclusive
+	// i.cur equals nil means the first time iterating, start at startIP, which is inclusive.
 	if i.cur == nil {
-		i.cur = r.RangeStart
-		i.startIP = i.cur
+		i.cur = i.startIP
 		if i.cur.Equal(r.Gateway) {
 			return i.Next()
 		}
@@ -203,19 +197,10 @@ func (i *RangeIter) Next() (*net.IPNet, net.IP) {
 
 	// If we've reached the end of this range, we need to advance the range
 	// RangeEnd is inclusive as well
-	if i.cur.Equal(r.RangeEnd) {
-		i.rangeIdx += 1
-		i.rangeIdx %= len(*i.rangeset)
-		r = (*i.rangeset)[i.rangeIdx]
+	i.rangeIdx, i.cur = nextIPInRangeSet(*i.rangeset, i.rangeIdx, i.cur)
+	r = (*i.rangeset)[i.rangeIdx]
 
-		i.cur = r.RangeStart
-	} else {
-		i.cur = ip.NextIP(i.cur)
-	}
-
-	if i.startIP == nil {
-		i.startIP = i.cur
-	} else if i.rangeIdx == i.startRange && i.cur.Equal(i.startIP) {
+	if i.rangeIdx == i.startRange && i.cur.Equal(i.startIP) {
 		// IF we've looped back to where we started, give up
 		return nil, nil
 	}
@@ -225,4 +210,16 @@ func (i *RangeIter) Next() (*net.IPNet, net.IP) {
 	}
 
 	return &net.IPNet{IP: i.cur, Mask: r.Subnet.Mask}, r.Gateway
+}
+
+// nextIPInRangeSet returns curIP's next ip and its index in rangeSet.
+func nextIPInRangeSet(ranges RangeSet, idx int, curIP net.IP) (int, net.IP) {
+	var nextIP net.IP
+	if ranges[idx].RangeEnd.Equal(curIP) {
+		idx = (idx + 1) % len(ranges)
+		nextIP = ranges[idx].RangeStart
+	} else {
+		nextIP = ip.NextIP(curIP)
+	}
+	return idx, nextIP
 }
--- a/plugins/ipam/host-local/backend/allocator/allocator_test.go
+++ b/plugins/ipam/host-local/backend/allocator/allocator_test.go
@@ -49,6 +49,23 @@ func mkalloc() IPAllocator {
 	return alloc
 }
 
+func mkMultiRangeAlloc() IPAllocator {
+	p := RangeSet{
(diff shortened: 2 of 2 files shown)


gongguan:
Although it's a very corner case, but I think it's a heavy problem caused host-local cni hang(into endless loop). 
cc @squeed @dcbw

plugins/ipam/host-local/backend/allocator/allocator.go
```diff
@@ -203,10 +203,12 @@ func (i *RangeIter) Next() (*net.IPNet, net.IP) {
```
mars1024, line 203: Good catch! But maybe assigning `i.startIP` before nexting `i.cur` is better than introducing a new variable `nextRange`.

gongguan: Yes, I've thought it, but it will break current logic(some tests failed).

gongguan: Sorry for misread it, do you mean assigning i.startIP before https://github.com/containernetworking/plugins/blob/master/plugins/ipam/host-local/backend/allocator/allocator.go#L205  ?

gongguan: > Sorry for misreading it, do you mean assigning i.startIP before https://github.com/containernetworking/plugins/blob/master/plugins/ipam/host-local/backend/allocator/allocator.go#L205 ?

I have tried it, it will break current tests.

plugins/ipam/host-local/backend/allocator/allocator.go
```diff
@@ -215,6 +217,10 @@ func (i *RangeIter) Next() (*net.IPNet, net.IP) {
 
 	if i.startIP == nil {
 		i.startIP = i.cur
+		if nextRange {
+			i.startRange += 1
```
squeed, line 221: hmm, I don't think we should be changing startRange, as we use it to detect if we've looped around.

plugins/ipam/host-local/backend/allocator/allocator.go
squeed: Which tests will it break? If the tests rely on internal api, then they can be changed.

plugins/ipam/host-local/backend/allocator/allocator.go
gongguan: startIP become first ip in next range, but startRange still previous one, so host-local won't jump out from https://github.com/containernetworking/plugins/blob/master/plugins/ipam/host-local/backend/allocator/allocator.go#L91-L104

because `i.rangeIdx == i.startRange && i.cur.Equal(i.startIP)` never true
https://github.com/containernetworking/plugins/blob/master/plugins/ipam/host-local/backend/allocator/allocator.go#L218

gongguan: for example rangeSet: [{start: 10.0.0.1, end: 10.0.0.1}, {start: 10.0.0.3, end: 10.0.0.3}]
it can make startIP 10.0.0.3, startRange 0
however it should be startIP 10.0.0.3, startRange 1

plugins/ipam/host-local/backend/allocator/allocator.go
gongguan: @squeed Sorry, do you mean assigning i.startIP before nexting i.cur and fix there broken tests in allocator_test.go?

gongguan: > Which tests will it break? If the tests rely on internal api, then they can be changed.

@squeed Sorry, do you mean assigning i.startIP before nexting i.cur and fix there broken tests in allocator_test.go?

gongguan: @mars1024 @squeed If assigning `i.startIP` before nexting `i.cur`, that most recent reserved ip won't be traversed.

gongguan:
@mars1024 @squeed Addressed your comment and make `GetIter()` `Next()` behavior more consistent and cleaner, PTAL.

mars1024:
@gongguan @squeed  IMO, the root cause of this issue is that `startIP` and `startRange` are mismatching in some corner cases, for example, the last reserved IP is just at the end of one range when the allocator have several ranges at same time. This will cause an endless loop, which means we will never reach this [line](https://github.com/containernetworking/plugins/blob/8feef71fd3cb32ff9a6bf3a377837486872f86aa/plugins/ipam/host-local/backend/allocator/allocator.go#L218) because of this mismatch.

And I provider a better way for this, just always keep `startIP` and `startRange` matching, PTAL to [https://github.com/containernetworking/plugins/compare/bugfix/wrong_startrange](https://github.com/containernetworking/plugins/compare/bugfix/wrong_startrange) .

gongguan:
> @gongguan @squeed IMO, the root cause of this issue is that `startIP` and `startRange` are mismatching in some corner cases, for example, the last reserved IP is just at the end of one range when the allocator have several ranges at same time. This will cause an endless loop, which means we will never reach this [line](https://github.com/containernetworking/plugins/blob/8feef71fd3cb32ff9a6bf3a377837486872f86aa/plugins/ipam/host-local/backend/allocator/allocator.go#L218) because of this mismatch.
> 
> And I provider a better way for this, just always keep `startIP` and `startRange` matching, PTAL to https://github.com/containernetworking/plugins/compare/bugfix/wrong_startrange .

@mars1024 very clean fix, thanks! This corner case might be caused by inconsistent init between whether lastReserveIP exist or not, why not make it consistent? startIP is the first one to traverse(for lastReserveIP case, it should be its next ip), and cur is nil means first time iterating.
So logic in `Next()` should take care of many corner cases, such as nil startIP and nil cur. Consistent init in `GetIter()` might slim it, WDYT?

gongguan:
I don't have strong opinion with these solutions, it's important to fix it at first.

plugins/ipam/host-local/backend/allocator/allocator.go
gongguan: > hmm, I don't think we should be changing startRange, as we use it to detect if we've looped around.

@mars1024 I have thought @squeed don't want change startRange in `Next()` function, so I updated this pr.

bboreham:
I find the code quite hard to understand, both in #579 and in #583. 
I guess we can wait for @squeed to comment.

gongguan:
I have reworded pr description https://github.com/containernetworking/plugins/pull/579#issue-565171358, which is easier to understand.
I have met this problem in production environment and looking forward it can be fixed upstream. Can you give your feedback? @dcbw @bboreham @squeed

bboreham:
Thank you for all your work; we spent a lot of time going through the algorithm and your changes.

Current status: if #583 can be shown to be a correct solution then that is preferrable to #579, as it is less code.
There is a suggestion that it can be made simpler still, which we await an answer to.

bboreham:
Might be an idea to open an issue stating the problem, so that this PR and #583 can be marked as a solution to the issue.
Other people with the same problem are likely to search issues, less likely to search PRs.

gongguan:
> Might be an idea to open an issue stating the problem, so that this PR and #583 can be marked as a solution to the issue.
> Other people with the same problem are likely to search issues, less likely to search PRs.

Thanks for your suggestion, it's useful for users with the same problem, I have created #594 for it.

bboreham:
Replaced by #583, thanks for all your work!

PR #585, opened 2021-02-18
pkg/ipam: convert dots to slashes in interface names for sysctl

dverbeir:
Since the sysctl key syntax already uses the dot as separator, dots in interface names that appear in sysctl keys, as is the case for the net.ipv6.conf.<iface>.* entries, must be replaced by slashes.

### DIFF
--- a/pkg/ipam/ipam_linux.go
+++ b/pkg/ipam/ipam_linux.go
@@ -27,7 +27,8 @@ import (
 )
 
 const (
-	DisableIPv6SysctlTemplate = "net.ipv6.conf.%s.disable_ipv6"
+	// Note: use slash as separator so we can have dots in interface name (VLANs)
+	DisableIPv6SysctlTemplate = "net/ipv6/conf/%s/disable_ipv6"
 )
 
 // ConfigureIface takes the result of IPAM plugin and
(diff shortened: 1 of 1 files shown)


pkg/ipam/ipam_linux.go
```diff
@@ -64,12 +65,16 @@ func ConfigureIface(ifName string, res *current.Result) error {
 			// Enabled IPv6 for loopback "lo" and the interface
 			// being configured
 			for _, iface := range [2]string{"lo", ifName} {
-				ipv6SysctlValueName := fmt.Sprintf(DisableIPv6SysctlTemplate, iface)
+				// Cannot have dots in interface name for sysctl, must replace by /
+				ipv6SysctlValueName := fmt.Sprintf(DisableIPv6SysctlTemplate, strings.ReplaceAll(iface, ".", "/"))
```
mars1024, line 69: IMO, since dots and slashes can be interchanged in sysctl, is it better to make `DisableIPv6SysctlTemplate` to use slashes as separators?

jellonek: :+1: for leaving dots in interface names and instead making template slash separated.
Also: https://git.kernel.org/pub/scm/network/iproute2/iproute2.git/tree/lib/utils.c?id=1f420318bda3cc62156e89e1b56d60cc744b48ad#n827 - iproute2 is checking if interface does not have slash in the name.

jellonek (changes requested):

pkg/ipam/ipam_linux.go
jellonek: Also please take a look on https://github.com/containernetworking/cni/blob/master/pkg/utils/utils.go#L61

dverbeir: > Also please take a look on [our utils in cni lib](https://github.com/containernetworking/cni/blob/eefc06974c66eb2e14b06d0849aa82ebff43b837/pkg/utils/utils.go#L61).

Did you mean to suggest adding a check using utils.ValidateInterfaceName()? I believe such check is already performed in skel.pluginMain().

mars1024 (approved):
/lgtm, thanks!

squeed (approved):

PR #589, opened 2021-02-25
portmap: use slashes in sysctl template to support interface names which separated by dots

mars1024:
Scanned the whole project based on #585  and found this left behind.


jellonek (approved):
Big :+1:

bboreham (approved):

PR #590, opened 2021-02-26
win-bridge: fix panic while calling HNS api

thxCode:
fix a nil pointer panic while calling HNS API (V1) on win-bridge.


mars1024 (approved):
/lgtm, good catch!

jellonek (approved):
LGTM

PR #592, opened 2021-03-02
Add multi IP support for SBR

### LINKED ISSUE #581: SBR Query: How to chain SBR plugin with multiple IPAMs in same config, in order to use different routing tables for each IPAM block ?
Hi, today, using multus + sbr + static CNIs, in network-attchment-definition (NAD), under "plugins" section, we can add one IPAM section and chain it with SBR plugin to configure a separate routing table for it (table 100). Also we can add multiple IP addresses to same IPAM section ("static" plugin), by specifying the list of IPs in the IPAM section, under NAD.

Requirement : Is there a way to have multiple IPAM sections + SBR for each section in same NAD ?
We want to add multiple IPAM sections (each with its own default gw), in same NAD, and configure "SBR" plugin for each such IPAM. Thus, having multiple routing tables (just like "table 100" in pod), each with its own IPAM entries.

Thanks in Advance !!

anuragensemble:
[sbr]: Use different tableID for every ipCfg       
Move default table routes which  match the ipCfg config

This allows SBR plugin to accommodate for multi-ip interfaces

Fixes #581

Signed-off-by: Anurag Dwivedi <anuragensemble1@gmail.com>

### DIFF
--- a/plugins/meta/sbr/main.go
+++ b/plugins/meta/sbr/main.go
@@ -225,6 +225,12 @@ func doRoutes(ipCfgs []*current.IPConfig, origRoutes []*types.Route, iface strin
 
 	linkIndex := link.Attrs().Index
 
+	// Get all routes for the interface in the default routing table
+	routes, err = netlink.RouteList(link, netlink.FAMILY_ALL)
+	if err != nil {
+		return fmt.Errorf("Unable to list routes: %v", err)
+	}
+
 	// Loop through setting up source based rules and default routes.
 	for _, ipCfg := range ipCfgs {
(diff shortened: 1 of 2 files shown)


bboreham:
CI failed with a lot of `operation not supported` - is that expected?

dcbw:
@plwhite what do you think about this one?

anuragensemble:
@bboreham Thanks for the comment.
I failed to resolve the test failure as the failure is in the vrf plugin which was untouched.
Executing the test suite locally with and without the above changeset resulted in the same vrf test failure.

Will be very helpful if you can provide any inputs here.

PR #593, opened 2021-03-06
Fix nil-pointer check

thetechnick:
I am writing my own ipam plugin for a little project I am working on and stumbled across this nil check that looked out of place.
Omitting the `"ipam"` key from the config, while still supplying args or `runtimeConfig` would resulting in a nil pointer panic.

Signed-off-by: Nico Schieder <nschieder@redhat.com>

### DIFF
--- a/plugins/ipam/static/main.go
+++ b/plugins/ipam/static/main.go
@@ -143,6 +143,9 @@ func LoadIPAMConfig(bytes []byte, envArgs string) (*IPAMConfig, string, error) {
 	if err := json.Unmarshal(bytes, &n); err != nil {
 		return nil, "", err
 	}
+	if n.IPAM == nil {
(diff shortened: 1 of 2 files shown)


mars1024 (approved):
/lgtm

mars1024:
Ping @dcbw @bboreham , I think the CI has been breaking for a while since [this commit](https://github.com/containernetworking/plugins/commit/0ea07b826990e4475bbe06110f456ac0a648e6ee), is this caused by porting spec 1.0.0?

bboreham:
Thanks for the heads-up; as you note this started before the 1.0 spec change was merged.
It looks to me like the CI runner became unable to use a VRF, perhaps due to changes in underlying implementation.
Filed #596 to keep things separate.

bboreham (approved):
Thanks!

PR #598, opened 2021-03-10
win-bridge: fix panic while alling HCN api

thxCode:
the removing action swallows the failure of adding HostComputeEndpoint
to the given namespace, and then cause a nil pointer panic in the caller.


thxCode:
PTAL @mars1024

mars1024 (approved):
/lgtm, good catch, thanks!

thxCode:
closed in flavor of https://github.com/containernetworking/plugins/pull/617

PR #599, opened 2021-03-10
host-local: support custom IPs allocation through runtime configuraton

### LINKED ISSUE #411: Support 'ips' runtimeConfig with format <ip>[/<prefix>]

mars1024:
A new version of #416 , try to fix #411 

From the conclusion of [https://github.com/containernetworking/plugins/pull/416#discussion_r351387959](https://github.com/containernetworking/plugins/pull/416#discussion_r351387959), I've introduced a new type `IP` to support parsing format `<ip>[/<prefix>]`.
And I think this can also be used in `static` plugin.

### DIFF
--- a/pkg/ip/ip.go
+++ b/pkg/ip/ip.go
@@ -1,0 +1,105 @@
+// Copyright 2021 CNI authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


squeed:
lgtm.

Issue #419, opened 2019-12-08
bridge: Not able to use in a custom network namspace


alxbse:
I would like to be able to use the bridge plugin and have it create/use a bridge in a custom network namespace. Configuration wise I would like to do something like this:

```json
{
	"name": "mynet",
	"type": "bridge",
	"bridge": "mynet0",
        "namespace": "mynamespace",
	"ipam": {}
}
```

This would create the bridge in the network namespace `mynamespace` rather than the host network namespace.

Is this something that could be added to the bridge plugin or is it an intentional limitation?

squeed:
It's not a bad idea, though can you describe the use-case?

FWIW, you can do this now just by executing the CNI binary in the desired "host" netns - it doesn't explicitly discover it. Now, k8s won't let you do this, of course...

alxbse:
My use case is a kubernetes project where all container hosts have two network interfaces. One for host services and one intended for pod traffic. All routes the pods are intended to use are announced via BGP but since the host services and pods ends up using the same routing table host services tries to reach resources external to the cluster through routes intended only for the pods.

My use case can be solved by creating custom routing tables and rules but a more elegant solution I have experimented with is to have a dedicated namespace for the CNI bridge. Moving the second interface and the BGP solution into the dedicated namespace gives me a nice clean separation of traffic without any need for manually create routing rules.

Hopefully that gives a bit more info on my use case. I have a local branch of the bridge plugin with namespace support to verify the concept and would be happy to create a pull request from it. I just want to verify that it's not out of scope for this project or if there is maybe another simple way of solving it.

mars1024:
IMO, this should be the job of CRI or something like CNIManager to decide the "host" namespace.

alxbse:
@mars1024 By CNIManager you mean a solution like Calico or Weave?

mars1024:
I mean something like `cniNetworkPlugin` in `k8s.io/kubernetes/pkg/kubelet/dockershim/cni`.

Thoro:
I do have another use case, which cost me quite some time to figure out.

This only affects routed network solutions on kubernetes (kube-router), with firewalling on bridge level (custom), and NAT  ipvs (kube-router).

If a bridge is firewalled via iptables using net.bridge.bridge-nf-call-iptables = 1 then the POSTROUTING table is called only ONCE for the packets. This leads to an impossible state if the local pod is calling a local service in ipvs because the outgoing packets to other pods cannot be masqueraded anymore:

[...]

PR #608, opened 2021-03-23
introduce bridge namespace

Thoro:
Possible solution for #419 

Also added better error messages for the cmdAdd to figure out where it fails

### DIFF
--- a/plugins/main/bridge/bridge.go
+++ b/plugins/main/bridge/bridge.go
@@ -55,6 +55,7 @@ type NetConf struct {
 	HairpinMode  bool   `json:"hairpinMode"`
 	PromiscMode  bool   `json:"promiscMode"`
 	Vlan         int    `json:"vlan"`
+	BrNamespace  string `json:"bridgeNamespace"`
 }
 
 type gwInfo struct {
@@ -273,7 +274,7 @@ func ensureVlanInterface(br *netlink.Bridge, vlanId int) (netlink.Link, error) {
 			return nil, fmt.Errorf("faild to find host namespace: %v", err)
 		}
 
-		_, brGatewayIface, err := setupVeth(hostNS, br, name, br.MTU, false, vlanId)
+		_, brGatewayIface, err := setupVeth(hostNS, hostNS, br, name, br.MTU, false, vlanId)
 		if err != nil {
 			return nil, fmt.Errorf("faild to create vlan gateway %q: %v", name, err)
 		}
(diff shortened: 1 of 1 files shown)


plugins/main/bridge/bridge.go
```diff
@@ -474,32 +509,44 @@ func cmdAdd(args *skel.CmdArgs) error {
...
+			})
+
 			if err != nil {
 				return err
 			}
+
 			if hostVeth.Attrs().OperState == netlink.OperUp {
 				break
-			}
+			}
```
mars1024, line 538: unexpected blank spaces

plugins/main/bridge/bridge.go
```diff
@@ -273,7 +274,7 @@ func ensureVlanInterface(br *netlink.Bridge, vlanId int) (netlink.Link, error) {
 			return nil, fmt.Errorf("faild to find host namespace: %v", err)
 		}
 
-		_, brGatewayIface, err := setupVeth(hostNS, br, name, br.MTU, false, vlanId)
+		_, brGatewayIface, err := setupVeth(hostNS, hostNS, br, name, br.MTU, false, vlanId)
```
mars1024, line 277: why double input same parameters here?

plugins/main/bridge/bridge.go
```diff
@@ -395,20 +405,45 @@ func cmdAdd(args *skel.CmdArgs) error {
 		return fmt.Errorf("cannot set hairpin mode and promiscous mode at the same time.")
 	}
 
-	br, brInterface, err := setupBridge(n)
+	bridgeNS, err := ns.GetCurrentNS()
```
mars1024, line 408: why not just load bridge namespace here if non empty?

plugins/main/bridge/bridge.go
```diff
@@ -395,20 +405,45 @@ func cmdAdd(args *skel.CmdArgs) error {
...
 	if err != nil {
-		return err
+		return fmt.Errorf("failed to get current netns: %v", err)
+	}
+
+	defer bridgeNS.Close()
```
mars1024, line 413: duplicate close

mars1024:
I still think that this demand needs more discussions, because this maybe a general ability (specified host namespace) for more plugins, I'm not sure this implementation is suitable here. cc @containernetworking/cni-maintainers

plugins/main/bridge/bridge.go
Thoro: actually, this part is not done - didn't have the need for VLAN, I'll check it out

plugins/main/bridge/bridge.go
Thoro: why is it duplicate? should close the CurrentNS retrieved before.

But make more sense to just move it into the else

plugins/main/bridge/bridge.go
Thoro: Thx, screwed editor settings

plugins/main/bridge/bridge.go
Thoro: I wanted to have the same logic to follow up everywhere. So I need to populate bridgeNS to the current host network namespace, so I can work with it the same way.

Is my assumption wrong here?

plugins/main/bridge/bridge.go
Thoro: On further review I'm not sure what this option should do in combination with the bridge namespace.

Should the veth be moved on one side into the bridge namespace, and the other end on the host itself? 

If that's the case, this should also happen for the no-vlan case, since we need to connect the bridge to the host somehow. And then the whole IsGW code still needs adaption.

On the other hand if only the hostNS is in use, should a veth be used to connect to the bridge, or directly the bridge interface? 

The most probably solution would be the following, so to not change the current behavior:

if bridgeNamspace != "" -> create veth for VLAN and non-VLan and connect the hostNS with the bridgeNS
else -> create the veth if VLAN is necessary, otherwise use the bridge master interface directly

But this is very specific for my requirement, maybe someone wouldn't want the connection between the bridgeNS and the hostNS?

PR #611, opened 2021-03-25
Fix broken links to online docs in plugin READMEs

SilverSoldier:
Fixes issue #610

PR #612, opened 2021-04-07
bridge: update a URL in plugin README.md file

masatake:
The original URL shows "Page Not Found."


mars1024 (commented):
duplicate with #611

PR #613, opened 2021-04-07
win-bridge/win-overlay: skip dsr if not supported

thxCode:
just skip the DSR configuration(`loopbackDSR: true`) if the host is not supported.

### DIFF
--- a/pkg/hns/netconf_windows.go
+++ b/pkg/hns/netconf_windows.go
@@ -73,6 +73,9 @@ func GetDefaultDestinationPrefix(ip *net.IP) string {
 }
 
 func (n *NetConf) ApplyLoopbackDSR(ip *net.IP) {
+	if err := hcn.DSRSupported(); err != nil {
+		return
+	}
 	value := fmt.Sprintf(`"Destinations" : ["%s"]`, ip.String())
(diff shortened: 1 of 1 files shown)


bboreham:
For my information, what is "DSR" please?

thxCode:
@bboreham  DSR means Direct Server Return (DSR), it is a feature introduced in [KB4571748(10.0.17763.1432)](https://support.microsoft.com/en-us/topic/august-20-2020-kb4571748-os-build-17763-1432-preview-fa1db909-8923-e70f-9aef-ba09edaee6f0). #331 allows for loopback DSR, but if we configure `loopbackDSR: true` in the host which is lower than 10.0.17763.143, the CNI plugin(win-bridge/win-overlay) will exclude all valid port(`netsh int ipv4 sh excludedportrange TCP`), finally this action will cause an error as below in `kube-proxy`.
```
Policy creation failed: hcnCreateLoadBalancer failed in Win32: Not enough memory resources are available to complete this operation. (0xe) {"Success":false,"Error":"Not enough memory resources are available to complete this operation. ","ErrorCode":2147942414
```

thxCode:
closed in flavor of https://github.com/containernetworking/plugins/pull/617

Issue #601, opened 2021-03-10
Need support for multiple routes to the same prefix


mccv1r0:
Currently if a user tries to add e.g. multiple default routes (but not limited to default routes), only the first is added, the rest are silently ignored.

This causes problem when a container has multiple interfaces.  If the interface which added the route fails, or is dynamically removed, the container is left with no default route.

PR #602, opened 2021-03-11
Allow multiple routes to be added for the same prefix.

mccv1r0:
Allow multiple routes to be added for the same prefix.
Enables ECMP

Fixes: https://github.com/containernetworking/plugins/issues/601

Signed-off-by: Michael Cambria <mcambria@redhat.com>

PR #615, opened 2021-04-14
Allow multiple routes to be added for the same prefix

mccv1r0:
Allow multiple routes to be added for the same prefix.
Enables ECMP

Fixes: #601


mccv1r0:
Formally https://github.com/containernetworking/plugins/pull/602 which entered rebase/revendor hell

plugins/main/bridge/bridge.go
```diff
@@ -122,7 +122,7 @@ func calcGateways(result *current.Result, n *NetConf) (*gwInfo, *gwInfo, error)
 
 		// Add a default route for this family using the current
 		// gateway address if necessary.
-		if n.IsDefaultGW && !gws.defaultRouteFound {
+		if n.IsDefaultGW {
```
dcbw, line 125: @mccv1r0 is this to allow multiple default routes for the bridge or something?

mccv1r0: Yes, but keeping it seems to behave too, so I added `&& !gws.defaultRouteFound` back.

PR #617, opened 2021-04-19
refactor(windows): win-bridge

thxCode:
There are some problems with `win-bridge` plugin at present, i.e: unclear error message, unclear function name, cannot apply v2 API in policies...

This PR is going to clear/group the function name, comments, and some error messages through two `chore` commits. Then, through two `refactor` commits, some adjustments were made to the process of creating and deleting endpoints, so that the API processing logic of both versions is almost the same. Finally, through a breaking `refactor` commit to unify the two versions of the policy applying processing.

I think the impact of this final breaking commit is very small as the v2 API lacks the use under dockershim. Merging `HcnPolicyArgs` into `Policies` field to unify the endpoint policy configuration in one place. We still keep the same as before under dockershim, but change as below when switching to containerd.

```diff
{
    "name":"cbr0",
    "type":"flannel",
    "delegate":{
+       "apiVersion":2,
        "type":"win-bridge",
        "dns":{
            "nameservers":[
                "11.0.0.10"
            ],
            "search":[
                "svc.cluster.local"
            ]
        },
        "policies":[
            {
                "name":"EndpointPolicy",
                "value":{
                    "Type":"OutBoundNAT",
-                   "ExceptionList":[
-                       "192.168.0.0/16",
-                       "11.0.0.0/8",
-                       "10.137.196.0/23"
-                   ]
+                   "Settings":{
+                      "Exceptions":[
+                           "192.168.0.0/16",
+                           "11.0.0.0/8",
+                           "10.137.196.0/23"
+                       ]
+                   }
                }
            },
            {
                "name":"EndpointPolicy",
                "value":{
-                   "Type":"ROUTE",
-                   "DestinationPrefix":"11.0.0.0/8",
-                   "NeedEncap":true
+                   "Type":"SDNRoute",
+                   "Settings":{
+                      "DestinationPrefix":"11.0.0.0/8",
+                      "NeedEncap":true
+                   }
                }
            },
            {
                "name":"EndpointPolicy",
                "value":{
-                   "Type":"ROUTE",
-                   "DestinationPrefix":"10.137.198.27/32",
-                   "NeedEncap":true
+                   "Type":"SDNRoute",
+                   "Settings":{
+                      "DestinationPrefix":"10.137.198.27/32",
+                      "NeedEncap":true
+                   }
                }
            }
        ],
        "loopbackDSR": true
    }
}
```


thxCode:
@nagiesek PTAL cc @mars1024 , @bboreham

nagiesek:
looks okay on a quick glance, but let run through this more in depth

nagiesek:
@vikas-bh

dcbw:
@vikas-bh @nagiesek any updates here?

dcbw:
@vikas-bh @nagiesek ping again for review, thanks!

pkg/hns/endpoint_windows.go
```diff
@@ -61,222 +62,136 @@ func GetIpString(ip *net.IP) string {
 	}
 }
 
+// GetDefaultDestinationPrefix returns the default destination prefix according to the given IP type.
+func GetDefaultDestinationPrefix(ip *net.IP) string {
+	destinationPrefix := "0.0.0.0/0"
+	if ipv6 := ip.To4(); ipv6 == nil {
```
dcbw, line 68: @thxCode how about `if ip.To4() == nil {` ?

jayunit100:
hi ! im gonna try to get some context for this ....

pkg/hns/endpoint_windows.go
thxCode: @dcbw, updated

jayunit100:
@daschott @jsturtevant 
- Can we assume this is safe since its a v2 upgrade ?
- Does the V2 golang client "know" how to translate to V1 API if needed?

plugins/main/windows/win-bridge/win-bridge_windows.go
```diff
@@ -98,89 +98,82 @@ func cmdHnsAdd(args *skel.CmdArgs, n *NetConf) (*current.Result, error) {
...
+		return nil, errors.Annotate(err, "error while constructing HostComputeEndpoint addition result")
 	}
-
 	return result, nil
 }
 
 func cmdAdd(args *skel.CmdArgs) error {
 	n, cniVersion, err := loadNetConf(args.StdinData)
 	if err != nil {
-		return errors.Annotate(err, "error while loadNetConf")
```
anfernee, line 183: nitpick: `error while loading NetConf` to be consistent.

anfernee (commented):
From my limited knowledge of CNI implementation on windows, it looks good. I don't know how do we verify the netconf part with any sort of integration test? It depends on HCN/HNS interfaces on Windows.

pkg/hns/netconf_windows.go
```diff
@@ -54,42 +55,74 @@ type PortMapEntry struct {
 	HostIP        string `json:"hostIP,omitempty"`
 }
 
+func (p *PortMapEntry) GetProtocolEnum() (uint32, error) {
```
nagiesek, line 58: Type defining the Protocols as uint32 and giving a list of constants might be preferred

pkg/hns/netconf_windows.go
```diff
@@ -101,136 +134,222 @@ func (n *NetConf) GetDNS() types.DNS {
...
+
+	// find OutBoundNAT policy
+	for i := range n.Policies {
+		p := &n.Policies[i]
 		if !strings.EqualFold(p.Name, "EndpointPolicy") {
 			continue
 		}
 
-		result = append(result, p.Value)
+		// filter OutBoundNAT policy
```
nagiesek, line 158: You're filtering here to see if LoopbackDSR is already added by outbound nat policies? Add a comment clarifying I think

pkg/hns/netconf_windows.go
```diff
@@ -101,136 +134,222 @@ func (n *NetConf) GetDNS() types.DNS {
...
-		typeValue, err := jsonparser.GetUnsafeString(p.Value, "Type")
-		if err != nil || len(typeValue) == 0 {
+		// filter OutBoundNAT policy
+		typeValue, _ := jsonparser.GetUnsafeString(p.Value, "Type")
+		if typeValue != "OutBoundNAT" {
 			continue
 		}
 
-		if !strings.EqualFold(typeValue, "OutBoundNAT") {
+		// parse exception CIRD list
```
nagiesek, line 230: Typo: CIRD -> CIDR

nagiesek (approved):
Large change moving a lot of stuff around. Little hard to follow but nothing jumps out so LGTM.  Some minor comments/clarifications could help.

pkg/hns/endpoint_windows.go
```diff
@@ -52,7 +53,7 @@ func GetSandboxContainerID(containerID string, netNs string) string {
 	return containerID
 }
 
-// short function so we know when to return "" for a string
+// GetIpString returns the given IP in string.
```
dcantah, line 56: Nit: "as a string"

pkg/hns/endpoint_windows.go
```diff
@@ -61,222 +62,136 @@ func GetIpString(ip *net.IP) string {
...
-		if hnsEndpoint.VirtualNetwork != epInfo.NetworkId {
-			_, err = hnsEndpoint.Delete()
-			if err != nil {
-				return nil, errors.Annotatef(err, "failed to delete endpoint %s", epInfo.EndpointName)
-			}
-			hnsEndpoint = nil
+		if strings.EqualFold(hnsEndpoint.VirtualNetwork, epInfo.NetworkId) {
+			return nil, fmt.Errorf("HNSEndpoint %s is already existed", epInfo.EndpointName)
+		}
+		// remove endpoint if corrupted
```
dcantah, line 91: What constitutes the endpoint being corrupted in this case?

thxCode: I think the endpoint is corrupted if it has the same name but mounted on the other network.

plugins/main/windows/win-bridge/win-bridge_windows.go
thxCode: I think the error returned by the `loadNetConf` is clear.
https://github.com/containernetworking/plugins/blob/78702e9d8a2cdb6931fab433f1d1e6789162954f/plugins/main/windows/win-bridge/win-bridge_windows.go#L50-L56

pkg/hns/endpoint_windows.go
```diff
@@ -296,7 +211,7 @@ func ConstructResult(hnsNetwork *hcsshim.HNSNetwork, hnsEndpoint *hcsshim.HNSEnd
 		CNIVersion: current.ImplementedSpecVersion,
 		Interfaces: []*current.Interface{resultInterface},
 		IPs:        []*current.IPConfig{resultIPConfig},
-		DNS:        types.DNS{
+		DNS: types.DNS{
```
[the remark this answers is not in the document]
thxCode: This is the result of gofmt.

pkg/hns/endpoint_windows.go
```diff
@@ -61,222 +62,136 @@ func GetIpString(ip *net.IP) string {
...
-			err := DeprovisionEndpoint(epName, netns, containerID)
-			if err != nil {
-				return nil, errors.Annotatef(err, "failed to Deprovsion after HotAttach failure")
+		if isNewEndpoint {
+			if err := RemoveHnsEndpoint(epName, netns, containerID); err != nil {
+				return nil, errors.Annotatef(err, "failed to remove the new HNSEndpoint %s after attaching container %s failure", hnsEndpoint.Id, containerID)
 			}
-		}
-		if hcsshim.ErrComputeSystemDoesNotExist == err {
+		} else if hcsshim.ErrComputeSystemDoesNotExist == err {
```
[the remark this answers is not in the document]
thxCode: I don't remember the details very much. The impression is that when dockershim starts the non-pause containers in the Pod, those containers may still be created instead of running.

pkg/hns/endpoint_windows.go
```diff
@@ -61,222 +62,136 @@ func GetIpString(ip *net.IP) string {
...
 
-	// check if endpoint already exists
-	createEndpoint := true
-	hnsEndpoint, err := hcsshim.GetHNSEndpointByName(epName)
-	if hnsEndpoint != nil && strings.EqualFold(hnsEndpoint.VirtualNetwork, expectedNetworkId) {
-		createEndpoint = false
+	// for shared endpoint, we expect that the endpoint already exists
+	if netns != pauseContainerNetNS {
+		if hnsEndpoint == nil {
+			return nil, errors.Annotatef(err, "failed to find HNSEndpoint %s", epName)
```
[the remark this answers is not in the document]
thxCode: here keeps the same error message as before this refactoring, I think we don't need to change it.

pkg/hns/endpoint_windows.go
thxCode: updated

pkg/hns/netconf_windows.go
thxCode: Yes, to confirm whether we have configured the same CIDR in the exception list of the outbound NAT policies. Just to not repeat the configuration.

pkg/hns/netconf_windows.go
thxCode: updated

pkg/hns/netconf_windows.go
thxCode: updated

dcbw:
Thanks for all the reviews and work.

PR #624, opened 2021-04-28
tuning: Add support of altering the allmulticast flag

### LINKED ISSUE #621: tuning: Support changing the allmulticast flag on interface
This is a proposal to add functionality to the meta plugin `tuning` to be able to change the all-multicast flag (`IFF_ALLMULTI`) on the interface. This will enable propagation of all multicast packets from the network.

A usecase example is when using IPv6 and having Neighbor Discovery Protocol (NDP) implemented in a Network Functions (NF) userland stack. The NF then needs to be able to receive multicast Ethernet packets which requires setting the flag `IFF_ALLMULTI` on the device to receive all multicast packets.

The NF should not set this flag itself because it requires `CAP_NET_ADMIN` capability which normally is considered too wide and insecure. Having the CNI do the scaffolding would be more proper.

This has also been discussed in k8snetworkplumbingwg/sriov-cni#175 and the proposal there is to have this functionality in the meta plugin `tuning`, which supports modifying the promisc flag already today.

bjosv:
This adds support to allow the tuning plugin to enable/disable the allmulticast mode of the interface.
When enabled it instructs the network driver to retrieve all multicast packets from the network.

Resolves #621

### DIFF
--- a/plugins/meta/tuning/tuning.go
+++ b/plugins/meta/tuning/tuning.go
@@ -29,6 +29,7 @@ import (
 
 	"github.com/j-keck/arping"
 	"github.com/vishvananda/netlink"
+	"golang.org/x/sys/unix"
 
 	"github.com/containernetworking/cni/pkg/skel"
 	"github.com/containernetworking/cni/pkg/types"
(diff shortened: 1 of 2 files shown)


bjosv:
If the PR vishvananda/netlink#635 is merged the import of sys/unix can be avoided, but since the tuning plugin is not amongst the Windows supported plugins (i.e not in `windows_only.txt`) it was assumed to be ok for now.

squeed:
@bjosv this looks good! Can you also file a PR at https://github.com/containernetworking/cni.dev/blob/main/content/plugins/current/meta/tuning.md adding this flag please?

bjosv:
> @bjosv this looks good! Can you also file a PR at https://github.com/containernetworking/cni.dev/blob/main/content/plugins/current/meta/tuning.md adding this flag please?

@squeed Great! Filed the corresponding documentation change in containernetworking/cni.dev#84

PR #636, opened 2021-06-10
bridge: Add mac field to specify container iface mac

EdDev:
Controlling the mac address of the interface (veth peer) in the
container is useful for functionalities that depend on the mac address.
Examples range from dynamic IP allocations based on an identifier (the
mac) and up to firewall rules (e.g. no-mac-spoofing).

Enforcing a mac address at an early stage and not through a chained
plugin assures the configuration does not have wrong intermediate
configuration. This is especially critical when a dynamic IP may be
provided already in this period.
But it also has implications for future abilities that may land on the
bridge plugin, e.g. supporting no-mac-spoofing.

The field name used (`mac`) fits with other plugins which control the
mac address of the container interface.

The mac address may be specified through the following methods:
- CNI_ARGS
- Args
- RuntimeConfig [1]

The list is ordered by priority, from lowest to higher. The higher
priority method overrides any previous settings.
(e.g. if the mac is specified in RuntimeConfig, it will override any
specifications of the mac mentioned in CNI_ARGS or Args)

[1] To use RuntimeConfig, the network configuration should include the
`capabilities` field with `mac` specified (`"capabilities": {"mac": true}`).

### DIFF
--- a/pkg/ip/link_linux_test.go
+++ b/pkg/ip/link_linux_test.go
@@ -72,7 +72,7 @@ var _ = Describe("Link", func() {
 		_ = containerNetNS.Do(func(ns.NetNS) error {
 			defer GinkgoRecover()
 
-			hostVeth, containerVeth, err = ip.SetupVeth(fmt.Sprintf(ifaceFormatString, ifaceCounter), mtu, hostNetNS)
+			hostVeth, containerVeth, err = ip.SetupVeth(fmt.Sprintf(ifaceFormatString, ifaceCounter), mtu, "", hostNetNS)
 			if err != nil {
 				return err
 			}
@@ -159,7 +159,7 @@ var _ = Describe("Link", func() {
 			_ = containerNetNS.Do(func(ns.NetNS) error {
 				defer GinkgoRecover()
 
-				_, _, err := ip.SetupVeth(containerVethName, mtu, hostNetNS)
+				_, _, err := ip.SetupVeth(containerVethName, mtu, "", hostNetNS)
 				Expect(err.Error()).To(Equal(fmt.Sprintf("container veth name provided (%s) already exists", containerVethName)))
 
 				return nil
@@ -189,7 +189,7 @@ var _ = Describe("Link", func() {
 		It("returns useful error", func() {
 			_ = containerNetNS.Do(func(ns.NetNS) error {
 				defer GinkgoRecover()
-				_, _, err := ip.SetupVeth(containerVethName, mtu, hostNetNS)
+				_, _, err := ip.SetupVeth(containerVethName, mtu, "", hostNetNS)
 				Expect(err.Error()).To(HavePrefix("failed to move veth to host netns: "))
 
 				return nil
@@ -207,7 +207,7 @@ var _ = Describe("Link", func() {
 			_ = containerNetNS.Do(func(ns.NetNS) error {
 				defer GinkgoRecover()
 
-				hostVeth, _, err := ip.SetupVeth(containerVethName, mtu, hostNetNS)
+				hostVeth, _, err := ip.SetupVeth(containerVethName, mtu, "", hostNetNS)
 				Expect(err).NotTo(HaveOccurred())
 				hostVethName = hostVeth.Name
 				return nil
@@ -233,6 +233,32 @@ var _ = Describe("Link", func() {
 			})
 		})
 
+		It("successfully creates a veth pair with an explicit mac", func() {
+			const mac = "02:00:00:00:01:23"
+			_ = containerNetNS.Do(func(ns.NetNS) error {
+				defer GinkgoRecover()
+
+				hostVeth, _, err := ip.SetupVeth(containerVethName, mtu, mac, hostNetNS)
+				Expect(err).NotTo(HaveOccurred())
+				hostVethName = hostVeth.Name
+
+				link, err := netlink.LinkByName(containerVethName)
+				Expect(err).NotTo(HaveOccurred())
+				Expect(link.Attrs().HardwareAddr.String()).To(Equal(mac))
+
+				return nil
+			})
+
+			_ = hostNetNS.Do(func(ns.NetNS) error {
+				defer GinkgoRecover()
+
+				link, err := netlink.LinkByName(hostVethName)
+				Expect(err).NotTo(HaveOccurred())
+				Expect(link.Attrs().HardwareAddr.String()).NotTo(Equal(mac))
+
+				return nil
+			})
+		})
 	})
 
 	It("DelLinkByName must delete the veth endpoints", func() {
--- a/plugins/main/bridge/bridge.go
+++ b/plugins/main/bridge/bridge.go
@@ -55,6 +55,25 @@ type NetConf struct {
 	HairpinMode  bool   `json:"hairpinMode"`
 	PromiscMode  bool   `json:"promiscMode"`
 	Vlan         int    `json:"vlan"`
+
+	Args struct {
+		Cni BridgeArgs `json:"cni,omitempty"`
+	} `json:"args,omitempty"`
+	RuntimeConfig struct {
+		Mac string `json:"mac,omitempty"`
+	} `json:"runtimeConfig,omitempty"`
+
+	mac string
+}
+
+type BridgeArgs struct {
+	Mac string `json:"mac,omitempty"`
+}
+
+// MacEnvArgs represents CNI_ARGS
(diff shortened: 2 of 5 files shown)


EdDev:
@squeed I have added the support for CNI_ARGS, Args and RunningConfig parameters.

Can you please approve the CI run?

mccv1r0:
I believe that this would fix https://github.com/containernetworking/plugins/issues/450 as well.

squeed:
This looks pretty good. @s1061123 would you mind giving this a quick review?

squeed:
Personal nit pick - can this be a single commit?

EdDev:
> Personal nit pick - can this be a single commit?

The idea was to make it easy for reviewers to go over it in stages. The changes are independent, stand by themselves and give a nicer "story" in the history (the details in the tests may look messy when looked as one change).
But I do not have any strong feelings over this, if you still prefer to squash it all, just let me know.

pkg/ip/link_linux_test.go
```diff
@@ -233,6 +233,37 @@ var _ = Describe("Link", func() {
 			})
 		})
 
+		It("successfully creates a veth pair with an explicit mac", func() {
+			const mac = "02:00:00:00:01:23"
```
mars1024, line 237: why this test case is separated into three steps? and I don't see the necessity of the third step

plugins/main/bridge/bridge.go
```diff
@@ -55,6 +55,24 @@ type NetConf struct {
...
 	PromiscMode  bool   `json:"promiscMode"`
 	Vlan         int    `json:"vlan"`
+	Mac          string `json:"mac,omitempty"`
+
+	Args struct {
+		Cni BridgeArgs `json:"cni,omitempty"`
+	} `json:"args,omitempty"`
+	RuntimeConfig struct {
+		Mac string `json:"mac,omitempty"`
+	} `json:"runtimeConfig,omit"`
```
mars1024, line 65: nit: omit --> omitempty

plugins/main/bridge/bridge.go
```diff
@@ -55,6 +55,24 @@ type NetConf struct {
 	HairpinMode  bool   `json:"hairpinMode"`
 	PromiscMode  bool   `json:"promiscMode"`
 	Vlan         int    `json:"vlan"`
+	Mac          string `json:"mac,omitempty"`
```
mars1024, line 58: `Mac` field here is internal, it should be ignored when marshalling

EdDev:
change: [Responding to review](https://github.com/containernetworking/plugins/compare/d1bcb404a670c5ba9adfa76f2c5b9628e1abbcb6..089b7d3673207c68782fc48064a6d118ccf75e94).

s1061123:
As far as I looked the diff, the 'mac' field could be provided through following ways:
 - CNI config
 - RuntimeConfig
 - 'args'

So if user fills different MAC address in those fields, which MAC will wins? 
You may need to add some comments (and document as well).

EdDev:
> As far as I looked the diff, the 'mac' field could be provided through following ways:
> 
>     * CNI config
> 
>     * RuntimeConfig
> 
>     * 'args'
> 
> 
> So if user fills different MAC address in those fields, which MAC will wins?
> You may need to add some comments (and document as well).

I think the order is suppose to be part of the [convention](https://github.com/containernetworking/cni/blob/master/CONVENTIONS).
`RuntimeConfig` overrides `args` which overrides `CNI_ARGS` which overrides `CNI config`.
The plugins which support all, use this convention.

There is an explicit statement  where `args` override `CNI_ARGS` [here](https://github.com/containernetworking/cni/blob/master/CONVENTIONS.md#cni_args).

@squeed , am I correct here? Is there another specification where this is defined? Do you think it is worth updating the document to clarify this?

s1061123:
In convention.md, it seems not to be mentioned clearly (explicitly) yet, I guess.

@squeed Do you have any plan to mention the priority among Args/CNI_ARGS/RuntimeConfig in spec?

plugins/main/bridge/bridge.go
dcbw: @EdDev this is the netconf for the overall Bridge, not the container interface. A 'mac' field here would be the MAC of the bridge itself, not the container veth. It shouldn't be here because per-container options are not specified int he top-level plugin config struct. That's what the runtime args and CNI_ARGS are for.

EdDev:
Change: Answered review requests
- Squashed all commits into one.
- Removed the ability to set the mac address from the plugin network config (netconf).

plugins/main/bridge/bridge.go
EdDev: Done

plugins/main/bridge/bridge_test.go
```diff
@@ -1967,6 +2010,66 @@ var _ = Describe("bridge Operations", func() {
 			})
 		}
 
+		It(fmt.Sprintf("[%s] uses an explicit MAC addresses for the container iface (from CNI_ARGS)", ver), func() {
```
alonSadan, line 2013: are all macs acceptable for a veth? if no perhaps it is also worth to add a test that shows what suppose to happen if someone sets a "wrong" mac?

alonSadan (commented):
thank you Eddi. Really Nice!

I am not so familiar with this subject, but (as I asked also in the inline comments) are all macs acceptable for a veth at this stage, or there are limitations ? what happens if the mac being set breaks the limitation?

For example I think that the bridges mac should be the smallest of all macs of the veths that are connected to this bridge. Do we care about this kind of limitation at this stage, and is there something should be done to enforce it ?

plugins/main/bridge/bridge.go
```diff
@@ -55,6 +55,25 @@ type NetConf struct {
...
+
+	mac string
+}
+
+type BridgeArgs struct {
+	Mac string `json:"mac,omitempty"`
+}
+
+// MacEnvArgs represents CNI_ARGS
+type MacEnvArgs struct {
```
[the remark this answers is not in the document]
squeed: No, this is the type boilerplate to automatically parse CNI_ARGS "MAC=FOO" strings.

plugins/main/bridge/bridge_test.go
```diff
@@ -78,6 +78,21 @@ type testCase struct {
 	DelErr020  string
 	AddErr010  string
 	DelErr010  string
+
+	envArgs       string // CNI_ARGS
```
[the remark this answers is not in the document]
dcbw: Let's keep it envArgs to be consistent with other plugins. Also we don't want to export this (although it's a testcase so it wouldn't matter, but still)

plugins/main/bridge/bridge.go
```diff
@@ -80,6 +99,26 @@ func loadNetConf(bytes []byte) (*NetConf, string, error) {
...
 		return nil, "", fmt.Errorf("invalid VLAN ID %d (must be between 0 and 4094)", n.Vlan)
 	}
+
+	if envArgs != "" {
+		e := MacEnvArgs{}
+		if err := types.LoadArgs(envArgs, &e); err != nil {
+			return nil, "", err
+		}
+
+		if e.MAC != "" {
```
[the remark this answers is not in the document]
dcbw: No objection, but I'm also OK with leaving it as-is.

plugins/main/bridge/bridge_test.go
dcbw: Yeah, they are all "valid" here. The kernel will validate and fail the request if the MAC is wrong in some way (all zeros or 1s) and we shouldn't bother trying to reproduce the validation the kernel does I think.

squeed:
@EdDev now that this has merged, can you file a PR on https://github.com/containernetworking/cni.dev/ that updates the bridge documentation (https://github.com/containernetworking/cni.dev/blob/main/content/plugins/current/main/bridge.md)

EdDev:
> @EdDev now that this has merged, can you file a PR on https://github.com/containernetworking/cni.dev/ that updates the bridge documentation (https://github.com/containernetworking/cni.dev/blob/main/content/plugins/current/main/bridge.md)

On it first thing tomorrow morning.
Thank you!

EdDev:
> > @EdDev now that this has merged, can you file a PR on https://github.com/containernetworking/cni.dev/ that updates the bridge documentation (https://github.com/containernetworking/cni.dev/blob/main/content/plugins/current/main/bridge.md)
> 
> On it first thing tomorrow morning.
> Thank you!

Done

PR #638, opened 2021-06-22
Fix confusing error msg invalid cidr

maiqueb:
By wrapping the error msg we ended up actually duplicating the exact same msg `net.ParseCIDR` replies, as seen in its implementation - [0] & [1]

This PR makes the error msg less confusing.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1897431

[0] - https://github.com/golang/go/blob/2ebe77a2fda1ee9ff6fd9a3e08933ad1ebaea039/src/net/net.go#L540
[1] - https://github.com/golang/go/blob/2ebe77a2fda1ee9ff6fd9a3e08933ad1ebaea039/src/net/ip.go#L739

### DIFF
--- a/plugins/ipam/static/main.go
+++ b/plugins/ipam/static/main.go
@@ -193,6 +193,10 @@ func LoadIPAMConfig(bytes []byte, envArgs string) (*IPAMConfig, string, error) {
 		// args IP overwrites IP, so clear IPAM Config
 		n.IPAM.Addresses = make([]Address, 0, len(n.Args.A.IPs))
 		for _, addr := range n.Args.A.IPs {
+			_, _, err := net.ParseCIDR(addr)
+			if err != nil {
+				return nil, "", fmt.Errorf("an entry in the 'ips' field is NOT in CIDR notation, got: '%s'", addr)
+			}
 			n.IPAM.Addresses = append(n.IPAM.Addresses, Address{AddressStr: addr})
 		}
 	}
@@ -202,6 +206,10 @@ func LoadIPAMConfig(bytes []byte, envArgs string) (*IPAMConfig, string, error) {
 		// runtimeConfig IP overwrites IP, so clear IPAM Config
 		n.IPAM.Addresses = make([]Address, 0, len(n.RuntimeConfig.IPs))
 		for _, addr := range n.RuntimeConfig.IPs {
+			_, _, err := net.ParseCIDR(addr)
+			if err != nil {
+				return nil, "", fmt.Errorf("an entry in the 'ips' field is NOT in CIDR notation, got: '%s'", addr)
+			}
 			n.IPAM.Addresses = append(n.IPAM.Addresses, Address{AddressStr: addr})
 		}
 	}
--- a/plugins/ipam/static/main.go
+++ b/plugins/ipam/static/main.go
@@ -192,12 +192,13 @@ func LoadIPAMConfig(bytes []byte, envArgs string) (*IPAMConfig, string, error) {
 	if n.Args != nil && n.Args.A != nil && len(n.Args.A.IPs) != 0 {
 		// args IP overwrites IP, so clear IPAM Config
 		n.IPAM.Addresses = make([]Address, 0, len(n.Args.A.IPs))
-		for _, addr := range n.Args.A.IPs {
-			_, _, err := net.ParseCIDR(addr)
+		for _, addrStr := range n.Args.A.IPs {
+			ip, addr, err := net.ParseCIDR(addrStr)
 			if err != nil {
-				return nil, "", fmt.Errorf("an entry in the 'ips' field is NOT in CIDR notation, got: '%s'", addr)
+				return nil, "", fmt.Errorf("an entry in the 'ips' field is NOT in CIDR notation, got: '%s'", addrStr)
 			}
-			n.IPAM.Addresses = append(n.IPAM.Addresses, Address{AddressStr: addr})
+			addr.IP = ip
+			n.IPAM.Addresses = append(n.IPAM.Addresses, Address{AddressStr: addrStr, Address: *addr})
 		}
 	}
 
@@ -205,12 +206,13 @@ func LoadIPAMConfig(bytes []byte, envArgs string) (*IPAMConfig, string, error) {
 	if len(n.RuntimeConfig.IPs) != 0 {
 		// runtimeConfig IP overwrites IP, so clear IPAM Config
 		n.IPAM.Addresses = make([]Address, 0, len(n.RuntimeConfig.IPs))
-		for _, addr := range n.RuntimeConfig.IPs {
-			_, _, err := net.ParseCIDR(addr)
+		for _, addrStr := range n.RuntimeConfig.IPs {
+			ip, addr, err := net.ParseCIDR(addrStr)
 			if err != nil {
-				return nil, "", fmt.Errorf("an entry in the 'ips' field is NOT in CIDR notation, got: '%s'", addr)
+				return nil, "", fmt.Errorf("an entry in the 'ips' field is NOT in CIDR notation, got: '%s'", addrStr)
 			}
-			n.IPAM.Addresses = append(n.IPAM.Addresses, Address{AddressStr: addr})
+			addr.IP = ip
+			n.IPAM.Addresses = append(n.IPAM.Addresses, Address{AddressStr: addrStr, Address: *addr})
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


plugins/ipam/static/main.go
```diff
@@ -161,7 +161,7 @@ func LoadIPAMConfig(bytes []byte, envArgs string) (*IPAMConfig, string, error) {
 
 				ip, subnet, err := net.ParseCIDR(ipstr)
 				if err != nil {
-					return nil, "", fmt.Errorf("invalid CIDR %s: %s", ipstr, err)
+					return nil, "", err
```
dougbtv, line 164: Maybe make some actionable error message such as:

```
fmt.Errorf("expected CIDR notation, got '%s' with error: %s", ipstr, err)
```

mccv1r0:
You pasted the same link twice:

```
[0] - https://github.com/golang/go/blob/2ebe77a2fda1ee9ff6fd9a3e08933ad1ebaea039/src/net/net.go#L540
[1] - https://github.com/golang/go/blob/2ebe77a2fda1ee9ff6fd9a3e08933ad1ebaea039/src/net/net.go#L540
```

maiqueb:
> You pasted the same link twice:
> 
> ```
> [0] - https://github.com/golang/go/blob/2ebe77a2fda1ee9ff6fd9a3e08933ad1ebaea039/src/net/net.go#L540
> [1] - https://github.com/golang/go/blob/2ebe77a2fda1ee9ff6fd9a3e08933ad1ebaea039/src/net/net.go#L540
> ```

Good catch. Corrected the second link.

plugins/ipam/static/main.go
maiqueb: @dougbtv your proposal would generate the string `"expected CIDR notation, got '10.10.0.1' with error: invalid CIDR address: 10.10.0.1"` which, imo, is also confusing. 

I don't mind using the first part of the string though. I think that as long as we stop wrapping the error we should be good; golang already gives us a pretty decent error msg w/ `"invalid CIDR address: 10.10.0.1"`

dougbtv: Yep, agreed. That definitely adds to the "chart junk" factor to have the IP address repeated like that. Probably more descriptive to use the field name (which I didn't look up, sorry), and I can probably make it a bit more actionable, maybe something like:

fmt.Errorf("The $foo field is expected to be in CIDR notation: %s", err)

maiqueb: The thing is that the IP can come from a field named 'ip' when passed via env vars, an array of IPs on a field named 'ips' on both cni args, and runtime config, and finally, an array of 'addresses'  via the ipam struct.

plugins/ipam/static/main.go
```diff
@@ -161,7 +161,7 @@ func LoadIPAMConfig(bytes []byte, envArgs string) (*IPAMConfig, string, error) {
 
 				ip, subnet, err := net.ParseCIDR(ipstr)
 				if err != nil {
-					return nil, "", fmt.Errorf("invalid CIDR %s: %s", ipstr, err)
+					return nil, "", fmt.Errorf("the 'ip' field is expected to be in CIDR notation, got: '%s'", ipstr)
```
mars1024, line 164: IMO, the new error message just make me more confused :), how about simplify to `failed to parse cidr: %v`

plugins/ipam/static/main.go
```diff
@@ -193,6 +193,10 @@ func LoadIPAMConfig(bytes []byte, envArgs string) (*IPAMConfig, string, error) {
 		// args IP overwrites IP, so clear IPAM Config
 		n.IPAM.Addresses = make([]Address, 0, len(n.Args.A.IPs))
 		for _, addr := range n.Args.A.IPs {
+			_, _, err := net.ParseCIDR(addr)
```
mars1024, line 196: the  string-format IP(or CIDR) will be parsed later, so why we check this twice?

plugins/ipam/static/main.go
maiqueb: +1. That was my original suggestion, but @dougbtv was keen on indicating which attribute was the offending one.

I can reset this back to `fmt.Errorf("Failed to parse CIDR: %s", ipstr)` . Would that work for both of you @dougbtv / @mars1024 ?

plugins/ipam/static/main.go
maiqueb: This was the way I found of indicating which parameter was the offending one (the string format CIDR validation will be performed for cni args / ipam config later on, and its impossible to know the parameter name by then).

But I agree with you - IMO, just saying we expect CIDR notation should be enough.

plugins/ipam/static/main.go
maiqueb: This is the current implementation I'm proposing.

@mars1024 , your thoughts ? fwiw, I agree with you. There's only CIDR looking thing on the spec, the users shouldn't be confused.

mars1024: Agreed on @dougbtv , let's stay the way now.

plugins/ipam/static/main.go
mars1024: OK, since the parse and validation already done here, could we append a parsed `Address` instead of a single string to avoid this to be parsed later?

maiqueb: I am not entirely fond of how it looks, but have provided a new commit w/ this change. 

Let me know if that's OK.

plugins/ipam/static/main.go
maiqueb: OK; resolving this item now.

maiqueb:
@mars1024 also, for whatever reason, CI stopped running the tests automatically; it complains a maintainer must approve testing for it.

Can you grant it ?

dougbtv (approved):
Thanks for the improvements, I think with this error messaging, it's more actionable for a user (they now can see that we're saying "hey, a.b.c.d that's not in CIDR format, you should change it." instead of just "this is wrong: a.b.c.d." (which doesn't tell us it's in the wrong format)

plugins/ipam/static/main.go
```diff
@@ -211,12 +221,15 @@ func LoadIPAMConfig(bytes []byte, envArgs string) (*IPAMConfig, string, error) {
 	numV6 := 0
 
 	for i := range n.IPAM.Addresses {
-		ip, addr, err := net.ParseCIDR(n.IPAM.Addresses[i].AddressStr)
-		if err != nil {
-			return nil, "", fmt.Errorf("invalid CIDR %s: %s", n.IPAM.Addresses[i].AddressStr, err)
+		if n.IPAM.Addresses[i].Address.IP == nil {
```
s1061123, line 224: Should we check `n.IPAM.Address[i].Address != nil` ibefore the accesss of `n.IPAM.Addresses[i].Address.IP` for corner case? (I believe that net.ParseCIDR() returns non nil for *IPNet), however, we may need to double check...

I'm worried segmentation fault at the access of `n.IPAM.Addresses[i].Address.IP`

maiqueb: > Should we check `n.IPAM.Address[i].Address != nil` ibefore the accesss of `n.IPAM.Addresses[i].Address.IP` for corner case? (I believe that net.ParseCIDR() returns non nil for *IPNet), however, we may need to double check...

I don't think `net.ParseCIDR` can return `nil` for `*IPNet` **without** also returning an error. 
https://cs.opensource.google/go/go/+/refs/tags/go1.15.13:src/net/ip.go;l=712

Furthermore, we're the ones putting the `net.Address` in the slice: https://github.com/containernetworking/plugins/pull/638/files/8ab23366fb09065838013bb464d83aaca1ac2f51#diff-0469cbf9b042cb5f9cd3a38b880957736c89e1b78770b600c1d845ca9d9be410R201
> 
> I'm worried segmentation fault at the access of `n.IPAM.Addresses[i].Address.IP`

I don't think that can happen (without `net.ParseCIDR` returning an error). But, for added safety, I can update the code to fail on

```golang
...
for _, addrStr := range n.Args.A.IPs {
	ip, addr, err := net.ParseCIDR(addrStr)
	if err != nil || addr == nil { // <==== HERE
		return nil, "", fmt.Errorf("an entry in the 'ips' field is NOT in CIDR notation, got: '%s'", addrStr)
	}
	addr.IP = ip
	n.IPAM.Addresses = append(n.IPAM.Addresses, Address{AddressStr: addrStr, Address: *addr})
}
...
```

maiqueb:
> @mars1024 also, for whatever reason, CI stopped running the tests automatically; it complains a maintainer must approve testing for it.
> 
> Can you grant it ?

ping @mars1024

mars1024 (approved):
/lgtm, thanks! Errors seem more pointed!

PR #645, opened 2021-07-21
ip: place veth peer in host namspace directly

dcbw:
Instead of moving the host side of the veth peer into the host
network namespace later, just create it in the host namespace
directly.


mars1024 (approved):
/lgtm

dcbw:
@squeed @mccv1r0 how does this one look?

Issue #646, opened 2021-07-30
macvlan plugin doesn't configure default route inside pod


syelburgi:
Hi,

when pod is created with macvlan cni plugin. inside the pod's network namespace, I don't see default route is getting configured. as a result pod cannot communicate to public network(not local).

MikeZappa87:
@syelburgi I left a comment on your PR #659 Does the "routes" key on the ipam not work here? See my network configuration in the PR comment! Have a good one!

PR #647, opened 2021-07-30
Fix issue 646 - default route for gateway

syelburgi:
Add default route for pod's namespace

### DIFF
--- a/plugins/main/macvlan/macvlan.go
+++ b/plugins/main/macvlan/macvlan.go
@@ -371,17 +371,6 @@ func cmdDel(args *skel.CmdArgs) error {
 			}
 		}
 
-        	routeToDstIP, err := netlink.RouteList(nil, netlink.FAMILY_ALL)
-        	if err == nil {
-            	    for _, v := range routeToDstIP {
-                	if v.Dst == nil {
-                    	    if err := netlink.RouteDel(&v); err != nil {
-                                fmt.Errorf("Failed to delete the default route")
-                        	return nil
-                    	    }
-                       }
-                    }
-                }
 		return nil
 	})
 
--- a/plugins/main/macvlan/macvlan.go
+++ b/plugins/main/macvlan/macvlan.go
@@ -308,6 +308,12 @@ func cmdAdd(args *skel.CmdArgs) error {
 					_ = arping.GratuitousArpOverIface(ipc.Address.IP, *contVeth)
 				}
 			}
+            		// add a gateway route
+            		link, err := netlink.LinkByName(args.IfName)
+            		route := netlink.Route{LinkIndex: link.Attrs().Index, Gw:ipamResult.IPs[0].Gateway, Flags: 0x10}
+            		if err := netlink.RouteAdd(&route); err != nil {
+				fmt.Errorf("failed to add route")
+            		}
 			return nil
 		})
 		if err != nil {
@@ -364,6 +370,18 @@ func cmdDel(args *skel.CmdArgs) error {
 				return err
 			}
 		}
+
+        	routeToDstIP, err := netlink.RouteList(nil, netlink.FAMILY_ALL)
+        	if err == nil {
+            	    for _, v := range routeToDstIP {
+                	if v.Dst == nil {
+                    	    if err := netlink.RouteDel(&v); err != nil {
+                                fmt.Errorf("Failed to delete the default route")
+                        	return nil
+                    	    }
+                       }
+                    }
+                }
 		return nil
 	})
 


syelburgi:
https://github.com/containernetworking/plugins/issues/646 - when pod is created with macvlan cni plugin. inside the pod's network namespace, I don't see default route is getting configured. as a result pod cannot communicate to public network(not local).

mars1024:
Hi @syelburgi , may I ask which IPAM plugin is invoking in your network configuration?

syelburgi:
Hi mars, I am using my own custom ipam .

mars1024:
> Hi mars, I am using my own custom ipam which works with calico

Actually, routes is configurable in macvlan plugin as long as `routes` in result was passed from IPAM plugin.
So if you want a default route, maybe code change should be in your custom IPAM plugin but not here.

syelburgi:
Even if the IPAM plugin passes the routes . it is not getting programmed and I don't see the code for it also.
result.Routes = ipamResult.Routes - These routes are fetched from the IPAM but not used for programming.

syelburgi:
Hi Mars,

I use Calcio with my Ipam plugin and I see calcio configuring the default route without no routes passed from ipam.
Any specific reason why macvlan requires special handling.?

mars1024:
> Hi Mars,
> 
> I use Calcio with my Ipam plugin and I see calcio configuring the default route without no routes passed from ipam.
> Any specific reason why macvlan requires special handling.?

Route configuration is a bigger topic more than the default route, so I think passing it in result is a more flexible and configurable way.

svallala:
> > Hi Mars,
> > I use Calcio with my Ipam plugin and I see calcio configuring the default route without no routes passed from ipam.
> > Any specific reason why macvlan requires special handling.?
> 
> Route configuration is a bigger topic more than the default route, so I think passing it in result is a more flexible and configurable way.

@mars1024, one correction is that if we use routes and pass destination as "0.0.0.0/0" then things are working fine. The question is all the other CNI plugins that we are using today are designed to take the "gateway" parameter passed in the IPAM result (as per the CNI spec) and configure the default gateway. Why is it that for MACVLAN CNI "gateway" parameter is not honored? Is this intentional or a miss? Do you see any issues with the fix being proposed by @syelburgi?

For example, in the CNI documentation at https://github.com/containernetworking/cni/blob/master/SPEC.md, the below example is shown where the gateway is passed in the IPAM. Why is MACVLAN not following this?

{
    "cniVersion": "1.0.0",
    "name": "dbnet",
    "type": "bridge",
    "bridge": "cni0",
    "keyA": ["some more", "plugin specific", "configuration"],
    "ipam": {
        "type": "host-local",
        "subnet": "10.1.0.0/16",
        "gateway": "10.1.0.1"
    },
    "dns": {
        "nameservers": [ "10.1.0.1" ]
    }
}

syelburgi:
@mars1024 any update on this PR

squeed:
So, this code is configuring the gateway as a default route, but `ipam.ConfigureInterface` already does that. It seems like the only difference is setting the RTNH_F_ONLINK flag.

Why do you need to set that? Is it something we should be setting for all plugins, or only for Macvlan?

plugins/main/macvlan/macvlan.go
```diff
@@ -308,6 +308,12 @@ func cmdAdd(args *skel.CmdArgs) error {
 					_ = arping.GratuitousArpOverIface(ipc.Address.IP, *contVeth)
 				}
 			}
+            		// add a gateway route
+            		link, err := netlink.LinkByName(args.IfName)
+            		route := netlink.Route{LinkIndex: link.Attrs().Index, Gw:ipamResult.IPs[0].Gateway, Flags: 0x10}
```
squeed, line 313: Be careful: this code can panic or return several errors:
1 - IPAM returning no addresses is pathological but valid
2 - IPAM results may not set a gateway.

`ipam.ConfigureInterface()` handles these cases correctly.

svallala: @squeed you are right regarding the handling of this. Checks should be in place to make sure the response has IPs and a gateway.

ipam.ConfigureIface() is handling the gateway, the only problem is that after getting the v4gw and v6gw, it is only configuring the gateway if the IPAM response has "routes" configured as well. This is not the behavior we are seeing with other CNI plugins for e.g SRIOV CNI.

Based on the spec "gateway" under "IPs" indicate that it is the default gateway and "routes" should be used when you have to configure static routes where both the "dst" and "gw" towards a particular host or a subnet should be specified.

@syelburgi based on the comments seems like instead of adding a function to add the route here, you should fix ipam.ConfiugreIface() to properly handle route addition

syelburgi:
For some Technical issue could not update the existing PR. New PR is raised. Url - https://github.com/containernetworking/plugins/pull/659

dcbw:
Closing in favor of the new PR.

PR #659, opened 2021-09-03
Fixes 646 - added route for default gateway

syelburgi:
Created New pull request

### DIFF
--- a/pkg/ipam/ipam_linux.go
+++ b/pkg/ipam/ipam_linux.go
@@ -103,21 +103,21 @@ func ConfigureIface(ifName string, res *current.Result) error {
 		ip.SettleAddresses(ifName, 10)
 	}
 
-	if v4gw != nil {
-        	route := netlink.Route{LinkIndex: link.Attrs().Index, Gw:v4gw, Flags: 0x10}
-			if err = netlink.RouteAddEcmp(&route); err != nil {
-				return fmt.Errorf("failed to add default route via %v dev %v': %v", v4gw, ifName, err)
-			}
-    	}
+    if v4gw != nil {
+        route := netlink.Route{LinkIndex: link.Attrs().Index, Gw:v4gw, Flags: 0x10}
+		if err = netlink.RouteAddEcmp(&route); err != nil {
+			return fmt.Errorf("failed to add default route via %v dev %v': %v", v4gw, ifName, err)
+		}
+    }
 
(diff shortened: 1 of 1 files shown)


pkg/ipam/ipam_linux.go
```diff
@@ -103,6 +103,20 @@ func ConfigureIface(ifName string, res *current.Result) error {
 		ip.SettleAddresses(ifName, 10)
 	}
 
+	if v4gw != nil {
+        	route := netlink.Route{LinkIndex: link.Attrs().Index, Gw:v4gw, Flags: 0x10}
```
dcbw, line 107: Looks like there's still some whitespace issues here?

dcbw:
@syelburgi looks like there are some testcase failures, could you investigate?

`plugin type="bridge" failed (add): failed to add default route via 10.12.2.1 dev eth0': invalid argument`

pkg/ipam/ipam_linux.go
dcbw: @syelburgi also, I think you might need to add the `Dst` option but set to 0.0.0.0 or ::0, right?

MikeZappa87: Dst should be used, it would be 0.0.0.0/0 in this case. I don't think the flags are needed here as well?

MikeZappa87:
I am curious about your IPAM json. The routes key allows this with no code changes. I tested this with the cnitool and verified the network namespace has a default route. @syelburgi are you able to supply the inputs? 

```json
{
   "cniVersion":"1.0.0",
   "name":"mynet",
   "plugins":[
      {
         "name":"mynet",
         "type":"macvlan",
         "master":"dummy0",
         "ipam":{
            "type":"host-local",
            "ranges":[
               [
                  {
                     "subnet":"10.88.0.0/16"
                  }
               ],
               [
                  {
                     "subnet":"2001:4860:4860::/64"
                  }
               ]
            ],
            "routes":[
               {
                  "dst":"0.0.0.0/0"
               },
               {
                  "dst":"::/0"
               }
            ]
         }
      }
   ]
}
```

squeed:
Hi there,

I'm pretty sure we don't want to synthesize a default route. There are use-cases where one doesn't want a container to have a default route. That's why we make you explicitly specify one in the host-local configuration...

PR #664, opened 2021-09-17
Use crypto/rand.Read, not crypto.Reader.Read

edef1c:
The current code accidentally ignores partial reads, since it doesn't
check the return value of (io.Reader).Read.

What we actually want is io.ReadFull(rand.Reader, buf), which is
conveniently provided by rand.Read(buf).


matthewdupre (approved):

jayunit100:
lgtm for sure bc I can't see why anyone would use the internal `Reader` implementation when there's a high level `Read(`

PR #665, opened 2021-09-17
Don't redundantly filepath.Clean the output of filepath.Join

edef1c:
filepath.Join is already specified to clean its output,
and the implementation indeed does so.


mars1024 (approved):
/lgtm

dcbw:
LGTM, thanks!

matthewdupre (approved):

PR #669, opened 2021-10-01
Allow setting sysctls on a particular interface

jellonek:
As in subject.


mars1024:
> An argument can be made that tuning plugin should have been interface specific from the beginning, that is, each e.g. `cniADD` for a conflist is specific to just one `IFNAME`. Should we just do that?

I'm wondering whether replacing `IFNAME` is too simple to be maintained or extended further, IMO, this problems can be solved well by template rendering, e.g. go-template or jinja2, no matter the arguments coming from CNI args, network config or dynamic configuration.

Issue #660, opened 2021-09-05
dhcp: support custom DHCP Request options


SilverBut:
Our production environment needs v4/v6 dual stack (and v6-only network), and we are using DHCP to manage address for different server/services/users. To get a address normally, DHCP `option 82`(https://www.rfc-editor.org/rfc/rfc3046.html) is required, which should be sent during DHCP-Request.

I'm writing code for this support, so in case of anyone else also writing code for this, please tell me if I can be any help.

SilverBut:
Also, #614 and #640 is related with processing received DHCP options. So maybe I can work on these two issue later, if possible. 

Besides, the new config syntax is required to hold those settings. I'm planning to use this as the config, which should be compatible with previous versions: 

```json
{
    "cniVersion": "0.4.0",
    "name": "example",
    "plugins": [
        {
            "type": "macvlan",
            "master": "bond0.150",
            "ipam": {
                "type": "dhcp",
                // use option name from dhcp-options(5)
                "request": [
                    {
                        "default": true // TODO: add some predefined field list which we can process
                    },
                    {
                        "option": "domain-search",
                        "required": false // by default false. if true, the lease is ignored, and wait for correct one
                    }
                ],
                "send": [
                    {
                        "default": false
                    },
                    {
                        "option": "agent.circuit-id",
                        "value": "blacksteel.cluster.server:bond0/vlan150"
                    },
                    {
                        "option": "agent.remote-id",
                        "value-source": "container-id" // possible values: "container-id", "net-ns", "if-name"
                    }
                ]
            }
        }
    ]
}
```

vivekskrishna:
Hi @SilverBut , have you made any progress in implementing this. i am happy to try this out.

SilverBut:
@vivekskrishna Hi. Not ready for release since it does not work as intended in my own server. I'm still debugging...

vivekskrishna:
@SilverBut Thanks for the update. Let me know if I can help in some way.

dcbw:
I think these would be a good improvement for the DHCP plugin. Thanks for looking into it!

SilverBut:
To make everyone know current status here is a brief report of my current work progress.

## Use argument from CNI arguments

[...]

PR #670, opened 2021-10-02
dhcp ipam: support customizing dhcp options from CNI args

### LINKED ISSUE #435: macvlan + dhcp doesn't forward hostname to dhcp server
I set up a macvlan network and when I start a container [0] it does get an address, however it appears that the container hostname is not passed to the dhcp (dnsmasq 2.80-4) server.

The entry in the dnsmasq leases file is:
1578703576 ce:a3:23:45:7b:b4 192.168.1.78 * 34:39:30:30:66:39:66:31:62:37:31:61:62:32 ( ... wth? 225 characters long for some reason!?)

Normal dhcp client lease entries:
1578670901 cc:46:4e:7b:0d:9c 192.168.1.34 little-bird 01:cc:46:4e:7b:0d:9c
1578703631 52:54:00:c0:81:1a 192.168.1.91 rhel7 *

Is there some other configuration point I need to add? Is this related to #367 ?

[0] # podman run -it --rm --hostname da1 --name da1 --network local <image>

~~~
90-local.conflist
{
    "cniVersion": "0.4.0",
    "name": "local",
    "plugins": [
        {
            "type": "macvlan",
            "master": "enp1s0",
            "ipam": {
                "type": "dhcp"
            }
        }
    ]
}
~~~

SilverBut:
Make some DHCP options sent to the server can be set from parsing CNI_ARGS.

This closes #435 and #660.

### DIFF
--- a/plugins/ipam/dhcp/daemon.go
+++ b/plugins/ipam/dhcp/daemon.go
@@ -71,9 +71,16 @@ func (d *DHCP) Allocate(args *skel.CmdArgs, result *current.Result) error {
 		return fmt.Errorf("error parsing netconf: %v", err)
 	}
 
+	optsRequesting, optsProviding, err := prepareOptions(args.Args, conf.IPAM.ProvideOptions, conf.IPAM.RequireOptions)
+	if err != nil {
+		return err
+	}
+
 	clientID := generateClientID(args.ContainerID, conf.Name, args.IfName)
 	hostNetns := d.hostNetnsPrefix + args.Netns
-	l, err := AcquireLease(clientID, hostNetns, args.IfName, d.clientTimeout, d.clientResendMax, d.broadcast)
+	l, err := AcquireLease(clientID, hostNetns, args.IfName,
+		optsRequesting, optsProviding,
+		d.clientTimeout, d.clientResendMax, d.broadcast)
 	if err != nil {
 		return err
 	}
--- a/plugins/ipam/dhcp/daemon.go
+++ b/plugins/ipam/dhcp/daemon.go
@@ -29,7 +29,6 @@ import (
 	"time"
 
 	"github.com/containernetworking/cni/pkg/skel"
-	"github.com/containernetworking/cni/pkg/types"
 	current "github.com/containernetworking/cni/pkg/types/100"
 	"github.com/coreos/go-systemd/v22/activation"
 )
@@ -62,7 +61,7 @@ func generateClientID(containerID string, netName string, ifName string) string 
 // Allocate acquires an IP from a DHCP server for a specified container.
 // The acquired lease will be maintained until Release() is called.
 func (d *DHCP) Allocate(args *skel.CmdArgs, result *current.Result) error {
-	conf := types.NetConf{}
+	conf := NetConf{}
 	if err := json.Unmarshal(args.StdinData, &conf); err != nil {
 		return fmt.Errorf("error parsing netconf: %v", err)
 	}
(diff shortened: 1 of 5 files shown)


jellonek (approved):
<3

vivekskrishna:
looks good

plugins/ipam/dhcp/daemon.go
```diff
@@ -55,21 +52,35 @@ func newDHCP(clientTimeout, clientResendMax time.Duration) *DHCP {
...
 }
 
+// TODO: current client ID is too long. At least the container ID should not be used directly.
+// A seperate issue is necessary to ensure no breaking change is affecting other users.
 func generateClientID(containerID string, netName string, ifName string) string {
-	return containerID + "/" + netName + "/" + ifName
+	clientID := containerID + "/" + netName + "/" + ifName
+	// defined in RFC 2132, length size can not be larger than 1 octet. So we truncate 254 to make everyone happy.
+	if len(clientID) > 254 {
+		clientID = clientID[0:254]
```
mars1024, line 61: IMO, `containerID` and `ifName` are enough to be composed as `clientID`.

plugins/ipam/dhcp/lease.go
```diff
@@ -432,7 +433,12 @@ func backoffRetry(resendMax time.Duration, f func() (*dhcp4.Packet, error)) (*dh
 
 		log.Print(err)
 
-		sleepTime = baseDelay + jitter(time.Second)
+		if fastRetryLimit == 0 {
```
dcbw, line 440: @SilverBut I'm a bit confused here... The CNI DHCP daemon that actually acquires the lease runs in the host network namespace and isn't dependent on the container network interface at all. So I don't understand what interface "is not up" in this scenario?

Can you explain this commit a bit more? Thanks!

plugins/ipam/dhcp/lease.go
```diff
@@ -432,7 +433,12 @@ func backoffRetry(resendMax time.Duration, f func() (*dhcp4.Packet, error)) (*dh
...
+			sleepTime = baseDelay + jitter(time.Second)
+		} else {
+			sleepTime = jitter(time.Second)
```
matthewdupre, line 443: I think jitter(1s) is uniformly distributed in [-1, 1], so half the time this will end up being a sleep(0) which is probably not intended.

plugins/ipam/dhcp/lease.go
SilverBut: Certainly! I think "is not up" is referring the whole DHCP infra, instead of the interface. DHCP is [not designed](https://datatracker.ietf.org/doc/html/rfc2131#:~:text=DHCP%20clients%20are%20responsible%20for%20all%20message%20retransmission.) for a perfect network without packet loss, and under some ~weird~ advanced network configurations, a client is not able to communicate with other hosts before an DHCP leases is granted by certain server. The DHCP server itself may also down for other reasons, including unplanned interruption or scheduled maintenance. 

So, in general, DHCP is not as reliable as other methods (like assigning address in a database like `etcd`). To make the address allocating run more smoothly, add retry here is necessary.

plugins/ipam/dhcp/lease.go
SilverBut: Oh, thanks for pointing it out! Fixed in c9d0423.

plugins/ipam/dhcp/daemon.go
SilverBut: This is actually not changed to prevent backward compatibility, since I'm not sure if anyone is using this default config in somewhere of their config...

plugins/ipam/dhcp/lease.go
dcbw: Thanks for the explanation.

dcbw:
/lgtm
Thanks!

PR #675, opened 2021-11-16
ipvlan: Send Gratuitous ARP after IPs are set

### LINKED ISSUE #673: IPVLAN does not support GARP
when using multus/IPVLAN, if pod schedule from one node to another node, GARP is not sent to update its neighbor arp cache. Then incoming traffic will not go to the new node instead going to old node until its neighbor arp cache timeout. 

The expected behavior is that when adding ip to ipvlan device, ipvlan can send gARP to update its neighbor arp cache, then incoming traffic can go to the correct node/pod.

TothFerenc:
This PR fixes the missing Gratuitous ARP functionality in the IPVLAN plugin.
The code is copied from the MACVLAN plugin: https://github.com/containernetworking/plugins/blob/e27c48b391539f5536918b9c379c59ef2793cb0d/plugins/main/macvlan/macvlan.go#L297-L311

Fixes #673

### DIFF
--- a/plugins/main/ipvlan/ipvlan.go
+++ b/plugins/main/ipvlan/ipvlan.go
@@ -18,8 +18,10 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"net"
 	"runtime"
 
+	"github.com/j-keck/arping"
 	"github.com/vishvananda/netlink"
 
 	"github.com/containernetworking/cni/pkg/skel"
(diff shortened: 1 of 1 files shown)


mars1024 (approved):
/lgtm, thanks!

jellonek (approved):

TothFerenc:
@dcbw Thanks for the merge!
Do you plan a release anytime soon?

PR #679, opened 2021-12-20
host-device: Bring interfaces up after moving into container

dpward:
If an interface is not configured with IPAM (because it functions at layer 2), it will not be brought up otherwise.

Fixes: #184


mars1024 (approved):
/lgtm, thanks!

mccv1r0:
Your test failed, can you look into it @dpward ?

dpward:
@mccv1r0 Sorry for the delay. This should now bring the device back to its previous up/down state after a failure when moving it from container back to host.

dcbw:
@dpward can you run gofmt on the hostdevice.go plugin?

```
go fmt checking failed:
plugins/main/host-device/host-device.go
```

PR #681, opened 2021-12-27
Fixed DHCP problem that broke when fast retry was added.

mjwaxios:
DHCP daemon could not get a lease when a retry was needed.  fastRetry caused backoffRetry to exit and not retry.  This change holds off the break from the retry loop until after it is back in normal delay.  This lets a retry happen and leases can be acquired.


mars1024 (approved):
/lgtm , good catch!

PR #686, opened 2022-01-06
Ignore NetNS path errors on delete

### LINKED ISSUE #685: If the NetNS path does not exist, error should not be returned, otherwise it may cause ip leakage
There are some situations that may cause the netns file (/proc/xxx/net/ns) to not exist, such as when the upper-level kubelet component cleans up pods.

Related discussion： https://github.com/kubernetes/kubernetes/issues/43014#issuecomment-287164444
> 	if NetNs is passed down by the Cloud Orchestration Engine, or if it called multiple times, so don't return an error if the device is already removed.

ip release failed, resulting in ip leak :
https://github.com/kubernetes/kubernetes/issues/107371

multus-cni also fixed this problem:
https://github.com/k8snetworkplumbingwg/multus-cni/pull/120

silenceper:
fix #685

### DIFF
--- a/plugins/main/bridge/bridge.go
+++ b/plugins/main/bridge/bridge.go
@@ -672,6 +672,13 @@ func cmdDel(args *skel.CmdArgs) error {
 	})
 
 	if err != nil {
+		//  if NetNs is passed down by the Cloud Orchestration Engine, or if it called multiple times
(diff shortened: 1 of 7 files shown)


squeed:
Great! Thanks

Issue #588, opened 2021-02-24
Use "arp_notify" sysctl instead of generating gratuitous arp ourselves


squeed:
@jellonek observed (in 2015....) that the kernel `arp_notify` sysctl will automatically generate gratuitous arps when interfaces change (i.e. go up or gain new v4 addresses).

We should just enable that, rather than manually generating the ARP packets ourselves.

songminglong:
Considering the compatibility of different systems and release versions, it is recommended to send the garp package directly to avoid relying on these special linux kernel versions

jayunit100:
related, https://github.com/containernetworking/plugins/pull/635 removes the OLD fix for this... 

and now, we want to remove THIS fix, and just rely on the kernel.

ryanlyy:
which kernel version can support it? looking at 4.18, it does not work and based the https://bugzilla.kernel.org/show_bug.cgi?id=194879, seems kernel does not plan to support it.

Thanks a lot

MikeZappa87:
To clarify this work is this just to remove the following:
https://github.com/containernetworking/plugins/blob/cc32993e9e9c88d0170fe5b03825ffa565e9354f/plugins/main/bridge/bridge.go#L563

https://github.com/containernetworking/plugins/blob/cc32993e9e9c88d0170fe5b03825ffa565e9354f/plugins/main/ipvlan/ipvlan.go#L270

And wherever else a gArp is sent? Would the consumer of the plugin be able to enable/disable arp_notify via the json config?

squeed:
@MikeZappa87 we used the gratruitous ARP in the case where addresses are "spinning" - i.e. on nodes where the available pool is small and there is a lot of churn. In these cases, we could have the case that one IP changes macs "relatively" quickly.

However, if there's evidence that assigning the address to the interface already sends a gratuitous arp, then we can remove the arping.

@ryanlyy this might still work, since we set the interface down and up. But I don't recall the exact order, so more investigation is required.

MikeZappa87:
Right on, I see where the case of where IP addresses are being assigned and immediately reclaimed to be assigned again where the mac address-table could potentially be stale with the previous mac address?

Does it make sense for just the container side to have arp_notify set?

Also kernel 4.18 is EOL'd? Does it work with 4.19

PR #687, opened 2022-01-07
Replace arping package with arp_notify

MikeZappa87:
This PR replaces the arping package with the linux arp_notify feature.

Each plugin with tested using the cnitool and wireshark. Each one generated the appropriate arp announcement. If you need pictures of this I can supply them. 

Resolves: #588

This might also resolve: https://github.com/containernetworking/plugins/pull/569


plugins/main/ptp/ptp.go
```diff
@@ -83,15 +83,17 @@ func setupContainerVeth(netns ns.NetNS, ifName string, mtu int, pr *current.Resu
...
-		if err = ipam.ConfigureIface(ifName, pr); err != nil {
-			return err
-		}
-
 		contVeth, err := net.InterfaceByName(ifName)
 		if err != nil {
 			return fmt.Errorf("failed to look up %q: %v", ifName, err)
 		}
 
+		_, _ = sysctl.Sysctl(fmt.Sprintf("net/ipv4/conf/%s/arp_notify", ifName), "1")
```
mccv1r0, line 91: It probably won't hurt anything (Linux should do the "right thing" ), but I don't think we need garp on a point to point link.

MikeZappa87: I can remove it, I was just keeping with the previous behavior.

MikeZappa87: @mccv1r0 I have removed the arp notify setting

squeed:
looks good to me, thanks!

matthewdupre (approved):

Issue #680, opened 2021-12-24
Investigation to speed up pod creation


MikeZappa87:
An effort is happening to speed up the pod creation and two areas of interest exist. One is the CNI setup and the other is the image pull, if both these items are improved pod setup can be done in less than a second. The area of interest in the CNI is this:

https://github.com/containernetworking/plugins/blob/master/pkg/ip/addr_linux.go

This method takes 1000ms - 1900ms to execute. This code is related to the Duplicate Address Detection feature of IPV6 and is looking at two flags Tentative and DadFailed. I still need to look over a few RFC's on this as well, including optimistic dad. 

If something was missed or overlooked let's tackle it!

Initial Observations:

- Having a boolean flag to bypass SettleAddreses reduced the execution to less than 100ms

- Errors from SettleAddresses are not handled and failures are silent

- Unable to find other codebases doing something similar (Calico as an example)

- Only related to IPV6 addresses

- Other parts of the code that assign IP addresses does not utilize SettleAddresses

- More likely to succeed vs fail

Questions:

- Is this method “SettleAddresses” still required?

- Was “SettleAddresses” implemented due to a shortcoming in the Linux Kernel/Netlink/Go Netlink Library? Has it since been resolved and renders this method obsolete?

- If Dad is disabled on the specific interfaces via “accept_dad” should SettleAddresses be used?

Proposals:

- Remove the code entirely

- Add a boolean flag to enable/disable SettleAddresses via runtime configuration (Eventually to be applied as an annotation to the pod spec)

- Configuration value that allows user to optionally not set the IPV6 address (Eventually to be applied as an annotation to the pod spec). User could specify IPV4,IPV6 or BOTH to assign

WorkArounds:
Do not use IPV6

Please feel free to reach out directly to me! I will help implement this and the go-cni/containerd side as well

uablrek:
> If Dad is disabled on the specific interfaces via “accept_dad” should SettleAddresses be used?

If accept_dad=0, does SettleAddresses really still take 1000-1900ms?

MikeZappa87:
> > If Dad is disabled on the specific interfaces via “accept_dad” should SettleAddresses be used?
> 
> If accept_dad=0, does SettleAddresses really still take 1000-1900ms?

I set accept_dad=0 to both sides of the veth and had no impact on execution time, might be something else? Will look at it late!

uablrek:
You must do it in the right order. I made a simple go program that does;
```go
	start := time.Now()
[...]

PR #688, opened 2022-01-08
Allow user to optionally disable IPV6 duplicate address detection via runtime configuration for bridge plugin 

MikeZappa87:
Hello, this PR is for #680 and more documentation exists around the motivation for this PR. An effort is underway to speed up pod creation time and the SettleAddresses method was the culprit and more specifically it was the duplicate address detection feature of IPV6. This PR allows the accept_dad setting to be set to "0" as a runtime configuration which disables that and allows the bridge setup to take less than 100 ms vs 1800-2200ms. Follow-on PR's will be submitted to go-cni,containerd and k8s as well. Look forward to hearing from you!

`{
   "cniVersion":"0.4.0",
   "name":"containerd-net",
   "plugins":[
      {
         "type":"bridge",
         "bridge":"cni0",
         "isGateway":true,
         "ipMasq":true,
         "promiscMode":true,
         "runtimeConfig":{
            "omitdad":true
         },
         "ipam":{
            "type":"host-local",
            "ranges":[
               [
                  {
                     "subnet":"10.88.0.0/16"
                  }
               ],
               [
                  {
                     "subnet":"2001:4860:4860::/64"
                  }
               ]
            ],
            "routes":[
               {
                  "dst":"0.0.0.0/0"
               },
               {
                  "dst":"::/0"
               }
            ]
         }
      },
      {
         "type":"portmap",
         "capabilities":{
            "portMappings":true
         }
      }
   ]
}`

`{
   "cniVersion":"0.4.0",
   "name":"containerd-net",
   "plugins":[
      {
         "type":"bridge",
         "bridge":"cni0",
         "isGateway":true,
         "ipMasq":true,
         "promiscMode":true,
         "omitdad":true,
         "ipam":{
            "type":"host-local",
            "ranges":[
               [
                  {
                     "subnet":"10.88.0.0/16"
                  }
               ],
               [
                  {
                     "subnet":"2001:4860:4860::/64"
                  }
               ]
            ],
            "routes":[
               {
                  "dst":"0.0.0.0/0"
               },
               {
                  "dst":"::/0"
               }
            ]
         }
      },
      {
         "type":"portmap",
         "capabilities":{
            "portMappings":true
         }
      }
   ]
}`

Related PR for Documentation: https://github.com/containernetworking/cni.dev/pull/95 Can submit another PR to update the documentation for the runtime configurations when implemented into containerd/k8s

### DIFF
--- a/plugins/main/bridge/bridge.go
+++ b/plugins/main/bridge/bridge.go
@@ -63,10 +63,12 @@ type NetConf struct {
 		Cni BridgeArgs `json:"cni,omitempty"`
 	} `json:"args,omitempty"`
 	RuntimeConfig struct {
-		Mac string `json:"mac,omitempty"`
+		Mac            string `json:"mac,omitempty"`
+		DisableIPV6DAD bool   `json:"disableipv6dad,omitempty"`
 	} `json:"runtimeConfig,omitempty"`
 
-	mac string
+	mac            string
+	disableipv6dad bool
 }
 
 type BridgeArgs struct {
@@ -121,6 +123,8 @@ func loadNetConf(bytes []byte, envArgs string) (*NetConf, string, error) {
 	if mac := n.RuntimeConfig.Mac; mac != "" {
 		n.mac = mac
 	}
+
+	n.disableipv6dad = n.RuntimeConfig.DisableIPV6DAD
 
 	return n, n.CNIVersion, nil
 }
@@ -301,7 +305,7 @@ func ensureBridge(brName string, mtu int, promiscMode, vlanFiltering bool) (*net
 	return br, nil
 }
 
-func ensureVlanInterface(br *netlink.Bridge, vlanId int) (netlink.Link, error) {
+func ensureVlanInterface(br *netlink.Bridge, vlanId int, disabledad bool) (netlink.Link, error) {
 	name := fmt.Sprintf("%s.%d", br.Name, vlanId)
 
 	brGatewayVeth, err := netlink.LinkByName(name)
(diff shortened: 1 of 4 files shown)


plugins/main/bridge/bridge.go
```diff
@@ -63,10 +63,12 @@ type NetConf struct {
 		Cni BridgeArgs `json:"cni,omitempty"`
 	} `json:"args,omitempty"`
 	RuntimeConfig struct {
-		Mac string `json:"mac,omitempty"`
+		Mac            string `json:"mac,omitempty"`
+		DisableIPV6DAD bool   `json:"disableipv6dad,omitempty"`
```
MikeZappa87, line 67: Lars had proposed 'omitDAD' vs 'DisableIPV6DAD' I actually like the omitDAD.

MikeZappa87:
I need to look at containerd since I believe I need to add ARGS support

squeed:
Question: do we need DAD at all? Can we disable it by default? Or replace it with optimistic dad?

squeed:
So, I had another think about this, and I'm not sure there ever will be a situation when we can't just enable optimistic DAD.

See, the only automatically configured interfaces - that is to say, ones that don't come from IPAM - will be autoconfigured link-local ones when we don't care about v6. That means that the address is uninteresting, and optimistic DAD is sufficient, since it will renumber the interface. All that will do is ensure no annoying log messages get to the kernel buffer.

So, what if we just set the sysctl `net/ipv6/conf/interface/optimistic_dad` before the link comes up and just be done with this?

MikeZappa87:
> So, I had another think about this, and I'm not sure there ever will be a situation when we can't just enable optimistic DAD.
> 
> See, the only automatically configured interfaces - that is to say, ones that don't come from IPAM - will be autoconfigured link-local ones when we don't care about v6. That means that the address is uninteresting, and optimistic DAD is sufficient, since it will renumber the interface. All that will do is ensure no annoying log messages get to the kernel buffer.
> 
> So, what if we just set the sysctl `net/ipv6/conf/interface/optimistic_dad = 1` before the link comes up and just be done with this?

@squeed I can investigate this today and let you know!

plugins/main/bridge/bridge.go
```diff
@@ -58,12 +58,14 @@ type NetConf struct {
 	PromiscMode  bool   `json:"promiscMode"`
 	Vlan         int    `json:"vlan"`
 	MacSpoofChk  bool   `json:"macspoofchk,omitempty"`
+	OmitDad      bool   `json:"omitdad,omitempty"`
```
MikeZappa87, line 61: One thing I am noticing is that accept_dad is an integer that has 0,1,2 values. I am not sure how that escaped me. I will need to resolve that

PR #689, opened 2022-01-11
bridge: add vlan trunk support

tjjh89017:
add vlan trunk support for veth
vlan trunk only support L2 only mode without any IPAM
refer ovs-cni design
https://github.com/k8snetworkplumbingwg/ovs-cni/blob/main/pkg/plugin/plugin.go

design:
origin "vlan" option will be PVID or untagged vlan for the network.
"vlanTrunk" will setup tagged vlan for veth.

entry type:
`{ "id": 100 }` will specify only tagged vlan 100
`{ "minID": 100, "maxID": 120 }` will specify tagged vlan from 100 to
120 (include 100 and 120)
vlanTrunk is a list of above entry type, so you can use this to add
tagged vlan
`[
  { "id": 100 },
  {
    "minID": 1000,
    "maxID": 2000
  }
]`

complete config will be like this
```json
{
  "cniVersion": "0.3.1",
  "name": "mynet",
  "type": "bridge",
  "bridge": "mynet0",
  "vlan": 100,
  "vlanTrunk": [
    { "id": 101 },
    { "minID": 1000, "maxID": 2000 },
    { "minID": 3000, "maxID": 4000 }
  ],
  "ipam": {}
}
```

Signed-off-by: Date Huang <date.huang@suse.com>

### DIFF
--- a/plugins/main/bridge/bridge.go
+++ b/plugins/main/bridge/bridge.go
@@ -21,6 +21,7 @@ import (
 	"net"
 	"os"
 	"runtime"
+	"sort"
 	"syscall"
 	"time"
 
@@ -46,17 +47,18 @@ const defaultBrName = "cni0"
 
 type NetConf struct {
 	types.NetConf
-	BrName       string `json:"bridge"`
-	IsGW         bool   `json:"isGateway"`
-	IsDefaultGW  bool   `json:"isDefaultGateway"`
-	ForceAddress bool   `json:"forceAddress"`
-	IPMasq       bool   `json:"ipMasq"`
-	MTU          int    `json:"mtu"`
-	HairpinMode  bool   `json:"hairpinMode"`
-	PromiscMode  bool   `json:"promiscMode"`
-	Vlan         int    `json:"vlan"`
-	MacSpoofChk  bool   `json:"macspoofchk,omitempty"`
-	EnableDad    bool   `json:"enabledad,omitempty"`
+	BrName       string       `json:"bridge"`
+	IsGW         bool         `json:"isGateway"`
+	IsDefaultGW  bool         `json:"isDefaultGateway"`
+	ForceAddress bool         `json:"forceAddress"`
+	IPMasq       bool         `json:"ipMasq"`
+	MTU          int          `json:"mtu"`
+	HairpinMode  bool         `json:"hairpinMode"`
+	PromiscMode  bool         `json:"promiscMode"`
+	Vlan         int          `json:"vlan"`
+	VlanTrunk    []*VlanTrunk `json:"vlanTrunk,omitempty"`
+	MacSpoofChk  bool         `json:"macspoofchk,omitempty"`
+	EnableDad    bool         `json:"enabledad,omitempty"`
 
 	Args struct {
 		Cni BridgeArgs `json:"cni,omitempty"`
@@ -65,7 +67,14 @@ type NetConf struct {
 		Mac string `json:"mac,omitempty"`
 	} `json:"runtimeConfig,omitempty"`
 
-	mac string
+	mac   string
+	vlans []int
+}
+
+type VlanTrunk struct {
+	MinID *int `json:"minID,omitempty"`
+	MaxID *int `json:"maxID,omitempty"`
+	ID    *int `json:"id,omitempty"`
 }
 
 type BridgeArgs struct {
@@ -101,6 +110,11 @@ func loadNetConf(bytes []byte, envArgs string) (*NetConf, string, error) {
 	if n.Vlan < 0 || n.Vlan > 4094 {
 		return nil, "", fmt.Errorf("invalid VLAN ID %d (must be between 0 and 4094)", n.Vlan)
 	}
+	var err error
+	if n.vlans, err = collectVlanTrunk(n.VlanTrunk); err != nil {
+		// fail to parsing
+		return nil, "", err
+	}
 
 	if envArgs != "" {
 		e := MacEnvArgs{}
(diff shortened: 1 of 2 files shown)


tjjh89017:
@dcbw could you take a look on it.
thanks a lot!

plugins/main/bridge/bridge.go
```diff
@@ -373,6 +440,16 @@ func setupVeth(netns ns.NetNS, br *netlink.Bridge, ifName string, mtu int, hairp
 		}
 	}
 
+	if vlans != nil {
+		// TODO setup vlan trunk
+		for _, v := range vlans {
+			err = netlink.BridgeVlanAdd(hostVeth, uint16(v), false, false, false, true)
+			if err != nil {
+				return nil, nil, fmt.Errorf("failed to setup vlan tag on interface %q: %v", hostIface.Name, err)
```
jonyhy96, line 448: may better to use `%w` rather than %v for err

tjjh89017: let me check the difference between %w and %v and feedback to you
tks for comment

tjjh89017: should we consider backward compatible to earlier golang?
if no, i will change to `%w`

plugins/main/bridge/bridge.go
```diff
@@ -373,6 +440,16 @@ func setupVeth(netns ns.NetNS, br *netlink.Bridge, ifName string, mtu int, hairp
 		}
 	}
 
+	if vlans != nil {
+		// TODO setup vlan trunk
```
squeed, line 444: Is this still a TODO?

squeed:
This is interesting. We had a question in the maintainers' meeting: how do you receive the packets inside the container? Are you using another process to create vlan-tagged interfaces inside? Or are you doing it in userspace?

squeed:
Another comment (mostly from @dcbw) -- is trunking a per-port setting? Is it possible / desired for different containers on the same bridge to have different trunking configurations?

plugins/main/bridge/bridge.go
tjjh89017: it seems i forgot to remove it.

tjjh89017:
> This is interesting. We had a question in the maintainers' meeting: how do you receive the packets inside the container? Are you using another process to create vlan-tagged interfaces inside? Or are you doing it in userspace?

In our scenario, kubevirt will setup another bridge to catch all traffic and let VM handle the VLAN tag
userspace program also could handle vlan tag with `NET_RAW`.
we also create vlan interface to filter vlan tag also.

> Another comment (mostly from @dcbw) -- is trunking a per-port setting? Is it possible / desired for different containers on the same bridge to have different trunking configurations?
Per port, yes!

linux equivalent command:
```bash
ip link add br0 type bridge vlan_filtering 1
ip link set br0 up
ip link set eth0 master br0
ip link add veth-red type veth peer name veth-blue
ip link set veth-red master br0
ip link set veth-red up
ip link add veth-green veth peer name veth-purple
ip link set veth-green master br0
ip link set veth-green up

birdge vlan add dev veth-red vid 100 pvid untagged
bridge vlan add dev veth-red vid 101
bridge vlan add dev veth-red vid 102

bridge vlan add dev veth-green vid 200 pvid untagged
bridge vlan add dev veth-green vid 201
bridge vlan add dev veth-green vid 202
```

plugins/main/bridge/bridge.go
dcbw: @tjjh89017 yes, please use %w. CNI follows current golang (1.17) and previous golang (1.16) which both have it. We are just a bit lazy in regularly bumping requirements.

dcbw:
@tjjh89017 could you rebase since we merged a fix for ipamDel()? Thanks!

plugins/main/bridge/bridge.go
```diff
@@ -377,6 +444,15 @@ func setupVeth(netns ns.NetNS, br *netlink.Bridge, ifName string, mtu int, hairp
 		}
 	}
 
+	if vlans != nil {
+		for _, v := range vlans {
+			err = netlink.BridgeVlanAdd(hostVeth, uint16(v), false, false, false, true)
+			if err != nil {
+				return nil, nil, fmt.Errorf("failed to setup vlan tag on interface %q: %v", hostIface.Name, err)
```
tjjh89017, line 451:
```suggestion
				return nil, nil, fmt.Errorf("failed to setup vlan tag on interface %q: %w", hostIface.Name, err)
```

tjjh89017:
@dcbw tks for help
already rebased it.
need your double chceck again and approve.

plugins/main/bridge/bridge.go
tjjh89017: changed to `%w` 
thanks for suggestion

tjjh89017:
@dcbw do we have any plan to support this feature?
thanks a lot

plugins/main/bridge/bridge.go
```diff
@@ -101,6 +110,11 @@ func loadNetConf(bytes []byte, envArgs string) (*NetConf, string, error) {
 	if n.Vlan < 0 || n.Vlan > 4094 {
 		return nil, "", fmt.Errorf("invalid VLAN ID %d (must be between 0 and 4094)", n.Vlan)
 	}
+	var err error
+	if n.vlans, err = collectVlanTrunk(n.VlanTrunk); err != nil {
```
austinvazquez, line 114: Would it be more consistent with the rest of the code to use the scoped variable and not define err?

plugins/main/bridge/bridge.go
```diff
@@ -124,6 +138,59 @@ func loadNetConf(bytes []byte, envArgs string) (*NetConf, string, error) {
 	return n, n.CNIVersion, nil
 }
 
+// This method is copied from https://github.com/k8snetworkplumbingwg/ovs-cni/blob/main/pkg/plugin/plugin.go
```
austinvazquez, line 141: Minor change. It may be useful to pin this link to a release version instead of following main.

tjjh89017:
@dcbw @squeed Hi 
Our project will still need this enhancement.
Could you guys give us some comment for this feature?
thanks a lot

bk201:
Please help reopen the issue. We still need this enhancement.

plugins/main/bridge/bridge.go
tjjh89017: We will still need to copy the value back to `n.vlans` later.
So I will consider to keep current code

plugins/main/bridge/bridge.go
tjjh89017: good idea
thanks a lot

squeed:
Arg, I somehow can't reopen. I'm sorry. Can you resubmit?

tjjh89017:
@squeed I will create a new PR for this commit.
I will link to here later.

PR #701, opened 2022-02-09
Added new option for preventing dhcp plugin to update ip table

MaxThom:
The DHCP plugin add a default route to the container. In Kubernetes, there is already a default route provided by k8s. 
The route added by the plugin make the container unable to access the network or the kubernetes intra network.

 This new option prevent the addition of the new route and mimic the host-local ipam type, but with the dhcp plugin.

# New option
`sudo nano /etc/cni/multus/net.d/sw-lan.conf`

```json
{
    "cniVersion": "0.4.0",
    "type": "macvlan",
    "name": "sw-lan",
    "master": "eno1",
    "mode": "bridge",
    "ipam": {
        "type": "dhcp",
        "discardDhcpRoutes": true
    }
}
```

**discardDhcpRoutes** is set to false by default. If set to true, it will prevent the dhcp plugin to update the container/pod ip route table.

### DIFF
--- a/plugins/ipam/dhcp/README.md
+++ b/plugins/ipam/dhcp/README.md
@@ -2,3 +2,34 @@
 This document has moved to the [containernetworking/cni.dev](https://github.com/containernetworking/cni.dev) repo.
 
 You can find it online here: https://cni.dev/plugins/current/ipam/dhcp/
+
+# Dev
+`sudo rm -f /run/cni/dhcp.sock`
+
+`sudo go run . daemon`
+
+# Building
+`go build`
+
+`sudo rm -f /run/cni/dhcp.sock`
+
+`sudo ./dhcp daemon`
+
+# New option
+`sudo nano /etc/cni/multus/net.d/sw-lan.conf`
+
+```json
+{
+    "cniVersion": "0.4.0",
+    "type": "macvlan",
+    "name": "sw-lan",
+    "master": "eno1",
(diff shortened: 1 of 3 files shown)


dcbw:
@s1061123 do you have any thoughts here?

dcbw:
It looks like the behavior is actually specific to the plugin in lease.go to add the default route:

```
	// The CNI spec says even if there is a gateway specified, we must
	// add a default route in the routes section.
	if gw := l.Gateway(); gw != nil {
		_, defaultRoute, _ := net.ParseCIDR("0.0.0.0/0")
		routes = append(routes, &types.Route{Dst: *defaultRoute, GW: gw})
	}
```

but the PR here would prevent *all* routes advertised by the DHCP server from being used. I think we only want to suppress the specific default route behavior here, right?

s1061123:
As far as I looked into the PR code, it seems to discard ALL routes from DHCP, even though you just wants to remove default route.

DHCP may provide additional route with 'option 121', described in https://datatracker.ietf.org/doc/html/rfc3442 hence discard all routes from DHCP is not what the DHCP provider (administrator) want to, I suppose. It may lose the network routes which may want to keep. 

If you just want to remove only default route, we have [cni-route-override](https://github.com/redhat-nfvpe/cni-route-override), which modify the CNI plugins' routing table information. This plugin also has `flushgateay` so you can use it with this option, it may be solved, I hope...

PR #702, opened 2022-02-14
call ipam.ExecDel after clean up device in netns

### LINKED ISSUE #666: bridge plugin:call ipam.ExecDel after clean device in netns
https://github.com/containernetworking/plugins/blob/master/plugins/main/bridge/bridge.go#L635
now bridge plugin call ipam.ExecDel firstly and then clean device in netns when DEL, if  execute ipam.ExecDel successful but clean device failed, in this situation the ip has been released in ipam server, while the ip was still used in this netns, may cause ip conflict if ipam allocate the ip to another new pod.

should call ipam.ExecDel after clean device in netns ?
thanks

gojoy:
call ipam.ExecDel to release ip in ipam after clean up device in netns.,avoid ip conflict
fix #666

Signed-off-by: gojoy <729324352@qq.com>

### DIFF
--- a/plugins/main/bridge/bridge.go
+++ b/plugins/main/bridge/bridge.go
@@ -631,14 +631,17 @@ func cmdDel(args *skel.CmdArgs) error {
 
 	isLayer3 := n.IPAM.Type != ""
 
-	if isLayer3 {
-		if err := ipam.ExecDel(n.IPAM.Type, args.StdinData); err != nil {
-			return err
+	ipamDel := func() error {
+		if isLayer3 {
+			if err := ipam.ExecDel(n.IPAM.Type, args.StdinData); err != nil {
+				return err
+			}
 		}
+		return nil
 	}
 
 	if args.Netns == "" {
-		return nil
+		return ipamDel()
 	}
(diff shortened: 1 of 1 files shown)


plugins/main/bridge/bridge.go
```diff
@@ -660,11 +663,16 @@ func cmdDel(args *skel.CmdArgs) error {
 		// https://github.com/kubernetes/kubernetes/issues/43014#issuecomment-287164444
 		_, ok := err.(ns.NSPathNotExistErr)
 		if ok {
-			return nil
+			return ipamDel()
 		}
 		return err
 	}
 
+	// call ipam.ExecEel after clean up device in netns
```
dcbw, line 671: @gojoy "ExecDel" I think?

gojoy: > @gojoy "ExecDel" I think?
yes,has been corrected

dcbw:
@gojoy could squash the new commit into the first one, so that the DCO is OK?

squeed:
lgtm, thanks

PR #704, opened 2022-02-18
Fix Host Veth Mac in CNI Result (Breaks Check)

MikeZappa87:
I noticed when using the cnitool and implementing the check command in containerd that the check command was constantly failing. The host side veth was always wrong. I will check the current state of the tests. 

I am not certain if this happens else, all my machines this does, however I don't know how it would work since I believe CRI-O calls CHECK right after ADD

Signed-off-by: Michael Zappa <Michael.Zappa@stateless.net>

### DIFF
--- a/plugins/main/bridge/bridge.go
+++ b/plugins/main/bridge/bridge.go
@@ -358,7 +358,6 @@ func setupVeth(netns ns.NetNS, br *netlink.Bridge, ifName string, mtu int, hairp
 	if err != nil {
 		return nil, nil, fmt.Errorf("failed to lookup %q: %v", hostIface.Name, err)
 	}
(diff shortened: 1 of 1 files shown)


plugins/main/bridge/bridge.go
```diff
@@ -358,7 +358,6 @@ func setupVeth(netns ns.NetNS, br *netlink.Bridge, ifName string, mtu int, hairp
 	if err != nil {
```
dcbw, line 358: Can you remove this block of code since it's now superfluous?

MikeZappa87: This turned out to be something we were doing. I am closing this.

PR #707, opened 2022-02-21
portmap: fix checkPorts result when chain does not exist

danwinship:
checkPorts would return nil rather than an error if the per-container DNAT chain didn't exist, meaning CHECK would erroneously return success rather than failure.

chain.check() already (correctly) checks that the chain exists, so there's no need to do it separately before calling that anyway.


danwinship:
cc @squeed @mccv1r0

dcbw:
LGTM

PR #709, opened 2022-02-27
ipam/dhcp: Fix client id in renew/release

fwiesel:
The client id was constructed differently in the acquire
function compared to the release and renew functions,
which caused the dhcp-server to consider it a different client.
This is now encapsulated in a common function.


squeed:
oops! Once this merges, I'll cut a new release. This is a silly one.

PR #712, opened 2022-03-04
bug: return errors when iptables and ip6tables are unusable

### LINKED ISSUE #711: If both ipv4 iptables fails and ipv6 tables fail, the failure reason is lost
https://github.com/containernetworking/plugins/blob/b8a10bbe111e9db72433f357c37077554ada7ca1/plugins/meta/portmap/portmap.go#L125-L127

If `maybeGetIptables` for both ipv4 and ipv6 returns nil, a generic error is returned that doesn't help the user diagnose the problem. If only one of them fails, it appears that a much more helpful error is generated.

https://github.com/containernetworking/plugins/blob/b8a10bbe111e9db72433f357c37077554ada7ca1/plugins/meta/portmap/portmap.go#L130-L132

Should those more helpful errors also be generated for the double failure?

joejulian:
When a problem occurs accessing both iptables and ip6tables in portmap, return an error that contains the origin errors.

Fixes #711

### DIFF
--- a/plugins/meta/portmap/portmap.go
+++ b/plugins/meta/portmap/portmap.go
@@ -120,10 +120,13 @@ func checkPorts(config *PortMapConf, containerNet net.IPNet) error {
 	dnatChain := genDnatChain(config.Name, config.ContainerID)
 	fillDnatRules(&dnatChain, config, containerNet)
 
-	ip4t := maybeGetIptables(false)
-	ip6t := maybeGetIptables(true)
+	ip4t, err4 := maybeGetIptables(false)
+	ip6t, err6 := maybeGetIptables(true)
 	if ip4t == nil && ip6t == nil {
-		return fmt.Errorf("neither iptables nor ip6tables usable")
+		err := fmt.Errorf("neither iptables nor ip6tables is usable")
+		err = fmt.Errorf("%v, (iptables) %v", err, err4)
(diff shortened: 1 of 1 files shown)


plugins/meta/portmap/portmap.go
```diff
@@ -120,10 +120,21 @@ func checkPorts(config *PortMapConf, containerNet net.IPNet) error {
...
 	fillDnatRules(&dnatChain, config, containerNet)
 
-	ip4t := maybeGetIptables(false)
-	ip6t := maybeGetIptables(true)
+	ip4t, err4 := maybeGetIptables(false)
+	ip6t, err6 := maybeGetIptables(true)
 	if ip4t == nil && ip6t == nil {
-		return fmt.Errorf("neither iptables nor ip6tables usable")
+		err := fmt.Errorf("neither iptables nor ip6tables is usable")
+		if err4 != nil {
```
mars1024, line 127: If `ip4t` and `ip6t` are both nil, then `err4` and `err6` should both be non-nil logically. So IMO this judgement is not necessary.

joejulian:
> /lgtm
> 
> (I thought about error wrapping, but since both the errors are at the same level, not sure that makes as much sense...)

I considered it, as well, but then thought the same.

mars1024 (approved):
/lgtm
(forget about the error wrapping)

Issue #714, opened 2022-03-09
loopback plugin need rootNS validation reinforcement


Poor12:
In the case of repeated creation and deletion of containers, when the request is queued and delayed processing, there is a probability that the netns /proc/{pid}/ns/net in the cni request expires, at which time the original container has been destroyed, the original process has exited, and pid is occupied by other host processes, resulting in the namespace becoming the host namespace. When uninstalling the container LO NIC, the host's LO NIC is uninstalled by mistake.

squeed:
This is definitely a problem! But, existing container runtimes (podman, docker) no longer pass pid-based namespaces anymore! Who is doing this!?

squeed:
We talked about this at the maintainer's meeting, and we decided it would be better to fix this in `skel` for all plugins. Basically, one should fail to execute if the plugin's netns and CNI_NETNS point to the same place. There could be an environment variable, CNI_NETNS_OVERRIDE if this is somehow desired.

Make sense? Is this something you could take care of, @Poor12?

Poor12:
I basically understand what you mean, but I want to confirm one thing, is CNI_NETNS_OVERRIDE similar to a switch to determine whether netns verification is required? @squeed

squeed:
@Poor12 exactly right - if that environment variable is set, then we skip the same-netns check. Otherwise we should just add this to `skel`, and thus include in every plugin.

PR #715, opened 2022-03-09
Fix 714-loopback plugin add rootNS validation reinforcement

Poor12:
loopback plugin add rootNS validation reinforcement.
fix: #714

### DIFF
--- a/plugins/main/loopback/loopback.go
+++ b/plugins/main/loopback/loopback.go
@@ -21,6 +21,7 @@ import (
 	"net"
 
 	"github.com/vishvananda/netlink"
+	"github.com/vishvananda/netns"
 
 	"github.com/containernetworking/cni/pkg/skel"
 	"github.com/containernetworking/cni/pkg/types"
@@ -47,6 +48,26 @@ func parseNetConf(bytes []byte) (*types.NetConf, error) {
 	}
 
 	return conf, nil
+}
+
+func GetRootNetNS() (netns.NsHandle, error) {
+	return netns.GetFromPid(1)
+}
+
+func IsRootNetNS(nsPath string) (bool, error) {
+	ns, err := netns.GetFromPath(nsPath)
+	if err != nil {
+		return false, err
+	}
+	defer ns.Close()
+
+	rootNS, err := GetRootNetNS()
+	if err != nil {
+		return false, err
+	}
+	defer rootNS.Close()
+
+	return rootNS.Equal(ns), nil
 }
 
 func cmdAdd(args *skel.CmdArgs) error {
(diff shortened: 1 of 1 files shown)


plugins/main/loopback/loopback.go
```diff
@@ -49,6 +50,36 @@ func parseNetConf(bytes []byte) (*types.NetConf, error) {
...
+}
+
+func IsRootNetNS(nsPath string) (bool, error) {
+	ns, err := netns.GetFromPath(nsPath)
+	if err != nil {
+		return false, err
+	}
+	defer func() {
+		err = ns.Close()
+		if err != nil {
```
squeed, line 64: Since the program is going away, you can ignore the result from `ns.Close()`

plugins/main/loopback/loopback.go
```diff
@@ -49,6 +50,36 @@ func parseNetConf(bytes []byte) (*types.NetConf, error) {
...
+			fmt.Errorf("close netns %s failed, %v", nsPath, err)
+		}
+	}()
+
+	rootNS, err := GetRootNetNS()
+	if err != nil {
+		return false, err
+	}
+	defer func() {
+		err = rootNS.Close()
```
squeed, line 74: same here - can drop this check

plugins/main/loopback/loopback.go
squeed: In other words, `defer ns.Close()`

matthewdupre (commented):
Code looks code otherwise (see test failures and other comments), but I don't think this should return an error to delete: printing a message is fine, but we should treat this as a successful delete because the intended namespace (and therefore lo) has already gone.

We don't want a runtime to retry, for example.

plugins/main/loopback/loopback.go
Poor12: Thanks for your advice. Done

Poor12:
> Code looks code otherwise (see test failures and other comments), but I don't think this should return an error to delete: printing a message is fine, but we should treat this as a successful delete because the intended namespace (and therefore lo) has already gone.
> 
> We don't want a runtime to retry, for example.

Exactly direct return is better than return an error.

Poor12:
@squeed @matthewdupre Beg your review and approval. Thanks.

squeed:
Something is wrong with github where it's not running actions. So hopefully CI runs soon :)

Poor12:
> Something is wrong with github where it's not running actions. So hopefully CI runs soon :)

seems CI succeed.

Poor12:
Need some reviews. BTW, when will the next release of the community be released?

plugins/main/loopback/loopback.go
```diff
@@ -49,6 +50,26 @@ func parseNetConf(bytes []byte) (*types.NetConf, error) {
 	return conf, nil
 }
 
+func GetRootNetNS() (netns.NsHandle, error) {
```
mars1024, line 53: /lgtm
But I think it's better to move these two functions to [https://github.com/containernetworking/plugins/blob/main/pkg/ns/ns_linux.go](https://github.com/containernetworking/plugins/blob/main/pkg/ns/ns_linux.go), seems better

Poor12: I have talked with squeed and I think it‘s better to fix it in `skel` for all plugins. Let's close it.

PR #725, opened 2022-03-31
V2 API support for win-overlay CNI

selansen:
This PR bring V2 API support into win-overlay CNI. With the current V1
API, only docker runtime works for win-overlay. By bringing new changes, we
should be able to use containerd as the runtime.Below are the key
points regarding this implementation.
	1. Clear seperation for V1 & V2 API support in cmdAdd
	2. New cni.conf sample that works for win-overlay
        3. cmdDel  handles HCN API changes for V2 API version

Signed-off-by: selansen <esiva@redhat.com>

### DIFF
--- a/plugins/main/windows/win-overlay/win-overlay_windows.go
+++ b/plugins/main/windows/win-overlay/win-overlay_windows.go
@@ -17,11 +17,13 @@ package main
 import (
 	"encoding/json"
 	"fmt"
+	"net"
 	"runtime"
 	"strings"
 
 	"github.com/Microsoft/hcsshim"
 
+	"github.com/Microsoft/hcsshim/hcn"
 	"github.com/containernetworking/cni/pkg/skel"
 	"github.com/containernetworking/cni/pkg/types"
 	current "github.com/containernetworking/cni/pkg/types/100"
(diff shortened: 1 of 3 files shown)


mansikulkarni96:
Fixes #713

selansen:
Want feedback from Microsoft networking team: @daschott   @Keith-Mange  PTAL

daschott (commented):
Nice work. I can see that AddHcnEndpoint is calling AddNamespaceEndpoint and the CNI ADD call looks good to me.

For the deletion, I believe it is missing an attempt to detach from namespace. Deleting the endpoint without detaching from namespace is fine as a fallback if the namespace is not found, but ideally we should try to detach from namespace first and attempt to clean up properly.

plugins/main/windows/win-overlay/win-overlay_windows.go
```diff
@@ -140,15 +232,34 @@ func cmdAdd(args *skel.CmdArgs) error {
...
+	return result, nil
+}
+func cmdAdd(args *skel.CmdArgs) error {
+	n, cniVersion, err := loadNetConf(args.StdinData)
+	if err != nil {
+		return err
+	}
+
+	var result *current.Result
+	if n.ApiVersion == 2 {
```
MikeZappa87, line 253: I am not 100% familiar with HNS/HCN, does HCN only work with API version 2? What about v1?

plugins/main/windows/win-overlay/win-overlay_windows.go
```diff
@@ -55,16 +57,106 @@ func loadNetConf(bytes []byte) (*NetConf, string, error) {
...
+		if err != nil {
+			return nil, errors.Annotatef(err, "error while executing IPAM addition")
+		}
+
+		// convert whatever the IPAM result was into the current result
+		result, err := current.NewResultFromResult(r)
+		if err != nil {
+			return nil, errors.Annotatef(err, "error while converting the result from IPAM addition")
+		}
+		if len(result.IPs) == 0 {
```
MikeZappa87, line 76: Does this still work with IPV6 addresses?

plugins/main/windows/win-overlay/win-overlay_windows.go
mansikulkarni96: yes that is my understanding from the [hcsshim code](https://github.com/microsoft/hcsshim/blob/master/hcn/hcn.go#L15) and the [MSFT docs](https://docs.microsoft.com/en-us/windows-server/networking/technologies/hcn/hcn-top). HNS works with V1. This PR is creating a separation between both workflows.

daschott:
Thanks for updating the delete call. There are some minor improvements that could be done w.r.t. querying HNS, but that isn't blocking and can come in a future PR. LGTM

plugins/main/windows/win-overlay/win-overlay_windows.go
```diff
@@ -55,16 +57,106 @@ func loadNetConf(bytes []byte) (*NetConf, string, error) {
...
+	if len(n.EndpointMacPrefix) != 0 {
+		if len(n.EndpointMacPrefix) != 5 || n.EndpointMacPrefix[2] != '-' {
+			return nil, fmt.Errorf("endpointMacPrefix [%v] is invalid, value must be of the format xx-xx", n.EndpointMacPrefix)
+		}
+	} else {
+		n.EndpointMacPrefix = "0E-2A"
+	}
+
+	networkName := n.Name
+	hnsNetwork, err := hcsshim.GetHNSNetworkByName(networkName)
```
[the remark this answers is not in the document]
selansen: @daschott I was running into an issue where I was not able to get the exact gateway IP from the HCN network object. The way win-bridge CNI gets the gateway IP address is returning me a "different IP" when I tried to use it in win-overlay. That is the only reason I am still using the HNS network object. I already discussed it with @mansikulkarni96 and will take it as a separate improvement PR.  I need to dig and find the right way to get the default IP.
If you are aware of getting the correct gateway from the HCN network object, pls do let me know.

ccamacho:
Folks, could you cut a new release so this change becomes available? Thanks!

mansikulkarni96:
Included in release v1.2.0

Issue #697, opened 2022-01-31
Proposal : Worker nodes running out of IPs in IP range after node reboot


Nishant-Nayan:
I have a kubernetes cluster with one worker node.
When I schedule a deployment of say 'n' pods and then reboot my worker node, then after the node comes up, I can see the old 'n' ips still present in '/var/lib/cni/networks/{dataDir}' directory, it also contains the newly assigned 'n' ips starting from the ip mentioned in 'last_reserved_ip' file. 
So totally I got 2n ips of which 'n' are unused. If I reboot my worker node further then eventually I will run out of Ip address throwing me then error "no IP addresses available in range set : <ip-range>" .

Proposed fix - One possible fix could be to track the reboots of worker nodes. When the node comes up then we can differentiate which ip files are created before the last reboot time and delete those as they are unused and keep the one's which are newly assigned (i.e after the last reboot).

Any suggestions/feedback on this would be appreciated.

mel1nn:
+1 for this kind of feature

dgloe-hpe:
I've noticed this same issue; here's the workaround we're using.

Create a file /usr/lib/tmpfiles.d/macvlan.conf with this content:
```
# Remove local macvlan reserved IP address information at boot time
r! /var/lib/cni/networks/*/*
```

That will cause systemd to remove the leftover reservations at boot time.

PR #728, opened 2022-04-06
Prevent IP overflow by deleting redundant files.

Nishant-Nayan:
Description: After reboot, the pods come up with a new set of IPs, but the previously assigned IP files remains in the "/var/lib/cni/networks/{dataDir}" directory. After some 'n' reboots, the worker node runs out of IP addresses due to the accumulation of redundant IP files.

Overview of fix: last reboot time of the worker node and last modified time of IP address files are being used.
If an IP is requested whose file is already present in '/var/lib/cni/networks/{dataDir}/' , then its modified time is checked, if that is before the last reboot time then the file is considered redundant and is deleted. After that a new file with same IP name will be created and new container id and interface name will be written into it.(IP gets allocated)

Resolves https://github.com/containernetworking/plugins/issues/697

Test Case :
Tested on a kubernetes cluster. Scheduled a deployment of 150 pods on 2 workers.
I kept rebooting the worker nodes but did not encounter IP shortage.

### DIFF
--- a/plugins/ipam/host-local/backend/disk/backend.go
+++ b/plugins/ipam/host-local/backend/disk/backend.go
@@ -18,9 +18,11 @@ import (
 	"io/ioutil"
 	"net"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"runtime"
 	"strings"
+	"time"
 
 	"github.com/containernetworking/plugins/plugins/ipam/host-local/backend"
 )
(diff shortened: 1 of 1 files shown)


dcbw:
Another option is to mount the cachedir as a tmpfs...

mars1024:
Is it better to allocate(or reuse) IP by `PodNamespace + Pod Name` rather than `SandboxID` in host-local plugin?

PR #730, opened 2022-04-19
Check for duplicated sysctl keys

mmirecki:
The tuning-cni config currently allows to specif duplicated sysctls, also with conflicting values. Duplicated sysctl entries are usually  not used intentionally, but are rather the result of an error in configuration.
This PR adds a check for duplicated sysctl's.

### DIFF
--- a/plugins/meta/tuning/tuning.go
+++ b/plugins/meta/tuning/tuning.go
@@ -304,6 +304,9 @@ func restoreBackup(ifName, containerID, backupPath string) error {
 }
 
 func cmdAdd(args *skel.CmdArgs) error {
+	if err := validateSysctlConflictingKeys(args.StdinData); err != nil {
+		return err
+	}
 	tuningConf, err := parseConf(args.StdinData, args.Args)
 	if err != nil {
 		return err
@@ -405,6 +408,9 @@ func main() {
 }
 
 func cmdCheck(args *skel.CmdArgs) error {
+	if err := validateSysctlConflictingKeys(args.StdinData); err != nil {
+		return err
+	}
 	tuningConf, err := parseConf(args.StdinData, args.Args)
 	if err != nil {
 		return err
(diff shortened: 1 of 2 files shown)


squeed:
@mmirecki looks good! Can you fix DCO and go fmt? Then I can merge when CI goes green.

plugins/meta/tuning/tuning.go
```diff
@@ -542,3 +545,27 @@ func readAllowlist() (bool, []string, error) {
...
+	if _, exists := sysctlDuplicatesMap[key]; exists {
+		return errors.New("duplicated sysctl keys are not allowed")
+	}
+	sysctlDuplicatesMap[key] = ""
+	return nil
+}
+
+func validateSysctlConflictingValues(data []byte) error {
+	// Need to reset values for use in tests
+	//sysctlDuplicatesMap = map[sysctlKey]interface{}{}
```
squeed, line 568: Did you mean to leave this commented out?

mmirecki: That's taken care of in the tests. Fixed.
Thanks

plugins/meta/tuning/tuning.go
```diff
@@ -542,3 +545,25 @@ func readAllowlist() (bool, []string, error) {
...
+func (d *sysctlKey) UnmarshalText(data []byte) error {
+	key := sysctlKey(string(data))
+	if _, exists := sysctlDuplicatesMap[key]; exists {
+		return errors.New("duplicated sysctl keys are not allowed")
+	}
+	sysctlDuplicatesMap[key] = ""
+	return nil
+}
+
+func validateSysctlConflictingValues(data []byte) error {
```
mars1024, line 566: Actually this function is checking conflicting keys not values, i think it should be renamed?

mmirecki: indeed, fixed

squeed:
@mmirecki @dcbw observes we should do this when parsing the config for CHECK too, right?

plugins/meta/tuning/tuning.go
mmirecki: > @mmirecki @dcbw observes we should do this when parsing the config for CHECK too, right?

@squeed @dcbw Done

mccv1r0 (approved):

PR #733, opened 2022-04-21
bridge: support IPAM DNS settings

### LINKED ISSUE #431: bridge:get dns config from /etc/cni/net.d default config, not from ipam plugins
Referring to ducuments, bridge plugins does not contain dns config which ipam host-local and static plugins does. 
But i found that bridge get dns config from /etc/cni/net.d original config, not from ipam plugins by reading and testing the source code.

Kern--:
Previously, the bridge plugin ignored DNS settings returned
from an IPAM plugin (e.g. the host-local plugin parsing
resolv.conf to configure DNS). With this change, the bridge plugin
uses IPAM DNS settings.

Similarly to #388, this change will use incoming DNS settings if set,
otherwise IPAM plugin returned DNS settings

Fixes #431

Signed-off-by: Kern Walster <walster@amazon.com>

### DIFF
--- a/plugins/main/bridge/bridge.go
+++ b/plugins/main/bridge/bridge.go
@@ -489,6 +489,7 @@ func cmdAdd(args *skel.CmdArgs) error {
 
 		result.IPs = ipamResult.IPs
 		result.Routes = ipamResult.Routes
+		result.DNS = ipamResult.DNS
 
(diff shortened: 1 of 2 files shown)


mars1024 (approved):
basically LGTM

Issue #737, opened 2022-04-27
plugins size add up and is too big


martinetd:
Hello,

I'm using podman with cni-plugins on a minimal alpine host, whose only job is to run containers.. and almost 1/4 of the installed size comes from cni-plugins.

Each plugin isn't so big but it adds up; this was on alpine package 1.0.1-r1
```
2.6M	/usr/libexec/cni/bandwidth
2.9M	/usr/libexec/cni/bridge
6.5M	/usr/libexec/cni/dhcp
2.9M	/usr/libexec/cni/firewall
2.6M	/usr/libexec/cni/host-device
2.2M	/usr/libexec/cni/host-local
2.7M	/usr/libexec/cni/ipvlan
2.3M	/usr/libexec/cni/loopback
2.7M	/usr/libexec/cni/macvlan
2.5M	/usr/libexec/cni/portmap
2.9M	/usr/libexec/cni/ptp
2.4M	/usr/libexec/cni/sbr
2.0M	/usr/libexec/cni/static
2.4M	/usr/libexec/cni/tuning
2.7M	/usr/libexec/cni/vlan
2.5M	/usr/libexec/cni/vrf
45M	total
```
And this is the package upgraded to 1.1.1-r2 (rebuild on same system):
```
3.2M	/usr/libexec/cni/bandwidth
3.5M	/usr/libexec/cni/bridge
8.2M	/usr/libexec/cni/dhcp
3.6M	/usr/libexec/cni/firewall
3.2M	/usr/libexec/cni/host-device
2.7M	/usr/libexec/cni/host-local
3.3M	/usr/libexec/cni/ipvlan
2.9M	/usr/libexec/cni/loopback
3.4M	/usr/libexec/cni/macvlan
3.2M	/usr/libexec/cni/portmap
3.4M	/usr/libexec/cni/ptp
3.0M	/usr/libexec/cni/sbr
2.5M	/usr/libexec/cni/static
2.9M	/usr/libexec/cni/tuning
3.3M	/usr/libexec/cni/vlan
3.0M	/usr/libexec/cni/vrf
55M	total
```

When my whole system fits under 200MB, cni-plugins is now over 1/4 of it (and podman a good runner up); I'd like to see if this could be improved.

A naive idea that comes to mind is to work like busybox or coreutils configured with --enable-single-binary: bundle everything in a single binary, and have the program behaviour depend on argv[0].
[...]

PR #738, opened 2022-04-28
Build everything in a single binary

martinetd:
Hello,

I wanted to get a ballpark figure for myself so I took a naive stab at #737 without waiting: it converts all plugins from binaries to package, and adds a new command that just calls into the appropriate one based on os.Args[0].

From quick testing it works ok, and 'bin' directory size went from 64MB on my laptop to 12MB so I'm quite happy with the result.

current status:
 - this implements linux only, but no reason windows couldn't work the same as far as I know -- just need to figure out what can be used instead of symlink. I think hardlinks work?
 - this adds a new 'plugins' file in cni dir which might confuse something?
 - I didn't check if tests needs adjusmtent
 - the `plugins` binary itself might do something a bit more useful, like list available plugins or handle link creation like busybox does.

Now I'll really wait for feedback on this to finish it! Thanks.


martinetd:
(Sorry for the spam for people who might receive emails on build failures/consecutive pushes, it took me a while to figure out files suffixed by _windows.go get ignored by the go build system on linux)

I've updated this and marked as ready, but it still needs tests on windows: I've only checked with wine and hardlinks that go binaries behave properly with regard to argv, but I have no idea if that'll work in practice on ntfs. If links don't work just copying would work, the size for bin/win-{bridge,overlay} are actually almost the same but host-local will get bigger:
```
-rwxr-xr-x 1 martinet martinet 3.2M May  9 15:27 bin/host-local.exe*
-rwxr-xr-x 1 martinet martinet 8.1M May  9 15:27 bin/win-bridge.exe*
-rwxr-xr-x 1 martinet martinet 8.1M May  9 15:27 bin/win-overlay.exe*
->
-rwxr-xr-x 4 martinet martinet 8.2M May  9 15:28 bin/plugins.exe*
```


That aside it's just mostly matter of taste in what the 'launcher' actually does or binary names, I'm happy to change or discuss anything. Please comment!

martinetd:
@dcbw (sorry to ping directly - you replied in the issue) -- would you know who I should ping for reviews?

Since I need to rebuild cni-plugins anyway for an earlier PR that didn't get released yet I'm thinking of updating to this for our products, but I'd rather avoid straying from upstream for an undefined period of time.
I think this should work for everyone and considerably helps with size, but if there are fundamental problems or disagreements with it I'd rather not include it in the first place than have my users yell at me when I go back to upstream release and the system no longer fits in the tiny system partition.

martinetd:
I see podman's focus changed from cni-plugins to netavark, so I have been losing my time... (Would have appreciated being told!)

I won't be needing this anymore so closing this PR, but feel free to reopen and ping me if there is interest later down the road.

AkihiroSuda:
This PR still makes a lot of sense for containerd and its relevant projects.
Could you consider reopening? 🙏

kaiyou:
I might want to take another shot at this, which comes with an additional nice goodie: plugins as packages may be bundled in k8s distros directly.

Should we reopen this or should I open a separate PR?

martinetd:
I don't think I can reasonably give you push access to my branch, but I can't re-open this myself anyway: please open a new PR.

PR #740, opened 2022-04-29
ci: only rerun failed jobs on `/retest`

jpsim:
Unlike `rerun_url`, this endpoint isn't in the run API response, but we can construct it manually easily.

Documentation for it is here: https://docs.github.com/en/rest/actions/workflow-runs#re-run-failed-jobs-from-a-workflow-run


jpsim:
I hope you don't mind the drive-by PR, I was looking for a way to rerun GitHub Actions on PR comments and found this repo, then wanted to make use of the [newer partial re-runs](https://github.blog/2022-03-16-save-time-partial-re-runs-github-actions/) feature. This existing action setup was so helpful to me, the least I can do is contribute back this improvement.

squeed:
Thanks!

Issue #466, opened 2020-03-20
Plugin to add IP address to dummy interface


mdgreenfield:
We have a use case where we'd like to add IP 169.254.169.254 (the same IP used by cloud provider Metadata Services) to the network namespace of a container for an application running to bind to (i.e. we're running a fake Metadata Service as an injected sidecar).

I have developed a more generic plugin to do this which creates a `dummy` interface in the container's network namespace and then adds any configured IPs.

Per the [CONTRIBUTING docs](https://github.com/containernetworking/plugins/blob/master/CONTRIBUTING.md#3rd-party-plugins), if this is something that seems to fit with this repo I'd be happy to open a PR with that plugin.

PR #743, opened 2022-05-12
dummy: Create a Dummy CNI plugin that creates a virtual interface.

aristaMircea:
Leverages the Linux dummy interface type to create network interfaces
that permit routing packets through the Linux kernel without
them being transmitted.

This solution allows use of arbitrary non-loopback IP addresses within
the container, and leverages existing CNI IPAM plugins.

Related to #466

### DIFF
--- a/plugins/main/dummy/README.md
+++ b/plugins/main/dummy/README.md
@@ -1,0 +1,39 @@
+---
+title: dummy plugin
+description: "plugins/main/dummy/README.md"
+date: 2022-05-12
+toc: true
+draft: true
+weight: 200
+---
+
+## Overview
+
+dummy is a useful feature for routing packets through the Linux kernel without transmitting.
+
+Like loopback, it is a purely virtual interface that allows packets to be routed to a designated IP address. Unlike loopback, the IP address can be arbitrary and is not restricted to the `127.0.0.0/8` range.
+
+## Example configuration
+
+```json
+{
+	"name": "mynet",
+	"type": "dummy",
+	"ipam": {
+		"type": "host-local",
+		"subnet": "10.1.2.0/24"
+	}
+}
+```
+
+## Network configuration reference
+
+* `name` (string, required): the name of the network.
+* `type` (string, required): "dummy".
+* `ipam` (dictionary, required): IPAM configuration to be used for this network.
+
+## Notes
+
+* `dummy` does not transmit packets.
+Therefore the container will not be able to reach any external network.
+This solution is designed to be used in conjunction with other CNI plugins (e.g., `bridge`) to provide an internal non-loopback address for applications to use.
--- a/plugins/main/dummy/dummy.go
+++ b/plugins/main/dummy/dummy.go
@@ -1,0 +1,300 @@
+// Copyright 2022 Arista Networks
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"net"
+
+	"github.com/vishvananda/netlink"
+
+	"github.com/containernetworking/cni/pkg/skel"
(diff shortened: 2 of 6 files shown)


plugins/main/dummy/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+## Network configuration reference
+
+* `name` (string, required): the name of the network.
+* `type` (string, required): "dummy".
+* `ipam` (dictionary, required): IPAM configuration to be used for this network.
+
+## Notes
+
+* `dummy` does transmit packets.
```
mccv1r0, line 37: does not?

plugins/main/dummy/dummy.go
```diff
@@ -0,0 +1,309 @@
+// Copyright 2022 Arista Networks
```
mars1024, line 1: typo: Copyright

plugins/main/dummy/dummy.go
```diff
@@ -0,0 +1,309 @@
...
+			Namespace: netlink.NsFd(int(netns.Fd())),
+		},
+	}
+
+	if err := netlink.LinkAdd(dm); err != nil {
+		return nil, fmt.Errorf("failed to create dummy: %v", err)
+	}
+
+	err = netns.Do(func(_ ns.NetNS) error {
+		err := ip.RenameLink(tmpName, ifName)
```
squeed, line 64: Again, the rename probably isn't necessary.

plugins/main/dummy/README.md
```diff
@@ -0,0 +1,39 @@
...
+{
+	"name": "mynet",
+	"type": "dummy",
+	"ipam": {
+		"type": "host-local",
+		"subnet": "10.1.2.0/24"
+	}
+}
+```
+
```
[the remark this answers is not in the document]
aristaMircea: Is an Interface plugin a good place for these capabilities (`ips`, `ipRanges`)? Wouldn't they conflict with IPAM plugins, as they also take RunningConfig input?

plugins/main/dummy/dummy.go
```diff
@@ -0,0 +1,309 @@
...
+	return conf, nil
+}
+
+func createDummy(conf *types.NetConf, ifName string, netns ns.NetNS) (*current.Interface, error) {
+
+	dummy := &current.Interface{}
+
+	// due to kernel bug we have to create with tmpname or it might
+	// collide with the name on the host and error out
+	tmpName, err := ip.RandomVethName()
```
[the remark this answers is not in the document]
aristaMircea: Based on legacy code from other Interface Plugins. Removed.

plugins/main/dummy/dummy.go
aristaMircea: Same as above.

plugins/main/dummy/dummy.go
mccv1r0: I think you need something GPL/MIT/APACHE ?

aristaMircea: Apache license follows

aristaMircea:
Hi, any updates on this PR?

aristaMircea:
Still waiting on replies for remaining threads of this PR.

plugins/main/dummy/dummy.go
```diff
@@ -0,0 +1,309 @@
...
+
+	dm := &netlink.Dummy{
+		LinkAttrs: netlink.LinkAttrs{
+			Name:      tmpName,
+			Namespace: netlink.NsFd(int(netns.Fd())),
```
dcbw, line 55: @aristaMircea If we're creating the dummy interface inside the net namespace itself, then we don't need to rename the link at all. If there's already a link named `ifName` inside the NetNS then we should certainly fail.

plugins/main/dummy/dummy.go
dcbw: @aristaMircea as above, if we're creating the link inside the containers namespace, we don't need any rename. We can just create it with the name requested by the runtime (ifName) and error out if there's a failure.

plugins/main/dummy/dummy.go
```diff
@@ -0,0 +1,300 @@
...
+}
+
+func createDummy(conf *types.NetConf, ifName string, netns ns.NetNS) (*current.Interface, error) {
+
+	dummy := &current.Interface{}
+
+	dm := &netlink.Dummy{
+		LinkAttrs: netlink.LinkAttrs{
+			Name:      ifName,
+			Namespace: netlink.NsFd(int(netns.Fd())),
```
dcbw, line 51: @aristaMircea do we also need to add `Flags: net.FlagUp` here too, or a LinkSetUp() somewhere? Or does the kernel set it up automatically?

dcbw:
LGTM

PR #747, opened 2022-05-30
ipam/dhcp: Fix not adding type for provided client id

### LINKED ISSUE #746: dhcp: user provided client identifier is not with the additional type char
When given ipam config like this:

```
        "ipam": {
          "type": "dhcp",
          "provide": [
            {
              "option": "dhcp-client-identifier",
              "fromArg": "K8S_POD_NAME"
            }
          ]
        }
```

And start a pod named `abc` in test environment, following request is received by DHCP server:

```
... DHCP4_LEASE_ADVERT ... cid=[61:62:63] ...
```

This is incorrect and different from behavior before 4744ec27b89c083194e7df498de50f03a8a1d3ec. The correct cid should be:

```
... DHCP4_LEASE_ADVERT ... cid=[00:61:62:63] ...
```

Here `00` means "not a physical address", according to [RFC 2132](https://datatracker.ietf.org/doc/html/rfc2132#section-9.14).

SilverBut:
When a 'provide' config is added and an explicit value is set for 'dhcp-client-identifier', the value was not padded with byte 0x00. This commit fix it. Test is added to prevent future error like this.

Fix #746.

### DIFF
--- a/plugins/ipam/dhcp/lease.go
+++ b/plugins/ipam/dhcp/lease.go
@@ -202,10 +202,19 @@ func (l *DHCPLease) Stop() {
 
 func (l *DHCPLease) getOptionsWithClientId() dhcp4.Options {
 	opts := make(dhcp4.Options)
-	opts[dhcp4.OptionClientIdentifier] = []byte(l.clientID)
+
+	var clientIdentifier []byte
+	if val, ok := l.optsProviding[dhcp4.OptionClientIdentifier]; ok {
+		clientIdentifier = val
+	} else {
+		// if not set by config, use the default one set by AcquireLease
+		clientIdentifier = []byte(l.clientID)
+	}
+
 	// client identifier's first byte is "type"
-	newClientID := []byte{0}
-	newClientID = append(newClientID, opts[dhcp4.OptionClientIdentifier]...)
+	// 0 means the identifier is not an hardware address
+	// reference: https://datatracker.ietf.org/doc/html/rfc2132#section-9.14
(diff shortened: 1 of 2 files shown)


plugins/ipam/dhcp/lease.go
```diff
@@ -202,10 +202,19 @@ func (l *DHCPLease) Stop() {
...
+		// if not set by config, use the default one set by AcquireLease
+		clientIdentifier = []byte(l.clientID)
+	}
+
 	// client identifier's first byte is "type"
-	newClientID := []byte{0}
-	newClientID = append(newClientID, opts[dhcp4.OptionClientIdentifier]...)
+	// 0 means the identifier is not an hardware address
+	// reference: https://datatracker.ietf.org/doc/html/rfc2132#section-9.14
+	newClientID := append([]byte{0}, clientIdentifier...)
```
dcbw, line 217: @SilverBut do we want to allow passing the client ID type along with the client ID via the ProvideOptions in the IPAM config? eg, if you use the HostDevice plugin to move a NIC into the container, then want to run DHCP on it, perhaps you do want to use the MAC address as the client ID.

SilverBut: Sounds good but I'm worrying this type field will be misused by users, and setting a value not in [the registry](https://www.iana.org/assignments/arp-parameters/arp-parameters.xhtml#arp-parameters-2). 

Instead, maybe we can have an option like "useDeviceMacAsClientIdentifier"?

PR #766, opened 2022-08-25
host-device: fix cmdDel if device was renamed

olivier-matz-6wind:
When a device previously passed to the container has to be moved back to
the host, netlink.LinkByName() is used to find it in the container. This
does not work if the device was renamed in the container.

Fix this by storing the initial interface name in the alias, in addition
to the host interface name. To find the device, we now iterate the link
list until we find one whose alias corresponds.

Signed-off-by: Olivier Matz <olivier.matz@6wind.com>

### DIFF
--- a/plugins/main/host-device/host-device.go
+++ b/plugins/main/host-device/host-device.go
@@ -222,8 +222,9 @@ func moveLinkIn(hostDev netlink.Link, containerNs ns.NetNS, ifName string) (netl
 		if err = netlink.LinkSetDown(contDev); err != nil {
 			return fmt.Errorf("failed to set %q down: %v", hostDev.Attrs().Name, err)
 		}
-		// Save host device name into the container device's alias property
-		if err := netlink.LinkSetAlias(contDev, hostDev.Attrs().Name); err != nil {
+		// Save host device name and initial ifName into the container device's alias property
+		alias := fmt.Sprintf("host-device/%s/%s", hostDev.Attrs().Name, ifName)
+		if err := netlink.LinkSetAlias(contDev, alias); err != nil {
 			return fmt.Errorf("failed to set alias to %q: %v", hostDev.Attrs().Name, err)
 		}
 		// Rename container device to respect args.IfName
@@ -247,6 +248,28 @@ func moveLinkIn(hostDev netlink.Link, containerNs ns.NetNS, ifName string) (netl
 	return contDev, nil
 }
 
+func getLinkFromAlias(ifName string) (netlink.Link, string, error) {
+	links, err := netlink.LinkList()
+	if err != nil {
+		return nil, "", fmt.Errorf("failed to list links: %v", err)
+	}
+
+	for _, link := range links {
+		saved_info := strings.Split(link.Attrs().Alias, "/")
+		if len(saved_info) != 3 {
+			continue
+		}
+		magic := saved_info[0]
+		hostIfName := saved_info[1]
+		podIfName := saved_info[2]
+		if magic == "host-device" && podIfName == ifName {
+			return link, hostIfName, nil
+		}
+	}
+
+	return nil, "", fmt.Errorf("failed to find %q", ifName)
+}
+
 func moveLinkOut(containerNs ns.NetNS, ifName string) error {
 	defaultNs, err := ns.GetCurrentNS()
 	if err != nil {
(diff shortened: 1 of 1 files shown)


olivier-matz-6wind:
Hi,

A more reliable way would probably be to store the ifindex of the interface instead of storing data in the ifalias field. Indeed, like the name, the ifalias can still be modified in the container, even it it seems uncommon. However, the ifindex never changes. I'd be happy to implement this, but I don't know how/if we can store information between cmdAdd and cmdDel. Any advice?

An alternative to ifalias is a kernel feature called "alternative name" (https://lwn.net/Articles/800927/), but it is only available in kernels >= 5.5, so it seems too recent.

One more question is about what to do when we cannot find the device. In my use-case, it blocks the destruction of the pod (when I'm using kubectl scale deployment my-pod --replicas=0) with the following error:

```
error killing pod: failed to "KillPodSandbox" for "90accdea-b2d9-4f1f-a9fe-8c838defd726" with KillPodSandboxError: "rpc error: code = Unknown desc = failed to destroy network for sandbox \"3ae636bf65be936512620499bca54789767f874d4cfd16497fc91a9fba02a6a2\": delegateDel: error invoking DelegateDel - \"host-device\": error in getting result from DelNetwork: failed to find \"net1\""
```

Would it make sense to ignore the error and just issue a warning? (how?) Indeed, in Linux, when a netns is destroyed, all physical devices that were present in the netns will respawn in init_net.

Thanks,
Olivier

dcbw:
@olivier-matz-6wind what is renaming the network device inside the container, away from the name that was specified in the CNI configuration (eg, net0 or eth0)?

eg, we'd expect that if the host-device plugin renamed the link to `net0` inside the container as requested by the caller, the link would still have that name at DEL time. If it doesn't, there's not much this plugin can or should do; something reconfigured the link after the plugin in fundamental ways.

olivier-matz-6wind:
Hi @dcbw,

> what is renaming the network device inside the container, away from the name that was specified in the CNI configuration (eg, net0 or eth0)?

We are running a virtual router inside the pod. This software is able to configure the network interfaces passed by the host: add an IP address, name the interface, create virtual interfaces like vlans on top of it, ...

> eg, we'd expect that if the host-device plugin renamed the link to net0 inside the container as requested by the caller, the link would still have that name at DEL time. If it doesn't, there's not much this plugin can or should do; something reconfigured the link after the plugin in fundamental ways.

Yes, I agree there are situations where the host won't be able to do anything. So maybe this pull request does not go in the right direction.

My problem is that when the host-device plugin fails to find the device (either because it was renamed, because the alias changed, or because it was moved in another netns), CmdDel() returns an error, and this error prevents the pod to be removed when I'm scaling down my deployment.

The following draft patch fixes the issue:

```
--- a/plugins/main/host-device/host-device.go
+++ b/plugins/main/host-device/host-device.go
@@ -257,7 +257,7 @@ func moveLinkOut(containerNs ns.NetNS, ifName string) error {
        return containerNs.Do(func(_ ns.NetNS) error {
                dev, err := netlink.LinkByName(ifName)
                if err != nil {
-                       return fmt.Errorf("failed to find %q: %v", ifName, err)
+                       return nil
                }
 
                // Devices can be renamed only when down

```

The question is: is this patch (or similar) acceptable? Would it make sense to simply log something in this case? In fact if the netdevice is not moved back to the host by the plugin, it will respawn in init_net when the pod netns will be deleted.

dcbw:
> But as we said previously, there are still corner cases where the CNI cannot do anything to restore the interface, and this would still block the downscaling of the container (and this prevents resources from being reused). Is it a problem to let the kernel respawn the netdevice in init_net when the container netns is deleted?

I know the kernel will do that for "physical" devices, but I didn't think it did it for virtual devices like macvlan, veth, etc. I think we were trying to keep the plugin consistent in it's device treatment so it would handle both.

There's always going to be corner-cases with the current architecture; we're trying to plan out what "CNI 2.0" would look like and it's likely going to include some kind of reconciliation capability that would allow more comprehensive cleanup.

Also, even if the kernel did return the device to the main netns, it wouldn't have the same name it did originally, right?

olivier-matz-6wind:
> > But as we said previously, there are still corner cases where the CNI cannot do anything to restore the interface, and this would still block the downscaling of the container (and this prevents resources from being reused). Is it a problem to let the kernel respawn the netdevice in init_net when the container netns is deleted?
> 
> I know the kernel will do that for "physical" devices, but I didn't think it did it for virtual devices like macvlan, veth, etc. I think we were trying to keep the plugin consistent in it's device treatment so it would handle both.

Correct, in case of virtual device, they will be deleted with the netns.

> Also, even if the kernel did return the device to the main netns, it wouldn't have the same name it did originally, right?

Correct, but it is something that could be handled in the plugin. I was thinking that it was still better than blocking.


I'll try to think again to see if I can find a better solution. Else, I'll go with the /var/run/cni/<network> solution.

Many thanks for the feedback.

PR #768, opened 2022-09-06
Cleanup Socket and Pidfile on exit

### LINKED ISSUE #760: Delete  `/run/cni/dhcp.sock` when `dhcp daemon` is stopped
I start `/opt/cni/bin/dhcp daemon` as described at https://www.cni.dev/plugins/current/ipam/dhcp/ .  The file-socket `/run/cni/dhcp.sock` is created by the process .  I stop the process with Ctrl+C.  The file-socket `/run/cni/dhcp.sock` is still there. I call again `/opt/cni/bin/dhcp daemon`, which prints:

> 2022/07/24 17:56:32 Error getting listener: listen unix /run/cni/dhcp.sock: bind: address already in use

`lsof` shows that `/run/cni/dhcp.sock` is not in use.  That said, on Ctrl+C `dhcp daemon` shall delete `/run/cni/dhcp.sock` which it created on start up, or reuse on startup.  No normal process under linux creates files on startup, which it does not clean by itself, and refuses to start the next time, if the files are present

EmilyShepherd:
These were previously left intact, even when exiting gracefully. As the
daemon also fails if the socket already exists, it became the caller's
responsibilityto check for and cleanup old socket files when performing
graceful / deliberate restarts.

This fixes #760


mccv1r0 (approved):

dcbw:
LGTM too. Thanks!

PR #769, opened 2022-09-06
host-local: remove unused Release(ip) from type Store interface

jingyuanliang:
This function seems unused and host-local compiles perfectly without it.

### DIFF
--- a/plugins/ipam/host-local/backend/disk/backend.go
+++ b/plugins/ipam/host-local/backend/disk/backend.go
@@ -94,10 +94,6 @@ func (s *Store) LastReservedIP(rangeID string) (net.IP, error) {
 	return net.ParseIP(string(data)), nil
 }
 
-func (s *Store) Release(ip net.IP) error {
-	return os.Remove(GetEscapedPath(s.dataDir, ip.String()))
-}
-
 func (s *Store) FindByKey(id string, ifname string, match string) (bool, error) {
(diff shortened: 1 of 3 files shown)


dcbw:
The only downside is that I know some people vendor host-local into their own plugins... and might use Release().

But not sure if that's something we care about a ton...

jingyuanliang:
I've discovered this when I was customizing host-local with the need to hook on the releasing code, and found that there's an unused Release.

I was thinking if I should attach my code to both Release and ReleaseByID - if I do so, it's extra work including the tests needed, and the work won't be used anyway (and it increases the chance of conflicts when syncing from upstream); if I don't do so, it attracts the next colleague to call the incomplete (without my extra logic) Release if they happen to need it, only to find it not working as intended potentially after long debugging.

I thought it's probably the best to remove the unused Release at least in my customization, then thought it's probably true that having the unused Release removed from upstream is also a good thing. I can think of a couple reasons:

- Resolve the issue I'm facing above for other people customizing host-local in a similar way
- Avoid confusion when someone tries to customize host-local and attaches their logic to Release, only to find it not working
- If someone happens to need this Release in their customization, they probably need to implement more and will need to customize it anyway. Considering the current logic is trivial, it's not hard to add it back as part of their customization.

squeed (approved):

PR #775, opened 2022-09-28
build: Remove dos2unix dependency

sebsoto:
Removes use of dos2unix when building and testing for Windows. This is unnecessary and is adding an additional dependency to build systems.

Signed-off-by: Sebastian Soto <ssoto@redhat.com>


sebsoto:
@dcbw Can you approve this for testing?

sebsoto:
Closing this for now, will see what else can be done as this works when building for windows on linux but not for windows or windows

PR #779, opened 2022-11-04
Fix path substitution to enable setting sysctls on vlan interfaces

mmirecki:
This commit changes the order of substituting sysctl path to first handle . to / change, before substituting the interface name. This is needed as vlan interfaces have a . in the name, which should not be changed.


mars1024 (commented):
/lgtm, do you mind adding some tests for this? @mmirecki

mmirecki:
> /lgtm, do you mind adding some tests for this? @mmirecki

To test it properly it would have to be tested with namespaces sysctl's, which is currently not doable with the current test framwork. Currently we test the sysctl's with "net.ipv4.conf.**all**" which would work here.

squeed:
@mmirecki this is a potential directory traversal issue, since we don't actually validate that IFNAME is sane. Can you check to make sure that IFNAME doesn't contain any "/"? 

We should also add a check to ensure that the resolved path doesn't escape `/proc/sys`.

mmirecki:
> @mmirecki this is a potential directory traversal issue, since we don't actually validate that IFNAME is sane. Can you check to make sure that IFNAME doesn't contain any "/"?
> 
> We should also add a check to ensure that the resolved path doesn't escape `/proc/sys`.

Added validateArgs to validate the interface name does not contain any path separators.

mmirecki:
@squeed could you please look again? Added the requested validation

squeed (approved):

PR #784, opened 2022-11-17
Tap plugin

mmirecki:
This PR adds a plugin to create tap devices.
The plugin adds a tap device to the container.
    
The plugin has a workaround for a golang netlink library
which does not allow for tap devices with no owner/group
to be created. When no tap owner/group is requested, the
plugin will fall back to using the ip tool for creating
the tap device. A fix to the golang netlink lib is pending: 
https://github.com/vishvananda/netlink/pull/835


plugins/main/tap/tap.go
```diff
@@ -0,0 +1,464 @@
...
+	}
+
+	output, err := exec.Command("ip", tapDeviceArgs...).CombinedOutput()
+	if err != nil {
+		return fmt.Errorf("failed to run command %s: %v", output, err)
+	}
+
+	if err := selinux.SetExecLabel("system_u:system_r:container_t:s0"); err != nil {
+		return fmt.Errorf("failed set socket label: %v", err)
+	}
```
maiqueb, line 135: isn't this duplicated ?

plugins/main/tap/tap.go
```diff
@@ -0,0 +1,464 @@
...
+
+	tapDeviceArgs := []string{"tuntap", "add", "mode", "tap", "name", tmpName}
+	if conf.MultiQueue {
+		tapDeviceArgs = append(tapDeviceArgs, "multi_queue")
+	}
+
+	output, err := exec.Command("ip", tapDeviceArgs...).CombinedOutput()
+	if err != nil {
+		return fmt.Errorf("failed to run command %s: %v", output, err)
+	}
```
maiqueb, line 131: I'm unsure how serious this PR is; we shouldn't - IMO - use the ip tool, but either use https://github.com/songgao/water, or use syscalls to create the tap. 

(I don't remember if netlink allowed for tap creation or if I'm confusing it w/ macvtap).

plugins/main/tap/tap.go
SchSeba: I don't think it duplicated we need this to run the second time the ip link command to set the MTU and mac address if needed

plugins/main/tap/tap.go
SchSeba: it's possible to create with netlink you can see that we do it when we are not running on selinux system.

and we use the ip command because we need to start a second binary so the setExecLabel will be applied to them.

another solution will be to create a different binary with netlink but that will make the testing (you need to have the binary compiled) and the ship (another binary in the CNI repo with is just a helper) more complicated.

but I think we are open to suggestions

maiqueb: I see.

I'd still vote for not using the ip tool despite its higher cost of having to ship the tool . 

For the sake of the discussion this is fine.

mmirecki: We need to use a new process to make use of the new label.
One option is to use some external one like 'ip' (which we assume will be on the nodes). Another option would be for the plugin to call itself again (with the new label) and use the golang libs for this (I assume netlink would do in this case).

maiqueb: > We need to use a new process to make use of the new label.

I'm aware you need a new process. I'm just saying you don't need IP for that. I think using a smaller binary with a netlink implementation that creates a tap device setting the owner / group / mtu would be enough.

mmirecki: Changed this to use a binary with netlink.
To avoid having multiple binaries the plugin binary will be used for that.

dougbtv:
cc: @mccv1r0 too

plugins/main/tap/tap.go
dougbtv: I like where this landed, thanks for the discussion, I really like the bin calling itself with the proper selinux, seems really all-in-one.

dougbtv:
Conceptually it seems good to me!

plugins/main/tap/tap.go
```diff
@@ -0,0 +1,472 @@
+// Copyright 2015 CNI authors
```
s1061123, line 1: Fix: 2015 -> 2022 (this year)

plugins/main/tap/tap.go
```diff
@@ -0,0 +1,472 @@
...
+	MultiQueue bool   `json:"multiQueue"`
+	MTU        int    `json:"mtu"`
+	Mac        string `json:"mac,omitempty"`
+
+	RuntimeConfig struct {
+		Mac string `json:"mac,omitempty"`
+	} `json:"runtimeConfig,omitempty"`
+}
+
+// MacEnvArgs represents CNI_ARG
```
s1061123, line 58: If you provide CNI_ARG, how about to add ['args in network config'](https://github.com/containernetworking/cni/blob/main/CONVENTIONS.md#args-in-network-config)? 

Example: 
https://github.com/containernetworking/plugins/blob/7e9ada51e751740541969e1ea5a803cbf45adcf3/plugins/meta/tuning/tuning.go#L61-L63

plugins/main/tap/tap.go
```diff
@@ -0,0 +1,472 @@
...
+	if n.RuntimeConfig.Mac != "" {
+		n.Mac = n.RuntimeConfig.Mac
+	}
+
+	return n, n.CNIVersion, nil
+}
+
+func CloseOnExec(fd int) {
+	syscall.CloseOnExec(fd)
+}
```
s1061123, line 97: This function is only used this file, so you can make not capital letter function (to limit the scope).
In addition, btw, why you create function for one-liner?

plugins/main/tap/tap.go
```diff
@@ -0,0 +1,472 @@
...
+
+func setContainerSeBool() error {
+	output, err := exec.Command("getsebool", "container_use_devices").CombinedOutput()
```
s1061123, line 100: Did you check this plugin without selinux? (i.e. linux machine without getsebool command)

plugins/main/tap/tap.go
```diff
@@ -0,0 +1,472 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to run getsebool command %s: %v", string(output), err)
+	}
+	if strings.Contains(string(output), "off") {
+		output, err := exec.Command("setsebool", "-P", "container_use_devices", "true").CombinedOutput()
```
s1061123, line 105: Did you check this plugin without selinux? (i.e. linux machine without setsebool command)

plugins/main/tap/tap.go
```diff
@@ -0,0 +1,472 @@
...
+			return nil
+		}
+		return err
+	}
+
+	return err
+}
+
+func main() {
+	if len(os.Args) > 2 && os.Args[1] == AddLink {
```
s1061123, line 390: May need to add comment that this plugin may be invoked with args (it is not usual in common CNI plugin).

s1061123 (changes requested):

plugins/main/tap/tap_suite_test.go
```diff
@@ -0,0 +1,27 @@
+// Copyright 2017 CNI authors
```
s1061123, line 1: 2017->2022

plugins/main/tap/tap_test.go
```diff
@@ -0,0 +1,331 @@
+// Copyright 2022 Arista Networks
```
s1061123, line 1: could you please double check that it is correct copyright or not.

mmirecki: Changed to: 2022 CNI authors
Looking at the other files, looks like this is correct

plugins/main/tap/tap.go
mmirecki: Yes, works fine
Node that we have a condition before we call getsebool or setsebool:

```
	if selinux.EnforceMode() != -1 {
		err = setContainerSeBool()
```

plugins/main/tap/tap.go
mmirecki: This is here as several other plugins go this way (allowing to override the mac).
I'm not even sure we need this at all, we can just take the mac from netconf.

plugins/main/tap/tap.go
mmirecki: Sounds a bit like a premature optimization. This is a workaround for a selinux issue (missing label on the plugin binary), so it might not be even needed or be completely different with other security modules.

maiqueb: I'm unsure ... the complexity of this PR is - IMO - keeping SELinux happy. 

Shoving those details in the face of the reader (who might not care about it) is something we should avoid imo. Thus I propose to abstract that, and hide it in a separate package.

s1061123: @mmirecki thank you for the clarification. In my opinion, I suppose we don't require to have interface abstraction because this code is not library and it is not reused other code for now.
@maiqueb is there any detailed plan to reuse this code, for example, you will soon create some other CNI plugin which uses selinux?

s1061123: I prefer that the previous one, without module, the reasons are:

1) Currently only tap plugin may require this mechanism (no other consumer)
2) (At least to me,) there is no other security module which may applied to the abstraction

In addition, if we really required such abstraction, then we should put it in more common place (not CNI).

plugins/main/tap/tap.go
mmirecki: Added a comment.
I'll also describe this in cni.dev in the plugin documentation.

EmilyShepherd:
NB: The test failure on this PR appears to be unrelated to these changes.

Please see https://github.com/containernetworking/plugins/pull/793 for a potential fix.

plugins/main/tap/tap.go
maiqueb: > @mmirecki thank you for the clarification. In my opinion, I suppose we don't require to have interface abstraction because this code is not library and it is not reused other code for now. @maiqueb is there any detailed plan to reuse this code, for example, you will soon create some other CNI plugin which uses selinux?

No. I just think code readability is important on any project, not only for libraries / reused code.

The code becomes a lot more readable if anything related to selinux is moved elsewhere - since the reader can choose to focus on those particular details only if desired.

I won't insist further on this.

mmirecki:
> NB: The test failure on this PR appears to be unrelated to these changes.
> 
> Please see #793 for a potential fix.

Thanks for the tip
Added the fix on top of my changes. Will remove it if the fix gets in first.

SchSeba (commented):
nice work!

two small nits

plugins/meta/portmap/portmap_integ_test.go
```diff
@@ -207,7 +207,7 @@ var _ = Describe("portmap integration tests", func() {
 
 				// Verify iptables rules are gone
 				_, err = ipt.List("nat", dnatChainName)
-				Expect(err).To(MatchError(ContainSubstring("iptables: No chain/target/match by that name.")))
+				Expect(err).To(HaveOccurred())
```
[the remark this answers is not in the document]
mmirecki: This change is in a separate commit, to make the tests pass.
Once https://github.com/containernetworking/plugins/pull/793 is merged this commit can be removed.

plugins/main/tap/tap.go
mccv1r0: > we are not running on selinux system

CNI plugins might run on systems that don't use selinux.

plugins/main/tap/tap.go
mmirecki: An alternative version with the selinux stuff moved to a separate package.
Please advise which you prefer.

plugins/main/tap/tap.go
mmirecki: This is fine. The check for selinux genenforce function will just return false, and we'll proceed with the default.

dougbtv (approved):

plugins/main/tap/tap.go
```diff
@@ -0,0 +1,415 @@
...
+	}
+	if mac != "" {
+		addr, err := net.ParseMAC(mac)
+		if err != nil {
+			return fmt.Errorf("invalid args %v for MAC addr: %v", mac, err)
+		}
+		linkAttrs.HardwareAddr = addr
+	}
+	if multique {
+		mv.Flags = netlink.TUNTAP_MULTI_QUEUE_DEFAULTS | netlink.TUNTAP_VNET_HDR | unix.IFF_TAP
```
maiqueb, line 117: Could you explain why the `unix.IFF_TAP` flag is required ? 

I guess I'm confused why that's only required by taps w/ multi-queue.

plugins/main/tap/tap.go
```diff
@@ -0,0 +1,415 @@
...
+	}
+
+	if n.RuntimeConfig.Mac != "" {
+		n.Mac = n.RuntimeConfig.Mac
+	}
+
+	return n, n.CNIVersion, nil
+}
+
+func createLink(tmpName string, mtu int, nsFd int, multique bool, mac string) error {
```
maiqueb, line 97: nit: the boolean argument name - `multique` has a typo.

furthermore, I don't think you should use boolean input arguments. Can't you receive the flags directly ? i.e. have two consts (or vars, if you can't use const), one for the multi-queue flags, another for the regular flags.

That makes the code more readable.

plugins/main/tap/tap.go
```diff
@@ -0,0 +1,415 @@
...
+				return err
+			}
+			f, err := plugin.Lookup("CreateLink")
+			if err != nil {
+				return err
+			}
+			err = f.(func(tmpName string, mtu int, nsFd int, nsPath string, multique bool, mac string)(error))(tmpName, conf.MTU, int(netns.Fd()), netns.Path(), conf.MultiQueue, conf.Mac)
+			if err != nil {
+				return err
+			}
```
s1061123, line 148: This code does not check that 'CreateLink' has right function type (i.e. argument type and name). How about to add check the lookuped symbol has right function type as following sample code?

https://github.com/vladimirvivien/go-plugin-example/blob/070e14c627a7724f7ebe5c23f672ccb9181d39be/greeter.go#L40-L58

plugins/main/tap/tap.go
```diff
@@ -0,0 +1,415 @@
...
+)
+
+const AddLink = "AddLink"
+
+type NetConf struct {
+	types.NetConf
+	MultiQueue bool   `json:"multiQueue"`
+	MTU        int    `json:"mtu"`
+	Mac        string `json:"mac,omitempty"`
+	Tapplugin  string `json:"tapplugin"`
```
maiqueb, line 47:
```suggestion
	TapPlugin  string `json:"tapPlugin"`
```

golang uses [mixed caps](https://go.dev/doc/effective_go#mixed-caps).

maiqueb (commented):
I think it would make sense to also have a short .md readme explaining how to use / what's the meaning of the CNI configuration.

MikeZappa87:
Can you also draft up the docs in the cni.dev repo?

plugins/main/tap/tap.go
mmirecki: fixed

plugins/main/tap/tap.go
mmirecki: fixed

plugins/main/tap/tap.go
mmirecki: fixed

plugins/main/tap/distro/rhel/rhel.go
```diff
@@ -0,0 +1,87 @@
...
+			return fmt.Errorf("failed to run setsebool command %s: %v", string(output), err)
+		}
+	}
+	return nil
+}
+
+// Due to issues with the vishvananda/netlink library (fix pending) this method is using the ip tool to set up
+// the tap device.
+func createSelinuxTap(tmpName string, mtu int, nsFd int, nsPath string, multiqueue bool, mac string, owner int, group int) error {
+	if err := selinux.SetExecLabel("system_u:system_r:container_t:s0"); err != nil {
```
dcbw, line 47: @mmirecki couldn't this just be a config file option, such that each user could customize it for their specific distro if it differs between distros/policies?

plugins/main/tap/distro/rhel/rhel.go
```diff
@@ -0,0 +1,87 @@
...
+	}
+	err = createSelinuxTap(tmpName, mtu, nsFd, nsPath, multique, mac, owner, group)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func setContainerSeBool() error {
+	output, err := exec.Command("cat", "/sys/fs/selinux/booleans/container_use_devices").CombinedOutput()
```
dcbw, line 31: @mmirecki does this come from the selinux userland utilities? Is it common across all distros that use selinux, or is it RHEL specific?

dcbw: Also, can't we just read this file with `ioutil.ReadFile()` or something? Why do we need to shell out to `cat`?

mmirecki: @dbcw - Sebools are a selinux concept, so would be there on any distro with selinux
As to the mount point for selinuxfs it is /sys/fs/selinux on RHEL, I have no experience with selinux on other distro's (and the docs are not quite verbose on this).

Changed to os.ReadFIle()

plugins/main/tap/distro/rhel/rhel.go
mmirecki: Done

squeed:
Summarizing some discussion we had in the maintainer's meeting:

- we can't be distro specific
- the SELinux labels should just be a configuration option; the maintainer can supply it.
- due to an [issue in netlink](https://github.com/vishvananda/netlink/pull/835/) you currently cannot specify file-less TAPs, so we need to fallback to the `ip` tool
- Just like the SELinux label, the uid and gid of the tap device should be a `*int32` in the plugin configuration

plugins/main/tap/distro/rhel/rhel.go
maiqueb: @mmirecki does this play along w/ MCS (Multi-Category Security) ? IIUC, if the exec labels on a pod feature MCS, they won't be able to open a generic `container_t` label.

mmirecki: @maiqueb I'm not an expert on MCS, but note that this code is not executed in a pod. The CNI plugin only operates on the pod's network namespace.

maiqueb: yes, but it ends up being the label for that particular container; consider for instance KubeVirt launcher pods (these encapsulate the VM): these will have the following label: `system_u:object_r:container_t:s0:c30,c157` . 

This would cause the processes running in this container to fail accessing resources with the `container_t` label.

AFAIU, this is in place so one container cannot access resources of **another** container.

If we'd just used the `container_t` label, any container (running without MCS, thus running w/ the `container_t` label) would be able to consume the taps of any other container - something which I think we don't want to allow.

mmirecki:
> Summarizing some discussion we had in the maintainer's meeting:
> 
> * we can't be distro specific
> * the SELinux labels should just be a configuration option; the maintainer can supply it.
> * due to an [issue in netlink](https://github.com/vishvananda/netlink/pull/835/) you currently cannot specify file-less TAPs, so we need to fallback to the `ip` tool
> * Just like the SELinux label, the uid and gid of the tap device should be a `*int32` in the plugin configuration

@squeed Could you please take a look again?

SchSeba (commented):
small comments

plugins/main/tap/tap.go
```diff
@@ -0,0 +1,454 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"github.com/opencontainers/selinux/go-selinux"
```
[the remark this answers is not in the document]
mmirecki: Done

plugins/main/tap/tap.go
```diff
@@ -0,0 +1,454 @@
...
+	}
+
+	if n.RuntimeConfig.Mac != "" {
+		n.Mac = n.RuntimeConfig.Mac
+	}
+
+	return n, n.CNIVersion, nil
+}
+
+// Due to issues with the vishvananda/netlink library (fix pending) it is not possible to create an ownerless/groupless
```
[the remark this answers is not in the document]
mmirecki: Done

plugins/main/tap/tap.go
```diff
@@ -0,0 +1,454 @@
...
+)
+
+type NetConf struct {
+	types.NetConf
+	MultiQueue     bool    `json:"multiQueue"`
+	MTU            int     `json:"mtu"`
+	Mac            string  `json:"mac,omitempty"`
+	Owner          *uint32 `json:"owner,omitempty"`
+	Group          *uint32 `json:"group,omitempty"`
+	SelinuxContext string  `json:"selinuxcontext"`
```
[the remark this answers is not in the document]
mmirecki: Done

plugins/main/tap/distro/rhel/rhel.go
mmirecki: The custom label would allow the user to either use the label and resolve this in the plugin, or resolve it outside the plugin.
I'm not an expert on the MCS, but from what I read it looks like it's not widely used. I think it should be enough to support the targeted mode for selinux (or have a custom label for mcs?)

maiqueb: If possible, we'd love to use this CNI from KubeVirt.

Thing is, we use MCS...

I'm just saying that this approach (user provided label) will quite probably be impossible to use for MCS-aware workloads. @jean-edouard can you keep me honest here ?

Having said that, it could be OK to move forward with this proposal now, and investigate how to account for MCS later on - I think we may be able to pass the MCS part of the string via CNI_ARGS or runtime config.

plugins/main/tap/tap.go
```diff
@@ -0,0 +1,456 @@
...
+	types.NetConf
+	MultiQueue     bool    `json:"multiQueue"`
+	MTU            int     `json:"mtu"`
+	Mac            string  `json:"mac,omitempty"`
+	Owner          *uint32 `json:"owner,omitempty"`
+	Group          *uint32 `json:"group,omitempty"`
+	SelinuxContext string  `json:"selinuxContext,omitempty"`
+	Args           struct {
+		Mac string `json:"mac,omitempty"`
+	} `json:"args,omitempty"`
```
s1061123, line 54: Thank you for adding 'args'. In addition, how about to add this args under cni, as `tuning`, `host-local`, `static` and others does, to make args format consistent among this repo?

https://github.com/containernetworking/plugins/blob/c4d24e80d64393d2c632a825a3486d1c2c0248ec/plugins/ipam/static/main.go#L40-L42

mmirecki: Yes, but not in this PR.

plugins/main/tap/distro/rhel/rhel.go
mmirecki: @maiqueb Could you please suggest a way to also satisfy the MCS way?

maiqueb: After speaking w/ @jean-edouard I was corrected on my understanding of MCS:
- a container using an MCS-aware label - i.e. `container_t:s0:c1,c2` **can** access a `container_t:s0` resource.

The only issue here would be the security risk of **any** container being able to consume taps created by another container - not optimal.

I think we can move forward with this implementation from my perspective.

We can tackle the MCS functionality as a follow-up - I think we can pass the label via runtime config / CNI args to the plugin.

plugins/main/tap/tap.go
s1061123: I'm not clear what you said. As far as I see your PR, you have Args as follows:

```
	Args           struct {
		Mac string `json:"mac,omitempty"`
	} `json:"args,omitempty"`
```

I just propose it as following. Adding 'cni', to consistent args format among the plugins.

```
	Args           struct {
                A * struct {
		   Mac string `json:"mac,omitempty"`
                } `json:"cni,omitempty"`
	} `json:"args,omitempty"`
```

s1061123 (changes requested):

plugins/main/tap/tap_suite_test.go
```diff
@@ -0,0 +1,27 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+
+	"testing"
```
maiqueb, line 21: nit: shouldn't this feat the opposite order (first the golang dependencies, then the testing pkgs) ?

plugins/main/tap/tap_test.go
```diff
@@ -0,0 +1,316 @@
...
+	types020 "github.com/containernetworking/cni/pkg/types/020"
+	types040 "github.com/containernetworking/cni/pkg/types/040"
+	types100 "github.com/containernetworking/cni/pkg/types/100"
+	"github.com/containernetworking/plugins/pkg/ns"
+	"github.com/containernetworking/plugins/pkg/testutils"
+	"github.com/containernetworking/plugins/plugins/ipam/host-local/backend/allocator"
+	"github.com/vishvananda/netlink"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
```
maiqueb, line 37: nit: I would move this up. Something like:

```golang
import (
	"encoding/json"
	"fmt"
	"io/ioutil"
	"net"
	"os"
	"strings"
	"syscall"

	. "github.com/onsi/ginkgo"
	. "github.com/onsi/gomega"

	"github.com/vishvananda/netlink"

	"github.com/containernetworking/cni/pkg/skel"
	"github.com/containernetworking/cni/pkg/types"
	types020 "github.com/containernetworking/cni/pkg/types/020"
	types040 "github.com/containernetworking/cni/pkg/types/040"
	types100 "github.com/containernetworking/cni/pkg/types/100"
	"github.com/containernetworking/plugins/pkg/ns"
	"github.com/containernetworking/plugins/pkg/testutils"
	"github.com/containernetworking/plugins/plugins/ipam/host-local/backend/allocator"
)
```

plugins/main/tap/tap_test.go
```diff
@@ -0,0 +1,316 @@
...
+)
+
+const TYPETAP = "tuntap"
+const IFNAME = "tapifc"
```
maiqueb, line 41: nit

```golang

```
```suggestion
const (
    IFNAME = "tapifc"
    TYPETAP = "tuntap"
)
```

plugins/main/tap/tap_test.go
```diff
@@ -0,0 +1,316 @@
...
+
+type Net struct {
+	Name          string                 `json:"name"`
+	CNIVersion    string                 `json:"cniVersion"`
+	Type          string                 `json:"type,omitempty"`
+	IPAM          *allocator.IPAMConfig  `json:"ipam"`
+	RawPrevResult map[string]interface{} `json:"prevResult,omitempty"`
+	PrevResult    types100.Result        `json:"-"`
+}
```
maiqueb, line 50: Hm, can't we just re-use libcni's `NetConf` [struct](https://github.com/containernetworking/cni/blob/f89a8c6a3e6fb2e9a78926f453993347d73bce42/pkg/types/types.go#L60) ?

plugins/main/tap/tap_test.go
```diff
@@ -0,0 +1,316 @@
...
+
+	AfterEach(func() {
+		Expect(os.RemoveAll(dataDir)).To(Succeed())
+		Expect(originalNS.Close()).To(Succeed())
+		Expect(testutils.UnmountNS(originalNS)).To(Succeed())
+		Expect(targetNS.Close()).To(Succeed())
+		Expect(testutils.UnmountNS(targetNS)).To(Succeed())
+	})
+
+	for _, ver := range testutils.AllSpecVersions {
```
maiqueb, line 191: I think using a [table](https://pkg.go.dev/github.com/onsi/ginkgo/extensions/table) would be more idiomatic.

plugins/main/tap/tap_test.go
```diff
@@ -0,0 +1,316 @@
...
+				    "name": "tapTest",
+				    "type": "tap",
+				    "owner": 0,
+				    "group": 0,
+				    "ipam": {
+						"type": "host-local",
+						"subnet": "10.1.2.0/24",
+						"dataDir": "%s"
+				    }
+				}`, ver, dataDir)
```
maiqueb, line 206: Please consider using an helper here.

plugins/main/tap/tap_test.go
```diff
@@ -0,0 +1,316 @@
...
+			By("Make sure the tap link exists in the target namespace")
+			err = targetNS.Do(func(ns.NetNS) error {
+				defer GinkgoRecover()
+
+				link, err := netlink.LinkByName(IFNAME)
+				Expect(err).NotTo(HaveOccurred())
+				Expect(link.Attrs().Name).To(Equal(IFNAME))
+				Expect(link.Type()).To(Equal(TYPETAP))
+
+				if macAddress != "" {
```
maiqueb, line 242: Hm, shouldn't this actually be an error if it happens on CNI >= 0.3.0 (since - AFAIU - it is expected for a MAC to have been reported) ?

plugins/main/tap/tap_test.go
```diff
@@ -0,0 +1,316 @@
...
+					hwaddr, err := net.ParseMAC(macAddress)
+					Expect(err).NotTo(HaveOccurred())
+					Expect(link.Attrs().HardwareAddr).To(Equal(hwaddr))
+				}
+				addrs, err := netlink.AddrList(link, syscall.AF_INET)
+				Expect(err).NotTo(HaveOccurred())
+				Expect(len(addrs)).To(Equal(1))
```
maiqueb, line 249:
```suggestion
				Expect(addrs).To(HaveLen(1))
```

It is more idiomatic, and the print when it fails is more detailed.

plugins/main/tap/tap_test.go
```diff
@@ -0,0 +1,316 @@
...
+		Expect(originalNS.Close()).To(Succeed())
+		Expect(testutils.UnmountNS(originalNS)).To(Succeed())
+		Expect(targetNS.Close()).To(Succeed())
+		Expect(testutils.UnmountNS(targetNS)).To(Succeed())
+	})
+
+	for _, ver := range testutils.AllSpecVersions {
+		ver := ver
+
+		It(fmt.Sprintf("[%s] add, check and remove a tap device run correctly", ver), func() {
```
maiqueb, line 194: I haven't checked out the other plugins are tested, so maybe what my proposal is not aligned with the project.

I apologize if that's the case. 

While I think the coverage added by this test is quite thorough, I would see smaller tests: one table for add / another for check / another for del. 

... I'm aware that would take more code since you'd need different test setups for each of the commands.

Just an opinion: feel free to disregard.

plugins/main/tap/tap.go
```diff
@@ -0,0 +1,460 @@
...
+		A *IPAMArgs `json:"cni"`
+	} `json:"args,omitempty"`
+	RuntimeConfig struct {
+		Mac string `json:"mac,omitempty"`
+	} `json:"runtimeConfig,omitempty"`
+}
+
+type IPAMArgs struct {
+	Mac string `json:"mac,omitempty"`
+}
```
maiqueb, line 62: hm, this puzzles me. 

I haven't read the rest of the code below, but seeing the MAC as a parameter of `IPAM` is imo weird.

plugins/main/tap/tap.go
```diff
@@ -0,0 +1,460 @@
...
+	types.NetConf
+	MultiQueue     bool    `json:"multiQueue"`
+	MTU            int     `json:"mtu"`
+	Mac            string  `json:"mac,omitempty"`
+	Owner          *uint32 `json:"owner,omitempty"`
+	Group          *uint32 `json:"group,omitempty"`
+	SelinuxContext string  `json:"selinuxContext,omitempty"`
+	Args           *struct {
+		A *IPAMArgs `json:"cni"`
+	} `json:"args,omitempty"`
```
maiqueb, line 54: IMO it is strange to provide the MAC address via something called IPAM args.

AFAICT IPAM should provide the configuration on how to provide the IP address to the pod interface - not the MAC.

I would only allow the configuration of the MAC via `RuntimeConfig`.

plugins/main/tap/tap.go
```diff
@@ -0,0 +1,460 @@
...
+// Due to issues with the vishvananda/netlink library (fix pending) it is not possible to create an ownerless/groupless
+// tap device. Until the issue is fixed, the workaround for creating a tap device with no owner/group is to use the iptool
+func createTapWithIptool(tmpName string, mtu int, multiqueue bool, mac string, owner *uint32, group *uint32) error {
+	minFDToCloseOnExec := 3
+	maxFDToCloseOnExec := 256
+	// we want to share the parent process std{in|out|err} - fds 0 through 2.
+	// Since the FDs are inherited on fork / exec, we close on exec all others.
+	for fd := minFDToCloseOnExec; fd < maxFDToCloseOnExec; fd++ {
+		syscall.CloseOnExec(fd)
+	}
```
maiqueb, line 114: I would move all this to a well named helper.

plugins/main/tap/tap.go
```diff
@@ -0,0 +1,460 @@
...
+	if multiqueue {
+		tapDeviceArgs = append(tapDeviceArgs, "multi_queue")
+	}
+
+	if owner != nil {
+		tapDeviceArgs = append(tapDeviceArgs, "user", fmt.Sprintf("%d", *owner))
+	}
+	if group != nil {
+		tapDeviceArgs = append(tapDeviceArgs, "group", fmt.Sprintf("%d", *group))
+	}
```
maiqueb, line 126: Likewise: I'd move this to a well named helper. 

Bonus points if you use a builder :P (it might be too much, since we have only 3 parameters, but it would avoid ugly boolean input arguments).

plugins/main/tap/tap.go
```diff
@@ -0,0 +1,460 @@
...
+
+	err = netns.Do(func(_ ns.NetNS) error {
+		err := createLink(tmpName, conf, netns)
+		if err != nil {
+			return err
+		}
+		err = ip.RenameLink(tmpName, ifName)
+
+		if err != nil {
```
maiqueb, line 212: nit: I'd inline this.

```golang
if err := ip.RenameLink(tmpName, ifName); err != nil { ... }
```

maiqueb (commented):
Really nice work.

Most of what I have are opinionated nits, but I would like to see a README with info on how to use this particular plugin, and I question the proposed API (providing the MAC address of the tap via something named `IPAMArgs` .

plugins/main/tap/tap.go
mmirecki: Done

plugins/main/tap/tap.go
mmirecki: With 3 paramters it's still manageable. Let's do that when we have more.

plugins/main/tap/tap.go
mmirecki: Done

plugins/main/tap/tap.go
mmirecki: We were asked for this at some stage. Removing

plugins/main/tap/tap_test.go
mmirecki: This is the pattern in all the tests, so I'm following it for the same for uniformity.
If we change this, we should do it in a separate PR for all tests.

plugins/main/tap/tap_test.go
mmirecki: Note that the 2 are not the same:
allocator.IPAMConfig != types.IPAM

plugins/main/tap/tap_test.go
mmirecki: Done

plugins/main/tap/tap_test.go
mmirecki: Done

plugins/main/tap/tap_test.go
maiqueb: Fair enough.

plugins/main/tap/tap.go
maiqueb: totally understandable.

dcbw:
I'm OK with the updates WRT selinux stuff and the /sbin/ip vs. netlink issues.

dcbw:
@squeed how about now?

plugins/main/tap/tap.go
```diff
@@ -0,0 +1,457 @@
...
+	SelinuxContext string  `json:"selinuxContext,omitempty"`
+	Args           *struct {
+		A *IPAMArgs `json:"cni"`
+	} `json:"args,omitempty"`
+	RuntimeConfig struct {
+		Mac string `json:"mac,omitempty"`
+	} `json:"runtimeConfig,omitempty"`
+}
+
+type IPAMArgs struct{}
```
maiqueb, line 60: what's the purpose of this struct now ? IIUC, you're not using it anywhere.

Apologies if I misread the code.

mmirecki: Removed

mmirecki:
@squeed Can you please have a look again?

dcbw (approved):

plugins/main/tap/tap.go
```diff
@@ -0,0 +1,454 @@
...
+		}
+		return createTapWithIptool(tmpName, conf.MTU, conf.MultiQueue, conf.Mac, conf.Owner, conf.Group)
+	} else if conf.Owner == nil || conf.Group == nil {
+		return createTapWithIptool(tmpName, conf.MTU, conf.MultiQueue, conf.Mac, conf.Owner, conf.Group)
+	} else {
+		return createLinkWithNetlink(tmpName, conf.MTU, int(netns.Fd()), conf.MultiQueue, conf.Mac, conf.Owner, conf.Group)
+	}
+}
+
+func createTap(conf *NetConf, ifName string, netns ns.NetNS) (*current.Interface, error) {
```
fahedouch, line 191: Hi @mmirecki,
thank you for your create job. Is it possible to export this func we may need to refacto [container rootlesskit code](https://github.com/rootless-containers/rootlesskit/blob/master/pkg/network/parentutils/parentutils.go#L14). If LGTY I can take care thanks

PR #802, opened 2022-12-28
Add IPv6 support for AddDefaultRoute

Y7n05h:
AddDefaultRoute assumes that the user is adding a default route for IPv4, not IPv6. This pr will fix it.

### DIFF
--- a/pkg/ip/route_linux.go
+++ b/pkg/ip/route_linux.go
@@ -42,6 +42,11 @@ func AddHostRoute(ipn *net.IPNet, gw net.IP, dev netlink.Link) error {
 
 // AddDefaultRoute sets the default route on the given gateway.
 func AddDefaultRoute(gw net.IP, dev netlink.Link) error {
-	_, defNet, _ := net.ParseCIDR("0.0.0.0/0")
+	var defNet *net.IPNet
+	if gw.To4() != nil {
+		_, defNet, _ = net.ParseCIDR("0.0.0.0/0")
+	} else {
+		_, defNet, _ = net.ParseCIDR("::/0")
+	}
 	return AddRoute(defNet, gw, dev)
 }


mccv1r0:
Is there an issue with a description of something that isn't working?  AFAICT this function isn't used.

A few other things... 

1. adding default isn't either/or it maybe (and probably is) both unless single stack IPv6 is used.    If IPv4 is defined add `0.0.0.0/0`.  If IPv6 is used, add `::/0`.

2. The default route for IPv6 is (supposed to be) obtained via RA.

Y7n05h:
> 2. The default route for IPv6 is (supposed to be) obtained via RA.

Usually the default route for IPv6 is automatically configured by Router Advertisements. But sometimes it is necessary to configure it manually.

- Some routing devices do not send Router Advertisements
- When there are multiple IPv6 devices, and the default route configured by Router Advertisements does not match the expected device

mccv1r0:
> > 2. The default route for IPv6 is (supposed to be) obtained via RA.
> 
> Usually the default route for IPv6 is automatically configured by Router Advertisements. But sometimes it is necessary to configure it manually.



> * Some routing devices do not send Router Advertisements
> * When there are multiple IPv6 devices, and the default route configured by Router Advertisements does not match the expected device

mccv1r0:
Closed due to lack of coffee.  Apologies 

> > > 2. The default route for IPv6 is (supposed to be) obtained via RA.
> > 
> > 
> > Usually the default route for IPv6 is automatically configured by Router Advertisements. But sometimes it is necessary to configure it manually.
> 
> > * Some routing devices do not send Router Advertisements

The (current) cloud world we live in sadly.   The PR makes sense as a result.  Is this for your plugin?  IIRC this isn't used by any of the CNI  reference plugins.  

> > * When there are multiple IPv6 devices, and the default route configured by Router Advertisements does not match the expected device

This I'm having a harder time parsing... "expected" device?  If the network tells you what to use it's best to use that and not what you expected.  I get 8 default routes from my ISP (cheapest plan), some come and go as ISP needs require.  It's safe for me not to expect any.

squeed (approved):

squeed:
Note: This function isn't used by the plugins, I can't guarantee it continues to exist.

PR #807, opened 2023-01-10
Use the mac addr of the host side of veth pair after all the dust settles

mccv1r0:
Use the mac addr of the host side of veth pair after all the dust settles
    
Fixes 805
    
Signed-off-by: Michael Cambria <mcambria@redhat.com>


mccv1r0:
72 commits?   should be one file.  Closing

PR #812, opened 2023-01-18
Update ginkgo to v2

liornoy:
This PR updates "Ginkgo" version to v2 and subsequently updates "Golang" version to 1.18.
In the first commit: update go.mod, go.sum, and the vendor dir.
In the second commit: update the tests themselves to use ginkgo/v2.


squeed:
Thanks for taking this on!

liornoy:
> Thanks for taking this on!

You're welcome :)

Now that I think of that I should have committed with the "Co-authored-by:" signature, as your previous work spared me most of the issues.

squeed (approved):

liornoy:
I meant @saschagrunert first PR *

liornoy:
Added `Co-authored-by: Sascha Grunert <sgrunert@redhat.com>` to the commits messages.

plugins/main/bridge/bridge_test.go
```diff
@@ -17,6 +17,7 @@ package main
 import (
 	"encoding/json"
 	"fmt"
+	"io/ioutil"
```
austinvazquez, line 20: Package `io/ioutil` is marked deprecated since Go 1.16. It would be better to use `os` package.

plugins/main/bridge/bridge_suite_test.go
```diff
@@ -15,13 +15,26 @@
...
 
+var (
+	resolvConf string
+	err        error
+)
+
 func TestBridge(t *testing.T) {
 	RegisterFailHandler(Fail)
+
+	resolvConf, err = newResolvConf()
```
austinvazquez, line 32: Would it be more clean to defer the cleanup from within the test to eliminate the need for global variables?

liornoy: It's not uncommon to have global vars in the test_suite used by the test files.
And I think it's more of a good practice to have the cleanup on the test_suite (as mentioned [here](https://onsi.github.io/ginkgo/#suite-setup-and-cleanup-beforesuite-and-aftersuite))
Plus, having a `defer deleteResolvConf(resolvConf)` within the test i.e in the Describe node won't work.

plugins/main/bridge/bridge_test.go
liornoy: removed, thanks

liornoy:
Fixed comments of reviews plus removed redundant changes on import lines from the auto editor.formatOnSave

plugins/main/bridge/bridge_suite_test.go
liornoy: @austinvazquez please let me know if this is ok with you, or do you have a different idea for how to make it cleaner?

liornoy:
> @liornoy could you rebase to fix up the go.sum conflict? Then I think it's OK to merge. Thanks!

Sure, done.

liornoy:
@dcbw please let me know if this is OK to merge now after I rebased and fixed the conflicts.

squeed (approved):

dcbw:
@liornoy thanks for sticking with it!

PR #814, opened 2023-01-19
Add support for in-container master for macvlans

mmirecki:
Signed-off-by: mmirecki <mmirecki@redhat.com>

### DIFF
--- a/plugins/main/macvlan/macvlan_test.go
+++ b/plugins/main/macvlan/macvlan_test.go
@@ -38,6 +38,7 @@ import (
 )
 
 const MASTER_NAME = "eth0"
+const MASTER_NAME_INCONTAINER = "eth1"
 
 type Net struct {
 	Name       string                `json:"name"`
@@ -54,6 +55,7 @@ type Net struct {
 	DNS           types.DNS              `json:"dns"`
 	RawPrevResult map[string]interface{} `json:"prevResult,omitempty"`
 	PrevResult    types100.Result        `json:"-"`
+	LinkContNs    bool                   `json:"linkInContainer"`
 }
 
 func buildOneConfig(netName string, cniVersion string, orig *Net, prevResult types.Result) (*Net, error) {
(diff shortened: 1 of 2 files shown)


squeed:
@s1061123 can you take a look?

plugins/main/macvlan/macvlan_test.go
```diff
@@ -54,6 +55,7 @@ type Net struct {
 	DNS           types.DNS              `json:"dns"`
 	RawPrevResult map[string]interface{} `json:"prevResult,omitempty"`
 	PrevResult    types100.Result        `json:"-"`
+	LinkContNs    bool                   `json:"linkInContainer"`
```
s1061123, line 58: This field, `LinkContNs` does not have `omitempty`. It could be optional field, so we could add `omitempty` in json annotation.

It is minor case, only for re-encode this structure again by someone, but it might be good to annotate.

plugins/main/macvlan/macvlan_test.go
```diff
@@ -225,544 +243,560 @@ var _ = Describe("macvlan Operations", func() {
...
+				conf := &NetConf{
+					NetConf: types.NetConf{
+						CNIVersion: ver,
+						Name:       "testConfig",
+						Type:       "macvlan",
+					},
+					Master:     masterInterface,
+					Mode:       "bridge",
+					MTU:        1500,
+					LinkContNs: isInContainer,
```
s1061123, line 267: Could you please add test case without 'LinkContNs' and verify default behavior (i.e. without 'LinkContNs') is the same as 'LinkContNs' is `false`?

plugins/main/macvlan/macvlan_test.go
mmirecki: Done

plugins/main/macvlan/macvlan_test.go
mmirecki: Done

s1061123:
It looks good to me, thank you so much for incorporate comments!

@squeed I reviewed the PR. Could you please double check that?

mccv1r0:
Tested both with and without `"linkInContainer": true` using cnitool and `host-local` ipam

dcbw:
@squeed does this look good to you now?

mmirecki:
@squeed Can you please have a look again?

squeed (approved):

PR #826, opened 2023-02-02
Fix tuning allowlist

wzshiming:
``` console
"failed to call cni.Setup: plugin type=\"tuning\" failed (add): open /etc/cni/tuning/allowlist.conf: permission denied"

```

### DIFF
--- a/plugins/meta/tuning/tuning.go
+++ b/plugins/meta/tuning/tuning.go
@@ -502,6 +502,9 @@ func cmdCheck(args *skel.CmdArgs) error {
 func validateSysctlConf(tuningConf *TuningConf) error {
 	isPresent, allowlist, err := readAllowlist()
 	if err != nil {
+		if os.IsPermission(err) {
+			return nil
+		}
 		return err
 	}
 	if !isPresent {


plugins/meta/tuning/tuning.go
```diff
@@ -502,6 +502,9 @@ func cmdCheck(args *skel.CmdArgs) error {
 func validateSysctlConf(tuningConf *TuningConf) error {
 	isPresent, allowlist, err := readAllowlist()
 	if err != nil {
+		if os.IsPermission(err) {
+			return nil
+		}
```
s1061123, line 507: This code means if allowList cannot be read, then just skip. I'm a bit curious that this behavior is intentional or not.
In such case, user should check the permission and fix that to read allowlist by tuning plugin. Otherwise, allowlist is not ignored. 

At least, to me, we should not ignore this error in such case.

wzshiming: I think you have a point, I'll go to nerdctl and raise an issue to discuss it
containerd/nerdctl#2082

PR #829, opened 2023-02-13
bridge: add vlan trunk support

tjjh89017:
add vlan trunk support for veth
vlan trunk only support L2 only mode without any IPAM
refer ovs-cni design
https://github.com/k8snetworkplumbingwg/ovs-cni/blob/main/pkg/plugin/plugin.go

design:
origin "vlan" option will be PVID or untagged vlan for the network.
"vlanTrunk" will setup tagged vlan for veth.

entry type:
`{ "id": 100 }` will specify only tagged vlan 100
`{ "minID": 100, "maxID": 120 }` will specify tagged vlan from 100 to
120 (include 100 and 120)
vlanTrunk is a list of above entry type, so you can use this to add
tagged vlan
`[
  { "id": 100 },
  {
    "minID": 1000,
    "maxID": 2000
  }
]`

complete config will be like this
```json
{
  "cniVersion": "0.3.1",
  "name": "mynet",
  "type": "bridge",
  "bridge": "mynet0",
  "vlan": 100,
  "vlanTrunk": [
    { "id": 101 },
    { "minID": 1000, "maxID": 2000 },
    { "minID": 3000, "maxID": 4000 }
  ],
  "ipam": {}
}
```

Original PR: https://github.com/containernetworking/plugins/pull/689

Signed-off-by: Date Huang <date.huang@suse.com>

### DIFF
--- a/plugins/main/bridge/bridge.go
+++ b/plugins/main/bridge/bridge.go
@@ -21,6 +21,7 @@ import (
 	"net"
 	"os"
 	"runtime"
+	"sort"
 	"syscall"
 	"time"
 
@@ -46,18 +47,19 @@ const defaultBrName = "cni0"
 
 type NetConf struct {
 	types.NetConf
-	BrName              string `json:"bridge"`
-	IsGW                bool   `json:"isGateway"`
-	IsDefaultGW         bool   `json:"isDefaultGateway"`
-	ForceAddress        bool   `json:"forceAddress"`
-	IPMasq              bool   `json:"ipMasq"`
-	MTU                 int    `json:"mtu"`
-	HairpinMode         bool   `json:"hairpinMode"`
-	PromiscMode         bool   `json:"promiscMode"`
-	Vlan                int    `json:"vlan"`
-	PreserveDefaultVlan bool   `json:"preserveDefaultVlan"`
-	MacSpoofChk         bool   `json:"macspoofchk,omitempty"`
-	EnableDad           bool   `json:"enabledad,omitempty"`
+	BrName              string       `json:"bridge"`
+	IsGW                bool         `json:"isGateway"`
+	IsDefaultGW         bool         `json:"isDefaultGateway"`
+	ForceAddress        bool         `json:"forceAddress"`
+	IPMasq              bool         `json:"ipMasq"`
+	MTU                 int          `json:"mtu"`
+	HairpinMode         bool         `json:"hairpinMode"`
+	PromiscMode         bool         `json:"promiscMode"`
+	Vlan                int          `json:"vlan"`
+	VlanTrunk           []*VlanTrunk `json:"vlanTrunk,omitempty"`
+	PreserveDefaultVlan bool         `json:"preserveDefaultVlan"`
+	MacSpoofChk         bool         `json:"macspoofchk,omitempty"`
+	EnableDad           bool         `json:"enabledad,omitempty"`
 
 	Args struct {
 		Cni BridgeArgs `json:"cni,omitempty"`
@@ -66,7 +68,14 @@ type NetConf struct {
 		Mac string `json:"mac,omitempty"`
 	} `json:"runtimeConfig,omitempty"`
 
-	mac string
+	mac   string
+	vlans []int
+}
+
+type VlanTrunk struct {
+	MinID *int `json:"minID,omitempty"`
+	MaxID *int `json:"maxID,omitempty"`
+	ID    *int `json:"id,omitempty"`
 }
 
 type BridgeArgs struct {
@@ -104,6 +113,17 @@ func loadNetConf(bytes []byte, envArgs string) (*NetConf, string, error) {
 	if n.Vlan < 0 || n.Vlan > 4094 {
 		return nil, "", fmt.Errorf("invalid VLAN ID %d (must be between 0 and 4094)", n.Vlan)
 	}
+	var err error
+	n.vlans, err = collectVlanTrunk(n.VlanTrunk)
+	if err != nil {
+		// fail to parsing
+		return nil, "", err
+	}
+
+	// Currently bridge CNI only support access port(untagged only) or trunk port(tagged only)
+	if n.Vlan > 0 && n.vlans != nil {
+		return nil, "", errors.New("cannot set vlan and vlanTrunk at the same time")
+	}
 
 	if envArgs != "" {
 		e := MacEnvArgs{}
@@ -125,6 +145,61 @@ func loadNetConf(bytes []byte, envArgs string) (*NetConf, string, error) {
 	}
 
 	return n, n.CNIVersion, nil
+}
+
+// This method is copied from https://github.com/k8snetworkplumbingwg/ovs-cni/blob/v0.27.2/pkg/plugin/plugin.go
+func collectVlanTrunk(vlanTrunk []*VlanTrunk) ([]int, error) {
+	if vlanTrunk == nil {
+		return nil, nil
+	}
+
+	vlanMap := make(map[int]struct{})
+	for _, item := range vlanTrunk {
+		var minID int
+		var maxID int
+		var ID int
+
+		switch {
+		case item.MinID != nil && item.MaxID != nil:
+			minID = *item.MinID
+			if minID <= 0 || minID > 4094 {
+				return nil, errors.New("incorrect trunk minID parameter")
+			}
+			maxID = *item.MaxID
+			if maxID <= 0 || maxID > 4094 {
+				return nil, errors.New("incorrect trunk maxID parameter")
+			}
+			if maxID < minID {
+				return nil, errors.New("minID is greater than maxID in trunk parameter")
+			}
+			for v := minID; v <= maxID; v++ {
+				vlanMap[v] = struct{}{}
+			}
+		case item.MinID == nil && item.MaxID != nil:
+			return nil, errors.New("minID and maxID should be configured simultaneously, minID is missing")
+		case item.MinID != nil && item.MaxID == nil:
+			return nil, errors.New("minID and maxID should be configured simultaneously, maxID is missing")
+		}
+
+		// single vid
+		if item.ID != nil {
+			ID = *item.ID
+			if ID <= 0 || ID > 4094 {
+				return nil, errors.New("incorrect trunk id parameter")
+			}
+			vlanMap[ID] = struct{}{}
+		}
+	}
+
+	if len(vlanMap) == 0 {
+		return nil, nil
+	}
+	vlans := make([]int, 0, len(vlanMap))
+	for k := range vlanMap {
+		vlans = append(vlans, k)
+	}
+	sort.Slice(vlans, func(i int, j int) bool { return vlans[i] < vlans[j] })
+	return vlans, nil
 }
 
 // calcGateways processes the results from the IPAM plugin and does the
@@ -316,7 +391,7 @@ func ensureVlanInterface(br *netlink.Bridge, vlanID int, preserveDefaultVlan boo
 			return nil, fmt.Errorf("faild to find host namespace: %v", err)
 		}
 
-		_, brGatewayIface, err := setupVeth(hostNS, br, name, br.MTU, false, vlanID, preserveDefaultVlan, "")
+		_, brGatewayIface, err := setupVeth(hostNS, br, name, br.MTU, false, vlanID, nil, preserveDefaultVlan, "")
 		if err != nil {
 			return nil, fmt.Errorf("faild to create vlan gateway %q: %v", name, err)
 		}
@@ -335,7 +410,7 @@ func ensureVlanInterface(br *netlink.Bridge, vlanID int, preserveDefaultVlan boo
 	return brGatewayVeth, nil
 }
 
-func setupVeth(netns ns.NetNS, br *netlink.Bridge, ifName string, mtu int, hairpinMode bool, vlanID int, preserveDefaultVlan bool, mac string) (*current.Interface, *current.Interface, error) {
+func setupVeth(netns ns.NetNS, br *netlink.Bridge, ifName string, mtu int, hairpinMode bool, vlanID int, vlans []int, preserveDefaultVlan bool, mac string) (*current.Interface, *current.Interface, error) {
 	contIface := &current.Interface{}
 	hostIface := &current.Interface{}
 
@@ -372,17 +447,25 @@ func setupVeth(netns ns.NetNS, br *netlink.Bridge, ifName string, mtu int, hairp
 		return nil, nil, fmt.Errorf("failed to setup hairpin mode for %v: %v", hostVeth.Attrs().Name, err)
 	}
 
-	if vlanID != 0 {
-		if !preserveDefaultVlan {
-			err = removeDefaultVlan(hostVeth)
-			if err != nil {
-				return nil, nil, fmt.Errorf("failed to remove default vlan on interface %q: %v", hostIface.Name, err)
-			}
+	if (vlanID != 0 || len(vlans) > 0) && !preserveDefaultVlan {
+		err = removeDefaultVlan(hostVeth)
+		if err != nil {
+			return nil, nil, fmt.Errorf("failed to remove default vlan on interface %q: %v", hostIface.Name, err)
 		}
+	}
 
+	// Currently bridge CNI only support access port(untagged only) or trunk port(tagged only)
+	if vlanID != 0 {
 		err = netlink.BridgeVlanAdd(hostVeth, uint16(vlanID), true, true, false, true)
 		if err != nil {
 			return nil, nil, fmt.Errorf("failed to setup vlan tag on interface %q: %v", hostIface.Name, err)
+		}
+	}
+
+	for _, v := range vlans {
+		err = netlink.BridgeVlanAdd(hostVeth, uint16(v), false, false, false, true)
+		if err != nil {
+			return nil, nil, fmt.Errorf("failed to setup vlan tag on interface %q: %w", hostIface.Name, err)
 		}
 	}
 
@@ -414,7 +497,10 @@ func calcGatewayIP(ipn *net.IPNet) net.IP {
 }
 
 func setupBridge(n *NetConf) (*netlink.Bridge, *current.Interface, error) {
-	vlanFiltering := n.Vlan != 0
+	vlanFiltering := false
+	if n.Vlan != 0 || n.VlanTrunk != nil {
+		vlanFiltering = true
+	}
 	// create bridge if necessary
 	br, err := ensureBridge(n.BrName, n.MTU, n.PromiscMode, vlanFiltering)
 	if err != nil {
@@ -463,7 +549,7 @@ func cmdAdd(args *skel.CmdArgs) error {
 	}
 	defer netns.Close()
 
-	hostInterface, containerInterface, err := setupVeth(netns, br, args.IfName, n.MTU, n.HairpinMode, n.Vlan, n.PreserveDefaultVlan, n.mac)
+	hostInterface, containerInterface, err := setupVeth(netns, br, args.IfName, n.MTU, n.HairpinMode, n.Vlan, n.vlans, n.PreserveDefaultVlan, n.mac)
 	if err != nil {
 		return err
 	}
--- a/plugins/main/bridge/bridge_test.go
+++ b/plugins/main/bridge/bridge_test.go
@@ -74,6 +74,7 @@ type testCase struct {
 	isLayer2          bool
 	expGWCIDRs        []string // Expected gateway addresses in CIDR form
 	vlan              int
+	vlanTrunk         []*VlanTrunk
 	removeDefaultVlan bool
 	ipMasq            bool
 	macspoofchk       bool
@@ -129,6 +130,23 @@ const (
 
 	vlan = `,
 	"vlan": %d`
+
+	vlanTrunkStartStr = `,
+	"vlanTrunk": [`
+
+	vlanTrunk = `
+	{
+		"id": %d
+	}`
+
+	vlanTrunkRange = `
+	{
+		"minID": %d,
+		"maxID": %d
+	}`
+
+	vlanTrunkEndStr = `
+	]`
 
 	preserveDefaultVlan = `,
 	"preserveDefaultVlan": false`
@@ -200,6 +218,23 @@ func (tc testCase) netConfJSON(dataDir string) string {
 			conf += preserveDefaultVlan
 		}
 	}
+
+	if tc.isLayer2 && tc.vlanTrunk != nil {
+		conf += vlanTrunkStartStr
+		for i, vlan := range tc.vlanTrunk {
+			if i > 0 {
+				conf += ","
+			}
+			if vlan.ID != nil {
+				conf += fmt.Sprintf(vlanTrunk, *vlan.ID)
+			}
+			if vlan.MinID != nil && vlan.MaxID != nil {
+				conf += fmt.Sprintf(vlanTrunkRange, *vlan.MinID, *vlan.MaxID)
+			}
+		}
+		conf += vlanTrunkEndStr
+	}
+
 	if tc.ipMasq {
 		conf += tc.ipMasqConfig()
 	}
@@ -541,7 +576,7 @@ func (tester *testerV10x) cmdAddTest(tc testCase, dataDir string) (types.Result,
 		}
 
 		// Check the bridge vlan filtering equals true
-		if tc.vlan != 0 {
+		if tc.vlan != 0 || tc.vlanTrunk != nil {
 			Expect(*link.(*netlink.Bridge).VlanFiltering).To(BeTrue())
 		} else {
 			Expect(*link.(*netlink.Bridge).VlanFiltering).To(BeFalse())
@@ -595,6 +630,25 @@ func (tester *testerV10x) cmdAddTest(tc testCase, dataDir string) (types.Result,
 			Expect(checkVlan(tc.vlan, vlans)).To(BeTrue())
 			if tc.removeDefaultVlan {
 				Expect(vlans).To(HaveLen(1))
+			}
+		}
+
+		// check VlanTrunks exist on the veth interface
+		if tc.vlanTrunk != nil {
+			interfaceMap, err := netlink.BridgeVlanList()
+			Expect(err).NotTo(HaveOccurred())
+			vlans, isExist := interfaceMap[int32(link.Attrs().Index)]
+			Expect(isExist).To(BeTrue())
+
+			for _, vlanEntry := range tc.vlanTrunk {
+				if vlanEntry.ID != nil {
+					Expect(checkVlan(*vlanEntry.ID, vlans)).To(BeTrue())
+				}
+				if vlanEntry.MinID != nil && vlanEntry.MaxID != nil {
+					for vid := *vlanEntry.MinID; vid <= *vlanEntry.MaxID; vid++ {
+						Expect(checkVlan(vid, vlans)).To(BeTrue())
+					}
+				}
 			}
 		}
 
@@ -852,7 +906,7 @@ func (tester *testerV04x) cmdAddTest(tc testCase, dataDir string) (types.Result,
 		}
 
 		// Check the bridge vlan filtering equals true
-		if tc.vlan != 0 {
+		if tc.vlan != 0 || tc.vlanTrunk != nil {
 			Expect(*link.(*netlink.Bridge).VlanFiltering).To(BeTrue())
 		} else {
 			Expect(*link.(*netlink.Bridge).VlanFiltering).To(BeFalse())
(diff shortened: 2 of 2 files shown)


tjjh89017:
@squeed Please use this PR for https://github.com/containernetworking/plugins/pull/689
Thanks a lot

squeed:
hey @SchSeba would you mind taking a quick look?

tjjh89017:
I found the test failed in "portmap" plugin.

SchSeba:
I will take a look on this one.
also @mlguerrero12 if you have time please take a look :)

tjjh89017:
Hi @squeed and @SchSeba 

I tried to fix up all lint issue and rebase to current main branche.
Could you help us to look on this PR?
Thanks a lot

plugins/main/bridge/bridge.go
```diff
@@ -127,6 +142,59 @@ func loadNetConf(bytes []byte, envArgs string) (*NetConf, string, error) {
 	return n, n.CNIVersion, nil
 }
 
+// This method is copied from https://github.com/k8snetworkplumbingwg/ovs-cni/blob/v0.27.2/pkg/plugin/plugin.go
+func collectVlanTrunk(vlanTrunk []*VlanTrunk) ([]int, error) {
+	if vlanTrunk == nil {
+		return nil, nil
+	}
+
+	vlanMap := make(map[int]bool)
```
maiqueb, line 151: might as well use a `map[int]struct{}` which I think is more memory efficient. IIUC, you're using a set (you only care about the keys, and never put the value `false`  in the map).

I also think it is more idiomatic, since the type clearly shows there's nothing in the value to be looked at.

plugins/main/bridge/bridge.go
```diff
@@ -127,6 +142,59 @@ func loadNetConf(bytes []byte, envArgs string) (*NetConf, string, error) {
...
+		}
+		if minID > 0 && maxID > 0 {
+			for v := minID; v <= maxID; v++ {
+				vlanMap[v] = true
+			}
+		}
+
+		// single vid
+		if item.ID != nil {
+			ID = *item.ID
```
maiqueb, line 179: nit: IMO, there's no need to have the `ID` variable.

plugins/main/bridge/bridge.go
```diff
@@ -127,6 +142,59 @@ func loadNetConf(bytes []byte, envArgs string) (*NetConf, string, error) {
...
+		}
+	}
+
+	if len(vlanMap) == 0 {
+		return nil, nil
+	}
+	vlans := make([]int, 0, len(vlanMap))
+	for k := range vlanMap {
+		vlans = append(vlans, k)
+	}
```
maiqueb, line 193: I may be missing something, but I don't understand why you first compute a set of the VLAN IDs, and then compute a slice from it.

Couldn't you have just created the slice directly ? And append the trunk IDs to it?

plugins/main/bridge/bridge.go
```diff
@@ -452,6 +531,10 @@ func cmdAdd(args *skel.CmdArgs) error {
 		return fmt.Errorf("cannot set hairpin mode and promiscuous mode at the same time")
 	}
 
+	if n.vlans != nil && len(n.vlans) > 0 && isLayer3 {
```
maiqueb, line 534: nit: I think `if n.vlans != nil && len(n.vlans) > 0` is redundant; len can handle a nil slice.

maiqueb (commented):
some nits.

plugins/main/bridge/bridge.go
tjjh89017: Just copy out and avoid to dereference `item.ID` every time.
If needed, I could remove ID here.

plugins/main/bridge/bridge.go
tjjh89017: I copy those from OVS cni.
IIRC, we want to deduplicate vlan ids and sort them.
If go netlink supports this [PR](https://github.com/vishvananda/netlink/pull/863) and we could call netlink once for contiguous vlan id.

plugins/main/bridge/bridge.go
tjjh89017: I will fix this part.

plugins/main/bridge/bridge.go
tjjh89017: ok got it.

plugins/main/bridge/bridge.go
mlguerrero12: I think it is with the aim to support config like:
{ "minID": 100, "maxID": 200 }
{ "minID": 150, "maxID": 200 }

if you use a slice directly, it will append ids 150 to 200 again.

Another thing, the config { "maxID": 120 } does not affect the code (if in 171 addresses this), but shouldn't we return an error saying that maxID was set but not minID?

plugins/main/bridge/bridge.go
```diff
@@ -386,6 +454,14 @@ func setupVeth(netns ns.NetNS, br *netlink.Bridge, ifName string, mtu int, hairp
 		}
 	}
 
+	// S1031: unnecessary nil check around range (gosimple)
```
mlguerrero12, line 457: no need to add this comment I'd say

plugins/main/bridge/bridge.go
```diff
@@ -386,6 +454,14 @@ func setupVeth(netns ns.NetNS, br *netlink.Bridge, ifName string, mtu int, hairp
...
+	for _, v := range vlans {
```
mlguerrero12, line 458: the default vlan will be present if the regular vlan is not set and preserveDefaultVlan is true. I would add this inside the if above (adding || len(vlans) != 0 and a if for the BridgeVlanAdd) to avoid this.

plugins/main/bridge/bridge.go
tjjh89017: > Another thing, the config { "maxID": 120 } does not affect the code (if in 171 addresses this), but shouldn't we return an error saying that maxID was set but not minID?

That is a problem, let me think how to deal with it.
Thanks for review!

plugins/main/bridge/bridge.go
maiqueb: I think this is to appease the linter.

tjjh89017: I just do a note to make sure no one will add a nil check there again.
I could remove it later

plugins/main/bridge/bridge.go
tjjh89017: I will suggest to move the preserveDefaultVlan condition to upper layer (not in if vlanID != 0 condition).


FROM: 
```
    if vlanID != 0 {
        if !preserveDefaultVlan {
            err = removeDefaultVlan(hostVeth)
            if err != nil {
                return nil, nil, fmt.Errorf("failed to remove default vlan on interface %q: %v", hostIface.Name, err)
            }
        }

        err = netlink.BridgeVlanAdd(hostVeth, uint16(vlanID), true, true, false, true)
        if err != nil {
            return nil, nil, fmt.Errorf("failed to setup vlan tag on interface %q: %v", hostIface.Name, err)
        }
    }
```

TO:
```
    if !preserveDefaultVlan {
        err = removeDefaultVlan(hostVeth)
        if err != nil {
            return nil, nil, fmt.Errorf("failed to remove default vlan on interface %q: %v", hostIface.Name, err)
        }
    }

    if vlanID != 0 {
        err = netlink.BridgeVlanAdd(hostVeth, uint16(vlanID), true, true, false, true)
        if err != nil {
            return nil, nil, fmt.Errorf("failed to setup vlan tag on interface %q: %v", hostIface.Name, err)
        }
    }
```

plugins/main/bridge/bridge.go
tjjh89017: ok, i removed it.

plugins/main/bridge/bridge.go
mlguerrero12: You can add the contents of the first if as an else of the second one. You don't need to check again if item.MinID or item.MaxID are != nil again.

plugins/main/bridge/bridge.go
mlguerrero12: the reasoning is that we only modify the default vlan when the vlan filtering is enabled (line 494).

plugins/main/bridge/bridge.go
tjjh89017: The problem is we have 3 case here
1. minID == nil && maxID == nil; need to check ID
2. (minID != nil && maxID == nil) || (minID == nil && maxID != nil); fail case; L174 try to capture.
3. minID != nil && maxID != nul; success case

L157 only capture case 3, so I need to use L174 to check case 1 or 2.
If case 2, and we return error.
otherwise, we should use the condition statement in case 2 in the code

tjjh89017: I fix this.
I had wrong statement.
current `else if` should be correct.

mlguerrero12:
I see a conflict with PVID here that has not been addressed.

When a port has PVID set (regular vlan id here), this becomes an access port. On the other hand, a trunk port has no PVID. Only tagged VIDs and no VID (default vlan). That's a check we need to add on this pr.

So, a use case for vlan trunk is a container with a trunk port on the bridge. This container can use multus for example to create additional vlan subinterfaces. Let's say two for vlan 100 and 200. Then, 3 containers can be created, one with vlan 100, the other one with 200 and the last one with no vlan. The subinterfaces on the first container should be able to talk to the containers with the specific vlan and untagged traffic should be forwarded thanks to the default vlan.

Now, regarding the default vlan, with the new parameter preserveDefaultVlan, it is possible to remove it. For instance, if a user only wants tagged traffic to go through the trunk port.

Hope this makes sense.

plugins/main/bridge/bridge.go
mlguerrero12: if item.MinID != nil || item.MaxID != nil {
   print error
} else {
  do work
}

this is what I meant

tjjh89017: So put `single vid work` in the else statement?

tjjh89017:
```
if item.MinID != nil || item.MaxID != nil {
print error
} else {
//do work
// add single vid 
}
```

or 

```
if item.MinID != nil || item.MaxID != nil {
print error
} else {
//do work
// add vlan range from minID to maxID
}
```

tjjh89017:
> Exactly, the pvid here is the native vlan, not the one coming from the parameter vlan. From you message, I think you believe the native vlan is configured with this parameter and it is not. We currently don't have a param to modify the native vlan of the bridge.
> 
> When we add the pvid (vlan param equal 100), the port has: 1 Egress untagged 100 PVID Egress untagged
> 
> Now, if you add trunk port. This will become.
> 
> 1 Egress untagged 100 PVID Egress untagged 200 300
> 
> Native vlan is still 1 (this a configuration of the bridge itself, not the ports of the bridge).
> 
> With the above config, untagged ingress traffic will have tag 100 (clearly not the native vlan) and on egress, vid 100 and untagged traffic will output untagged.
> 
> I'm familiar with cisco switches as well and I don't remember a config to do this.

So did it mean that we need to always remove the vid 1 when vlan is presented?

```
When we add the pvid (vlan param equal 100), the port has:
1 Egress untagged
100 PVID Egress untagged
```
In this config, vid 1 didn't take any effect and it might have some issue on this port.

---

> Native vlan is still 1 (this a configuration of the bridge itself, not the ports of the bridge).

In this bridge cni, I think we never modify the bridge's vlan config.
but only port's vlan config.
So the native vlan that I mentioned before is always port's native vlan.

---

> With the above config, untagged ingress traffic will have tag 100 (clearly not the native vlan) and on egress, vid 100 and untagged traffic will output untagged.

I'm not really confident that I understand this part.
ingress is ok, follow PVID config.
egress part.
If we have two port eth0 & eth1 and those ports are bridge ports.
```
eth0: 1 pvid egress untagged
eth1: 100 pvid egress untagged
      1-99
```
traffic from eth0 to eth1 will be captured with vid 1 tagged in eth1 egress direction.

The only config with problem will be like this

```
eth0: 1 pvid egress untagged
eth1: 1 egress untagged
      100 pvid egress untagged
```
traffic from eth0 to eth1 will be captured with untagged in eth1 egress direction.
but traffic from eth1 to eth0 will be catpured with tagged vid 100 in eth0 egress direction (so drop.)

---

> I'm familiar with cisco switches as well and I don't remember a config to do this.
This is my config of my switch.
Cisco didn't have "bridge's vlan config" in Linux.
cisco only have port's vlan config.
So I'm not sure which config that is not in cisco switch.

c2960s
```
interface GigabitEthernet1/0/48
 description to-4100
 switchport trunk native vlan 101
 switchport trunk allowed vlan 100,101,999
 switchport mode trunk
!
```

---

So from my side, I will suggest to remove default vlan always when vlanID is presented.

Stay away from the config as this 
```
// Bad 
1 Egress untagged
100 PVID Egress untagged

// Good (vlanID == 100)
100 PVID Egress untagged

// Good (vlanID == 0)
1 PVID Egress untagged
```

tjjh89017:
If we took this cisco config as example to impelment.
```
interface GigabitEthernet1/0/48
 description to-4100
 switchport trunk native vlan 101
 switchport trunk allowed vlan 100,101,999
 switchport mode trunk
!
```

```
vlanID := 101
vlanTrunk := [{ID: 100}, {ID: 101}, {ID: 999}]
```

In Linux Bridge, we should get something like this below.

```
veth0 101 PVID Egress Untagged
      100
      999
```
without `1 Egress Untagged` inside.

I want to ensure that we have the same config translation in this Cisco config.

mlguerrero12:
If you want to achieve

veth0 101 PVID Egress Untagged
      100
      999

you need to create a bridge with a default vlan equal to 101 and then add tags 100 and 999 with this pr.

In my view, you want to achieve the same with a hack. You want to add a vlan with PVID (access port) 101 and then delete the default vlan (with preserveDefaultVlan equal to false) and add tags 100 and 999.

Conceptually, this is not correct.

mlguerrero12:
All the issues come from not having removed the default vlan of the bridge from the ports since day 1.

If we allow users this kind of flexibility, there will be confusion. They can set a different default vlan by setting a PVID and having preserveDefaultVlan equal to false. But this is contradicting, they will have a default vlan but they are explicitly saying that they don't want to preserve the default vlan. It works, but it's not semantically correct.

Another issue, if they leave the default vlan of the bridge and set a PVID. They will have on port:
1 Egress Untagged
101 PVID Untagged

Untagged ingress traffic is fine, it will be tagged with 101, but egress traffic with tag 101 and 1 will be untagged. In a cisco switch, if the default vlan is set to 101, the trunk will not modify any other traffic. Traffic with vlan 1 will have tag 1. This will not be the case here.

In my opinion, we should stick to basic concepts, at least for now to avoid these issues. A trunk port should only allow tagged vids and the native vlan, the one configured on the bridge.

@maiqueb @squeed, what do you think? See my comment above also.

tjjh89017:
> All the issues come from not having removed the default vlan of the bridge from the ports since day 1.
> 
> If we allow users this kind of flexibility, there will be confusion. They can set a different default vlan by setting a PVID and having preserveDefaultVlan equal to false. But this is contradicting, they will have a default vlan but they are explicitly saying that they don't want to preserve the default vlan. It works, but it's not semantically correct.
> 
> Another issue, if they leave the default vlan of the bridge and set a PVID. They will have on port: 1 Egress Untagged 101 PVID Untagged
> 
> Untagged ingress traffic is fine, it will be tagged with 101, but egress traffic with tag 101 and 1 will be untagged. In a cisco switch, if the default vlan is set to 101, the trunk will not modify any other traffic. Traffic with vlan 1 will have tag 1. This will not be the case here.
> 
> In my opinion, we should stick to basic concepts, at least for now to avoid these issues. A trunk port should only allow tagged vids and the native vlan, the one configured on the bridge.
> 
> @maiqueb @squeed, what do you think? See my comment above also.

As discussion with @mlguerrero12, I think we could make this PR with some limitation and leave something to the future version.
and I will file a ticket to deprecate some current behavior with breaking change to make sure there is no conflict in network scope and vlan/pvid/vids usage.

For now, I will modify the PR to support access port (untagged only) and trunk port (tagged only).
Leaving hybrid port support in the future.

---

by the way, I will suggest not to use the "default" vlan term.
Even Linux bridge could do some weird config like this [1]
But in current CNI config design, we only allow/accept common switch style config.
I will leave this to the next ticket to fix up and probably we need to remove `preserveDefaultVlan` option in the future.


[1]:
```
1 Egress Untagged
101 PVID Untagged
```

mlguerrero12:
cool, thanks @tjjh89017

plugins/main/bridge/bridge.go
mlguerrero12: ignore my last comment, you latest fix is ok

plugins/main/bridge/bridge.go
```diff
@@ -127,6 +142,58 @@ func loadNetConf(bytes []byte, envArgs string) (*NetConf, string, error) {
...
+			if maxID <= 0 || maxID > 4094 {
+				return nil, errors.New("incorrect trunk maxID parameter")
+			}
+			if maxID < minID {
+				return nil, errors.New("minID is greater than maxID in trunk parameter")
+			}
+			for v := minID; v <= maxID; v++ {
+				vlanMap[v] = struct{}{}
+			}
+		} else if item.MinID != nil || item.MaxID != nil {
```
maiqueb, line 172: isn't this clause wrong ? Shouldn't we be looking if one of them is nil instead ? 

And we probably can print exactly what is missing in the error, instead of saying one of them is missing.

plugins/main/bridge/bridge.go
maiqueb: all good.

maiqueb (changes requested):
Would it make sense to also unit-test the `collectVlanTrunk` function ? Especially it's "negative" scenario ? (e.g. check a trunk without maxID but with minID, etc).

plugins/main/bridge/bridge.go
tjjh89017: hmm ok, let me split it to 3 conditions.

maiqueb:
> All the issues come from not having removed the default vlan of the bridge from the ports since day 1.
> 
> If we allow users this kind of flexibility, there will be confusion. They can set a different default vlan by setting a PVID and having preserveDefaultVlan equal to false. But this is contradicting, they will have a default vlan but they are explicitly saying that they don't want to preserve the default vlan. It works, but it's not semantically correct.
> 
> Another issue, if they leave the default vlan of the bridge and set a PVID. They will have on port: 1 Egress Untagged 101 PVID Untagged
> 
> Untagged ingress traffic is fine, it will be tagged with 101, but egress traffic with tag 101 and 1 will be untagged. In a cisco switch, if the default vlan is set to 101, the trunk will not modify any other traffic. Traffic with vlan 1 will have tag 1. This will not be the case here.
> 
> In my opinion, we should stick to basic concepts, at least for now to avoid these issues. A trunk port should only allow tagged vids and the native vlan, the one configured on the bridge.
> 
> @maiqueb @squeed, what do you think? See my comment above also.

I agree w/ @mlguerrero12 - having a single, crystal clear flow regarding VLANs is preferable.

tjjh89017:
From my side,

> A trunk port should only allow tagged vids and the native vlan, the one configured on the bridge.

I didn't agree with this one in the future.
But currently, it's ok.

plugins/main/bridge/bridge.go
```diff
@@ -174,9 +175,9 @@ func collectVlanTrunk(vlanTrunk []*VlanTrunk) ([]int, error) {
 			for v := minID; v <= maxID; v++ {
 				vlanMap[v] = struct{}{}
 			}
-		} else if item.MinID == nil && item.MaxID != nil {
+		case item.MinID == nil && item.MaxID != nil:
```
maiqueb, line 178: nit: maybe omit the != nil ? you know one of them is empty ; just checking for the empty ones would be OK, right ?

plugins/main/bridge/bridge.go
```diff
@@ -174,9 +175,9 @@ func collectVlanTrunk(vlanTrunk []*VlanTrunk) ([]int, error) {
...
 			return nil, errors.New("minID and maxID should be configured simultaneously, minID is missing")
-		} else if item.MinID != nil && item.MaxID == nil {
+		case item.MinID != nil && item.MaxID == nil:
```
maiqueb, line 180: ditto

plugins/main/bridge/bridge.go
```diff
@@ -372,18 +447,27 @@ func setupVeth(netns ns.NetNS, br *netlink.Bridge, ifName string, mtu int, hairp
...
 		}
+	}
 
+	// Currently bridge CNI only support access port(untagged only) or trunk port(tagged only)
+	if vlanID != 0 {
 		err = netlink.BridgeVlanAdd(hostVeth, uint16(vlanID), true, true, false, true)
 		if err != nil {
 			return nil, nil, fmt.Errorf("failed to setup vlan tag on interface %q: %v", hostIface.Name, err)
 		}
+	} else {
```
mlguerrero12, line 463: you don't need this else

tjjh89017: because we block the condition in the early condition?

tjjh89017: Ok remove it.

maiqueb (commented):
just some opinionated nits.

plugins/main/bridge/bridge.go
tjjh89017: I will consider to add detail condition to make sure others could read it and understand it immediately.

mlguerrero12: the 2 of them could be empty, the != nil is needed I think

tjjh89017: yep, because it's switch-case
i'm not so sure about that.

maiqueb: we would need more unit tests then, to assert this works as intended :)

tjjh89017: I take a look in test.go
but I'm not sure how to add a unit test without in cmdAddDel Test case hmm.

tjjh89017: all test is via cmdAddDelTest.
So it's not easy to insert an unit test for collectVlanTrunk.

maiqueb: I would just add a new context in the describe block and have something like:

```golang

DescribeTable(
    "collectVlanTrunk succeeds", 
    func(vlanTrunks ...*vlanTrunk, expectedVIDs []int) {
        Expect(vlancollectVlanTrunk(vlanTrunks)).To(Equal(expectedVIDs))
    }, 
    Entry("", &vlanTrunk{}, []int{}),
    Entry("", &vlanTrunk{minID: 100, maxID: 105}, []int{100, 101, 102, 103, 104, 105}),
...
)

DescribeTable(
    "collectVlanTrunk **fails**", 
    func(vlanTrunks ...*vlanTrunk, expectedError error) {
        Expect(vlancollectVlanTrunk(vlanTrunks)).To(MatchError(expectedError))
    }, 
    Entry("", &vlanTrunk{minID: 100}, fmt.Errorf("minID and maxID should be configured simultaneously, maxID is missing")),
    Entry("", &vlanTrunk{maxID: 100}, "minID and maxID should be configured simultaneously, minID is missing"),
    Entry("", &vlanTrunk{minID: -100, maxID: 105}, "incorrect trunk minID parameter"),
    Entry("", &vlanTrunk{minID: 10000, maxID: 105}, "incorrect trunk minID parameter"),
    Entry("", &vlanTrunk{minID: 100, maxID: 10000}, "incorrect trunk minID parameter"),
    Entry("", &vlanTrunk{minID: 100, maxID: -1}, "incorrect trunk minID parameter"),
...
...
)

```

plugins/main/bridge/bridge_test.go
```diff
@@ -1672,6 +1764,25 @@ var _ = Describe("bridge Operations", func() {
...
+		correctMaxID int = 105
+	)
+
+	DescribeTable(
+		"collectVlanTrunk succeeds",
+		func(vlanTrunks []*VlanTrunk, expectedVIDs []int) {
+			Expect(collectVlanTrunk(vlanTrunks)).To(Equal(expectedVIDs))
+		},
+		Entry("", &[]*VlanTrunk{}, []int{}),
+		Entry("", &[]*VlanTrunk{
```
maiqueb, line 1778: please fill the entry descriptions; otherwise, if the tests fail, we will not know which of them has failed.

maiqueb (changes requested):
missing descriptions in the table `Entry`s and also missing the "error" scenario for checking `collectVlanTrunk` fails when provided wrong input.

plugins/main/bridge/bridge_test.go
tjjh89017: I'm trial and error right now.
I'm not sure the reason I cannot test it on my PC.
So I'm asking github action to help.
Please hold on.

tjjh89017: Sorry to bother.
I used another branch to test right now.

tjjh89017: I used my repo to run action.
so it works to test.

maiqueb, line 1778: The type of vlanTrunk is wrong.

```suggestion
		Entry("", []*VlanTrunk{}, []int{}),
		Entry("", []*VlanTrunk{
```

plugins/main/bridge/bridge_test.go
```diff
@@ -1672,6 +1764,25 @@ var _ = Describe("bridge Operations", func() {
...
 
+	var (
+		correctMinID int = 100
+		correctMaxID int = 105
+	)
+
+	DescribeTable(
+		"collectVlanTrunk succeeds",
+		func(vlanTrunks []*VlanTrunk, expectedVIDs []int) {
+			Expect(collectVlanTrunk(vlanTrunks)).To(Equal(expectedVIDs))
```
maiqueb, line 1775:
```suggestion
			Expect(collectVlanTrunk(vlanTrunks)).To(ConsistOf(expectedVIDs))
```

otherwise you get the following error:
```
  [FAILED] Expected
      <[]int | len:0, cap:0>: nil
  to equal
      <[]int | len:0, cap:0>: []
  In [It] at: /home/mduarted/github/cni-plugins/plugins/main/bridge/bridge_test.go:1775
```

maiqueb (changes requested):

tjjh89017:
@maiqueb I think this version of unit test should work.
Please help me to review
Thanks a lot!

plugins/main/bridge/bridge_test.go
tjjh89017: Thanks for advice
this one is better.

maiqueb: what you had (ensure the return val is `nil` is also OK). You can use it if you prefer.

plugins/main/bridge/bridge_test.go
```diff
@@ -1672,6 +1764,41 @@ var _ = Describe("bridge Operations", func() {
...
+			_, err := collectVlanTrunk(vlanTrunks)
+			Expect(err).To(MatchError(expectedError))
+		},
+		Entry("when not passed the maxID", []*VlanTrunk{{MinID: &correctMinID}}, fmt.Errorf("minID and maxID should be configured simultaneously, maxID is missing")),
+		Entry("when not passed the minID", []*VlanTrunk{{MaxID: &correctMaxID}}, fmt.Errorf("minID and maxID should be configured simultaneously, minID is missing")),
+		Entry("when the minID is negative", []*VlanTrunk{{MinID: &negativeID, MaxID: &correctMaxID}}, fmt.Errorf("incorrect trunk minID parameter")),
+		Entry("when the minID is larger than 4094", []*VlanTrunk{{MinID: &overID, MaxID: &correctMaxID}}, fmt.Errorf("incorrect trunk minID parameter")),
+		Entry("when the maxID is larger than 4094", []*VlanTrunk{{MinID: &correctMinID, MaxID: &overID}}, fmt.Errorf("incorrect trunk maxID parameter")),
+		Entry("when the maxID is negative", []*VlanTrunk{{MinID: &correctMinID, MaxID: &overID}}, fmt.Errorf("incorrect trunk maxID parameter")),
+	)
```
maiqueb, line 1800: please add a test where the maxID < minID to assert that error.

plugins/main/bridge/bridge.go
tjjh89017: Unit tests are included in latest commit

maiqueb (changes requested):
Nice, I'm just missing a few tests.

Also missing tests for the `vlanTrunk`.ID attribute; I would say:
- positive tests: a vlan trunk w/ just the ID
- positive tests: a vlan trunk w/ a range + ID
- negative test: a a vlan trunk where the ID has a negative value
- negative test: a a vlan trunk where the ID == 10000

plugins/main/bridge/bridge_test.go
tjjh89017: got it.

tjjh89017: added more tests.

maiqueb (approved):
Thank you. 

Please squash all commits into one once the PR is ready. 

@mlguerrero12 can you take another look ?

tjjh89017:
> Thank you.
> 
> Please squash all commits into one once the PR is ready.
> 
> @mlguerrero12 can you take another look ?

I will squash commit in the last moment.
Please notify me if all reviewers accept and it's ready to merge.
Thanks a lot.

mlguerrero12:
thank you @tjjh89017

tjjh89017:
Ok thanks all of you guys!
I already squashed all commits into one.
Please check it.
Thanks again

tjjh89017:
> please don't forget to update the docs, the example sets both vlan and vlantrunk

Sure, I will!

squeed (approved):

PR #831, opened 2023-02-16
Fix overwritten error var in getMTUByName

### LINKED ISSUE #830: panic when master dev doesn't exist within a container
macvlan and vlan plugins panic when the master interface within a container does not exist.

mlguerrero12:
this prevents the error to be lost which was causing the panic while accesing a nil var.

Fix #830

Signed-off-by: Marcelo Guerrero Viveros <marguerr@redhat.com>


mccv1r0 (approved):
/lgtm

PR #832, opened 2023-02-17
tap: allow for a tap device to be created as a bridge port

maiqueb:
This extends the tap plugin API enabling the user to instruct the CNI plugin the created tap device must be set as a port of an *existing* linux bridge on the pod network namespace.

This is helpful for KubeVirt, allowing network connectivity to be extended from the pod's interface into the Virtual Machine running inside the pod.

### DIFF
--- a/plugins/main/tap/tap.go
+++ b/plugins/main/tap/tap.go
@@ -47,6 +47,7 @@ type NetConf struct {
 	Owner          *uint32   `json:"owner,omitempty"`
 	Group          *uint32   `json:"group,omitempty"`
 	SelinuxContext string    `json:"selinuxContext,omitempty"`
+	Bridge         string    `json:"bridge,omitempty"`
 	Args           *struct{} `json:"args,omitempty"`
 	RuntimeConfig  struct {
 		Mac string `json:"mac,omitempty"`
@@ -214,6 +215,18 @@ func createTap(conf *NetConf, ifName string, netns ns.NetNS) (*current.Interface
 		link, err := netlink.LinkByName(ifName)
 		if err != nil {
 			return fmt.Errorf("failed to refetch tap %q: %v", ifName, err)
+		}
+
+		if conf.Bridge != "" {
+			bridge, err := netlink.LinkByName(conf.Bridge)
+			if err != nil {
+				return fmt.Errorf("failed to get bridge %s: %v", conf.Bridge, err)
+			}
+
+			tapDev := link
+			if err := netlink.LinkSetMaster(tapDev, bridge); err != nil {
+				return fmt.Errorf("failed to set tap %s as a port of bridge %s: %v", tap.Name, conf.Bridge, err)
+			}
 		}
 
 		err = netlink.LinkSetUp(link)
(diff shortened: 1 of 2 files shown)


tjjh89017:
@maiqueb Hi
I'm curious to know about the usage of this TAP CNI.
IIRC, tap device will be inside pod.
So this PR will ask CNI to set Tap's master to bridge inside the same pod.
Is it correct?
thanks a lot.

maiqueb:
> @maiqueb Hi I'm curious to know about the usage of this TAP CNI. IIRC, tap device will be inside pod. So this PR will ask CNI to set Tap's master to bridge inside the same pod. Is it correct? thanks a lot.

Yes, both the tap device **and** the bridge must be inside the same network namespace (e.g. the pod).

squeed (approved):

squeed:
This looks good, but could you rebase real quick?

maiqueb:
@squeed 

> This looks good, but could you rebase real quick?

Done. Thanks for the quick review / approval.

maiqueb:
@mlguerrero12 IIUC this PR is failing on the lint stage for something unrelated to it.

Should I rebase #871 to address those ?

maiqueb:
@dcbw / @squeed can we proceed with a merge for this PR ? 

... I now need #871 in before, to address the lint errors.

mlguerrero12:
gingko-linter was just merged and this introduced new errors

maiqueb:
> gingko-linter was just merged and this introduced new errors

wow :sweat_smile: I usually point these ones out ... Should not have blindly copied an existing test .

Let's see if it is now OK.

maiqueb:
OK, now I am not introducing any errors, but seems we now pull an additional linter that frowns upon the way ginkgo is used.

... don't get me wrong, I think the linter is actually right. 

... but I also think it should complain exclusively of code introduced in this PR, not existing code.

mlguerrero12:
Correct, I was not referring to the new test of this pr, but just to let you know that we have now new issues in general.

maiqueb:
Is this somehow a [flake](https://github.com/containernetworking/plugins/actions/runs/4608894105/jobs/8145369318?pr=832) @squeed ?
```
Summarizing 1 Failure:
  [FAIL] chain tests [It] deletes chains idempotently in parallel
  /home/runner/work/plugins/plugins/plugins/meta/portmap/chain_test.go:224
```

I find it very strange that the contents of this PR (main tap plugin) somehow impact the meta portmap plugin ...

I'm not touching any library.

plugins/main/tap/tap.go
```diff
@@ -210,6 +211,22 @@ func createTap(conf *NetConf, ifName string, netns ns.NetNS) (*current.Interface
 		}
 		tap.Name = ifName
 
+		if conf.Bridge != "" {
```
mlguerrero12, line 214: why don't you add this block after re-fetching the link and set it up? You'll save one LinkByName.

maiqueb: Good catch, I didn't notice it. 

Thanks for pointing it out.
@mlguerrero12

maiqueb:
I have no clue as to why the linter fails, but, from looking at my source code, I don't think that's the issue. 

I don't think the linter should complain about code that is not being modified in a certain PR.

mlguerrero12 (approved):

dcbw:
Thanks!

PR #852, opened 2023-03-13
bridge, spoofcheck: Apply nftable rules in the pod netns

EdDev:
The spoofcheck nftables rules have been applied on the node root network namespace for ingress traffic that arrives from the pod netns.

When there are a large amount of rules visible, the user-space processing of adding and removing rules becomes slow. Field results showed the `nft` commands taking minutes to complete.

In addition, the removal of the rules is dependent on the execution of the DEL command, which is not always assured to occur. Therefore, the current solution where the rules are set on the node root network namespace may cause nftables configuration leaks.

This change suggests to place the rules in the pod network namespace, solving both issues.

It adds rules at two places: IP and Bridge stacks, on the postrouting hook.

### DIFF
--- a/pkg/link/spoofcheck.go
+++ b/pkg/link/spoofcheck.go
@@ -23,8 +23,8 @@ import (
 )
 
 const (
-	natTableName            = "nat"
-	preRoutingBaseChainName = "PREROUTING"
+	natTableName             = "nat"
+	postRoutingBaseChainName = "POSTROUTING"
 )
 
 type NftConfigurer interface {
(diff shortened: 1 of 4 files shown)


EdDev:
> So far so good. Now my question is how will you know if a pod interface (which is created via CNI) is a port of a bridge that will be created out of band (**after** CNI).

My plan is to add the rule to both families, no matter what the usage is.
As you mentioned, the CNI cannot know how it will be used after it finished.

maiqueb:
> > So far so good. Now my question is how will you know if a pod interface (which is created via CNI) is a port of a bridge that will be created out of band (**after** CNI).
> 
> My plan is to add the rule to both families, no matter what the usage is. As you mentioned, the CNI cannot know how it will be used after it finished.

Another thing; are you sure the nft `inet` family can match mac addresses ? ... 

I don't think it can ..

EdDev:
> Another thing; are you sure the nft `inet` family can match mac addresses ? ...
> 
> I don't think it can .. (but I'll gladly be corrected if it is indeed possible).

Based on the test you conducted, it seems it can't.

PR #856, opened 2023-03-18
Added trivy-scan for plugins

yashsingh74:
This PR introduces the security scanning for plugins repository. This job will run once any pull and push request changes made on the `main` branch.

**1. Why is this pull request needed and what does it do?**
The CVE scanning for the plugins repository will be enabled by this PR. It will make it easier to track down the HIGH and CRITICAL CVE being used.

**Test result:**
Screenshot 2023-03-18 at 8 55 14 PM

### DIFF
--- a/.github/workflows/trivy-scan.yaml
+++ b/.github/workflows/trivy-scan.yaml
@@ -1,0 +1,28 @@
+name: Trivy Security Scan
+on:
+  push:
+    branches: ["main"]
+  pull_request:
+    branches: ["main"]
+
+jobs:
+  build:
+    name: build
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+
+      - name: Run Trivy vulnerability scanner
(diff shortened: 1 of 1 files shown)


yashsingh74:
cc: @squeed @dcbw

squeed:
I'm willing to try this; but I'm concerned about endless false positives. the CNI plugins make no HTTP or TLS requests, provide no network services, and and don't parse end-user-provided input. Dealing with the output of security scanners is, in my experience, a tiring exercise in pulling signal from noise.

CNI, as a volunteer-operated project, only has so many resources, and dealing with spurious false positives might not be a good use of that time.

Does trivy do any sort of code path analysis to determine if CVEs are applicable in the style of govulncheck?

yashsingh74:
Yes, we are not testing any HTTP or TLS requests. But this will help to test the overall code. 
Also, it does make some noise but would be good to look on it as there are so many different user who will be contributing or using the plugins. 
Overall it will help to catch the CVEs before merging any new PRs.

yashsingh74:
@squeed Please review the PR.

PR #865, opened 2023-03-23
Remove default native vlan on bridge

mlguerrero12:
These changes prevent pods connected to a VLAN
to receive the native traffic of the bridge.

Fixes: #667

### DIFF
--- a/plugins/main/bridge/bridge.go
+++ b/plugins/main/bridge/bridge.go
@@ -256,6 +256,9 @@ func bridgeByName(name string) (*netlink.Bridge, error) {
 }
 
 func ensureBridge(brName string, mtu int, promiscMode, vlanFiltering bool) (*netlink.Bridge, error) {
+	// Set to 0 to remove the default native vlan
+	var defaultPVID uint16
+
 	br := &netlink.Bridge{
 		LinkAttrs: netlink.LinkAttrs{
 			Name: brName,
@@ -269,6 +272,7 @@ func ensureBridge(brName string, mtu int, promiscMode, vlanFiltering bool) (*net
 	}
 	if vlanFiltering {
 		br.VlanFiltering = &vlanFiltering
+		br.VlanDefaultPVID = &defaultPVID
 	}
 
 	err := netlink.LinkAdd(br)
@@ -287,6 +291,14 @@ func ensureBridge(brName string, mtu int, promiscMode, vlanFiltering bool) (*net
 	br, err = bridgeByName(brName)
 	if err != nil {
 		return nil, err
+	}
+
+	// Remove native vlan in case the bridge already exists and has one set
+	if vlanFiltering && *br.VlanDefaultPVID != defaultPVID {
+		err = netlink.BridgeSetVlanDefaultPVID(br, defaultPVID)
+		if err != nil {
+			return nil, fmt.Errorf("could not remove native vlan %q: %v", brName, err)
+		}
 	}
 
 	// we want to own the routes for this interface
--- a/plugins/ipam/host-local/backend/disk/backend.go
+++ b/plugins/ipam/host-local/backend/disk/backend.go
@@ -118,7 +118,6 @@ func (s *Store) FindByID(id string, ifname string) bool {
 	s.Lock()
 	defer s.Unlock()
 
-	found := false
 	match := strings.TrimSpace(id) + LineBreak + ifname
 	found, err := s.FindByKey(match)
 
@@ -155,7 +154,6 @@ func (s *Store) ReleaseByKey(match string) (bool, error) {
 // N.B. This function eats errors to be tolerant and
 // release as much as possible
 func (s *Store) ReleaseByID(id string, ifname string) error {
-	found := false
 	match := strings.TrimSpace(id) + LineBreak + ifname
 	found, err := s.ReleaseByKey(match)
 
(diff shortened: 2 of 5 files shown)


plugins/main/bridge/bridge.go
```diff
@@ -289,6 +293,14 @@ func ensureBridge(brName string, mtu int, promiscMode, vlanFiltering bool) (*net
 		return nil, err
 	}
 
+	// Remove native vlan in case the bridge already exists and has one set
+	if vlanFiltering && *br.VlanDefaultPVID != defaultPVID {
+		err = netlink.BridgeSetVlanDefaultPVID(br, defaultPVID)
```
phoracek, line 298: This is problematic - the bridge may be managed by something else (e.g. NetworkManager) and the default PVID may be set intentionally. I think that unless asked to, the CNI should only manipulate the device it creates (veth), not the bridge it connects to

phoracek (changes requested):

plugins/main/bridge/bridge.go
mlguerrero12: We can say the same for promiscuous mode which is above. It's also changed even when the bridge is present.   
 Also, the bridge might have been created by the cni itself from a different attachment.

Anyway, I agree with you, something doesn't add up. I was expecting the function "bridgeByName" to also check that the attributes are the same and return and error otherwise, but it only checks whether the device is a bridge or not. In my opinion, that should be the case to force external managed bridge to comply with the requested configuration and/or to prevent different attachments to change the bridge configuration. The first attachment would set the configuration.

What do you think? c.c. @maiqueb

maiqueb:
Yup, from my manual tests seems this works.

# CNI configuration
```json
{
  "cniVersion": "0.4.0",
  "name": "override-default-pvid",
  "plugins": [
    {
        "type": "bridge",
        "bridge": "mynet0",
        "vlan": 300
    }
  ]
}

```

# Without this fix:
## CNI result
```json
{
    "cniVersion": "0.4.0",
    "interfaces": [
        {
            "name": "mynet0",
            "mac": "e6:77:3b:87:f4:d6"
        },
        {
            "name": "vethd6bf70bc",
            "mac": "be:81:8e:d8:c6:4d"
        },
        {
            "name": "ens123",
            "mac": "b2:9d:9f:45:09:4f",
            "sandbox": "/var/run/netns/defaultpvid"
        }
    ],
    "dns": {}
}
```

## bridge vlan
```
bridge vlan                
port              vlan-id  
mynet0            1 PVID Egress Untagged
vethd6bf70bc      1 Egress Untagged
                  300 PVID Egress Untagged
```

# With this fix:
## CNI result
```json
{
    "cniVersion": "0.4.0",
    "interfaces": [
        {
            "name": "mynet0",
            "mac": "7e:c2:ba:15:50:63"
        },
        {
            "name": "veth444c6f40",
            "mac": "7e:c2:ba:15:50:63"
        },
        {
            "name": "ens123",
            "mac": "9e:53:9b:9f:42:a8",
            "sandbox": "/var/run/netns/defaultpvid"
        }
    ],
    "dns": {}
}
```
## bridge vlan
```
bridge vlan
port              vlan-id  
veth444c6f40      300 PVID Egress Untagged
```

@phoracek

Thanks for the amazing work @mlguerrero12

plugins/main/bridge/bridge.go
maiqueb: For transparency, we discussed offline if it would make sense to expose yet another knob allowing the user to configure the default PVID of the bridge. This would be a somewhat cumbersome way of preserving the PVID of the bridge - which may be valid in some use cases.

I didn't understand if that would happen on this PR, or on a follow-up @mlguerrero12 .

phoracek: I would argue that this "default-pvid" parameter should be only applied on ports. Since the CNI creates these ports, it will be safe to do. In this case, CNI's "default-pvid" attribute can be simply a boolean: true -> do nothing, false -> strip the default pvid from the port.

The problem with changing anything on the bridge itself is, that if the bridge is managed by something other than the CNI (e.g. NetworkManager), we may end up with two processes setting bridge's attributes up and down.

maiqueb: Conceptually, I agree. 

But I think that if the user has a way to specify the default pvid via CNI conf, it possible to persist the PVID value of the bridge it doesn't own. 

... sure, that's cumbersome, involves extra-steps, and may be error prone (a user forgets to do this, for instance).

Having said that, I think this alternative is the simplest thing we can do to tackle the issue; we'll discuss this a bit more, see if we can come up w/ a better alternative though. 

Maybe disallowing CNI from setting args on the bridge when it finds the bridge **already exists** (i.e. it would be able to attach veths to it) would work ?... I still fear that could cause regressions elsewhere.

mlguerrero12: will create another pr with a different approach (following @phoracek suggestion. It's what nmstate does if I'm not mistaken)

mlguerrero12:
@phoracek @maiqueb, please have a look at https://github.com/containernetworking/plugins/pull/875. 

It adds a new parameter to disable the default vlan. The default value is false which preserves the current behavior. It doesn't need changes in netlink and only deals with the config of the bridge ports and not the global config of the bridge.

PR #871, opened 2023-03-27
Fix wastedassign linter errors

mlguerrero12:

PR #873, opened 2023-03-29
bridge, spoof check: remove drop rule index

maiqueb:
Rules are appendend by default, thus using an index is redundant. Using an index also requires the full NFT cache, which causes a CNI ADD to be extremely slow.


maiqueb:
The failed lint error is fixed by #871 ; we should merge it asap.

Let me know if I should rebase it.

squeed (approved):

Issue #866, opened 2023-03-24
VRF plugin is not populating routes


korroot:
Hi,
I wanted to use VRF plugin chained with Static IPAM, an example configuration of NAD:

```
apiVersion: k8s.cni.cncf.io/v1
kind: NetworkAttachmentDefinition
metadata:
  name: default-access
  namespace: default
spec:
  config: |-
    {
      "cniVersion": "0.3.1",
      "plugins": [
        {
          "ipam": {
            "addresses": [
              {
                "address": "192.168.2.10/24"
              }
            ],
            "routes": [
              {
                "dst": "192.168.0.0/24",
                "gw": "192.168.2.1"
              }
            ],
            "type": "static"
          },
          "name": "access"
        },
        {
          "type": "vrf",
          "vrfname": "vrf-access"
        }
      ]
    }
```

The expected result (from inside container):
```
/ # ip r show table 1
192.168.0.0/24 via 192.168.2.1 dev access
192.168.2.0/24 dev access scope link  src 192.168.2.10
local 192.168.2.10 dev access scope host  src 192.168.2.10
broadcast 192.168.2.255 dev access scope link  src 192.168.2.10
```

Actual result:
```
/ # ip r show table 1
192.168.2.0/24 dev access scope link  src 192.168.2.10
local 192.168.2.10 dev access scope host  src 192.168.2.10
broadcast 192.168.2.255 dev access scope link  src 192.168.2.10
```
So missing route added by static IPAM...

PR #874, opened 2023-03-31
Add routes propagation for VRF plugin

korroot:
Up until now, if previous plugin assigned routes to interface, movement of this interface to new VRF cause routes to be deleted.

This patch adds funtionality to VRF plugin to save the routes before interface is assgined to VRF, and then re-apply all saved routes to new VRF.

Closes: #866

### DIFF
--- a/plugins/meta/vrf/vrf.go
+++ b/plugins/meta/vrf/vrf.go
@@ -104,6 +104,19 @@ func addInterface(vrf *netlink.Vrf, intf string) error {
 	if err != nil {
 		return fmt.Errorf("failed getting ipv6 addresses for %s", intf)
 	}
+
+	// Save all routes that are not local and connected, before setting master,
+	// because otherwise those routes will be deleted after interface is moved.
+	filter := &netlink.Route{
+		LinkIndex: i.Attrs().Index,
+		Scope:     netlink.SCOPE_UNIVERSE, // Exclude local and connected routes
+	}
+	filterMask := netlink.RT_FILTER_IIF | netlink.RT_FILTER_SCOPE // Filter based on link index and scope
+	routes, err := netlink.RouteListFiltered(netlink.FAMILY_ALL, filter, filterMask)
+	if err != nil {
+		return fmt.Errorf("failed getting all routes for %s", intf)
+	}
+
 	err = netlink.LinkSetMaster(i, vrf)
 	if err != nil {
 		return fmt.Errorf("could not set vrf %s as master of %s: %v", vrf.Name, intf, err)
@@ -127,6 +140,18 @@ CONTINUE:
 		err = netlink.AddrAdd(i, &toFind)
 		if err != nil {
 			return fmt.Errorf("could not restore address %s to %s @ %s: %v", toFind, intf, vrf.Name, err)
+		}
+	}
+
+	// Apply all saved routes for the interface that was moved to the VRF
+	for _, route := range routes {
+		r := route
+		// Modify original table to vrf one,
+		r.Table = int(vrf.Table)
+		// equivalent of 'ip route replace <address> table <int>'.
+		err = netlink.RouteReplace(&r)
+		if err != nil {
+			return fmt.Errorf("could not add route '%s': %v", r, err)
 		}
 	}
 
--- a/plugins/meta/vrf/vrf_test.go
+++ b/plugins/meta/vrf/vrf_test.go
@@ -17,10 +17,13 @@ package main
 import (
 	"encoding/json"
 	"fmt"
+	"net"
+	"strings"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 	"github.com/vishvananda/netlink"
+	"golang.org/x/sys/unix"
 
 	"github.com/containernetworking/cni/pkg/skel"
 	"github.com/containernetworking/cni/pkg/types"
@@ -107,7 +110,7 @@ var _ = Describe("vrf plugin", func() {
 				},
 			})
 			Expect(err).NotTo(HaveOccurred())
-			_, err = netlink.LinkByName(IF0Name)
+			_, err = netlink.LinkByName(IF1Name)
 			Expect(err).NotTo(HaveOccurred())
 			return nil
 		})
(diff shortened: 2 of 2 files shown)


squeed (changes requested):
Really nice!
Can you add a test case that includes both address families?

korroot:
Added IPv6 routing test. All works fine.

squeed:
I talked to @mlguerrero12 , who observed that this is a behavior change, and might break existing setups. He's going to run some internal tests on it.

If so, we may need to put this behind a configuration variable, something like `preserveRoutes`.

Separately, can you fix the lint / format issues?

korroot:
@squeed Lint error fixed.

In terms of behavior change - does it mean that deleting the routing in some cases is desirable? 
I am not sure how exactly this configuration variable to enable/disable routes propagation should be implemented?

plugins/meta/vrf/vrf.go
```diff
@@ -104,6 +104,14 @@ func addInterface(vrf *netlink.Vrf, intf string) error {
 	if err != nil {
 		return fmt.Errorf("failed getting ipv6 addresses for %s", intf)
 	}
+
+	// Save routes that are setup for the interface, before setting master,
+	// because otherwise the routes will be deleted after interface is moved.
+	routes, err := netlink.RouteList(i, netlink.FAMILY_ALL)
+	if err != nil {
+		return fmt.Errorf("failed getting ipv4 routes for %s", intf)
```
mlguerrero12, line 112: not only ipv4

plugins/meta/vrf/vrf.go
```diff
@@ -130,6 +138,14 @@ CONTINUE:
 		}
 	}
 
+	// Apply all saved routes for the interface that was moved to the VRF
+	for _, route := range routes {
+		// modify original table to vrf one
+		// equivalent of 'ip route add <address> table <int>'
+		route.Table = int(vrf.Table)
+		netlink.RouteAdd(&route)
```
mlguerrero12, line 146: unhandled error

mlguerrero12, line 146: to be on the safe side, create a local copy of route inside the for loop

korroot: Yes, I noticed it before, but it is not handled by purpose - the error can be ignored, because if adding a route returns error - it can mean that is already there.

mlguerrero12: in that case, to be consistent with the logic of ipv6 addresses which is above this block, only add the routes that do not exist

mlguerrero12: Then check if the error is because the route is already there.  You can't assume that the error will always be due to that. A user might complain that the routes are not being added and we won't have logs to troubleshoot.

korroot: Ok, I have added a check, if route exists, and if not, return error during adding a route.

plugins/meta/vrf/vrf.go
```diff
@@ -130,6 +139,21 @@ CONTINUE:
...
+	// Apply all saved routes for the interface that was moved to the VRF
+	for _, route := range routes {
+		r := route
+		// Modify original table to vrf one,
+		// equivalent of 'ip route add <address> table <int>'.
+		r.Table = int(vrf.Table)
+		err = netlink.RouteAdd(&r)
+		if err != nil {
+			// If route is already present, returned error is "file exists"
+			if !strings.Contains(fmt.Sprintf("%v", err), "file exists") {
```
mlguerrero12, line 151: does err.Error() not give you the string?

plugins/meta/vrf/vrf.go
```diff
@@ -130,6 +139,21 @@ CONTINUE:
...
+				return fmt.Errorf("error while adding route \"%s\": %v\n", r, err)
```
mlguerrero12, line 152: no need to add \n and perhaps something like "failed adding ..." or "could not add ..." would be more consistent with other errors in this file

plugins/meta/vrf/vrf.go
```diff
@@ -130,6 +139,21 @@ CONTINUE:
 		}
 	}
 
+	// Apply all saved routes for the interface that was moved to the VRF
+	for _, route := range routes {
+		r := route
+		// Modify original table to vrf one,
+		// equivalent of 'ip route add <address> table <int>'.
+		r.Table = int(vrf.Table)
+		err = netlink.RouteAdd(&r)
```
mlguerrero12, line 148: you can use netlink.RouteReplace instead (ip route replace)

korroot: Hmm what do we get by using RouteReplace? I think if the route is there, we maybe not suppose to replace it...

korroot: Make sense, thanks for the explanation. I will address it shortly.

korroot:
@mlguerrero12 your latest comment is addressed, thanks for improving this PR :) 
Do you see something more to adjust, or are we closer to merge :D ?

plugins/meta/vrf/vrf.go
```diff
@@ -130,6 +143,18 @@ CONTINUE:
 		}
 	}
 
+	// Apply all saved routes for the interface that was moved to the VRF
+	for _, route := range routes {
+		r := route
+		// Modify original table to vrf one,
+		r.Table = int(vrf.Table)
+		// equivalent of 'ip route replace <address> table <int>'.
+		err = netlink.RouteReplace(&r)
```
mlguerrero12, line 152: I think you can use add here now since you are making sure to copy only routes that are not supposed to exist. In case there is a conflict (error), users should be aware of it and fix it.

korroot: I tried to use the RouteAdd here but it seems that one of IPv6 route is copied to saved routes:

```
  [FAILED] Unexpected error:
      <*errors.errorString | 0xc0001b56a0>: {
          s: "cmdAdd failed: could not add route '{Ifindex: 2 Dst: abcd:1234:ffff::/64 Src: <nil> Gw: <nil> Flags: [] Table: 1 Realm: 0}': file exists",
      }
      cmdAdd failed: could not add route '{Ifindex: 2 Dst: abcd:1234:ffff::/64 Src: <nil> Gw: <nil> Flags: [] Table: 1 Realm: 0}': file exists
  occurred
```

From the vrf_test.go: https://github.com/containernetworking/plugins/pull/874/files#diff-d8b21f1b7cbf938511615ea01f7b56215325cb65d64b45b5cd089e01c210bf6eR197

So instead I have used the RouteReplace

korroot:
@squeed can you re-review this PR? Also, can somebody re-run the CI? it broke on timeout, not code issues.

korroot:
@squeed thanks for running the CI, can you approve this PR? Or comment if there are any more improvements needed?

squeed (approved):

PR #875, opened 2023-03-31
Add parameter to disable default vlan

mlguerrero12:
This new parameter allows users to remove the default vlan
    
Fixes: #667

### DIFF
--- a/plugins/main/bridge/bridge.go
+++ b/plugins/main/bridge/bridge.go
@@ -46,17 +46,18 @@ const defaultBrName = "cni0"
 
 type NetConf struct {
 	types.NetConf
-	BrName       string `json:"bridge"`
-	IsGW         bool   `json:"isGateway"`
-	IsDefaultGW  bool   `json:"isDefaultGateway"`
-	ForceAddress bool   `json:"forceAddress"`
-	IPMasq       bool   `json:"ipMasq"`
-	MTU          int    `json:"mtu"`
-	HairpinMode  bool   `json:"hairpinMode"`
-	PromiscMode  bool   `json:"promiscMode"`
-	Vlan         int    `json:"vlan"`
-	MacSpoofChk  bool   `json:"macspoofchk,omitempty"`
-	EnableDad    bool   `json:"enabledad,omitempty"`
+	BrName              string `json:"bridge"`
+	IsGW                bool   `json:"isGateway"`
+	IsDefaultGW         bool   `json:"isDefaultGateway"`
+	ForceAddress        bool   `json:"forceAddress"`
+	IPMasq              bool   `json:"ipMasq"`
+	MTU                 int    `json:"mtu"`
+	HairpinMode         bool   `json:"hairpinMode"`
+	PromiscMode         bool   `json:"promiscMode"`
+	Vlan                int    `json:"vlan"`
+	PreserveDefaultVlan bool   `json:"preserveDefaultVlan"`
+	MacSpoofChk         bool   `json:"macspoofchk,omitempty"`
+	EnableDad           bool   `json:"enabledad,omitempty"`
 
 	Args struct {
 		Cni BridgeArgs `json:"cni,omitempty"`
@@ -94,6 +95,8 @@ func init() {
 func loadNetConf(bytes []byte, envArgs string) (*NetConf, string, error) {
 	n := &NetConf{
 		BrName: defaultBrName,
+		// Set default value equal to true to maintain existing behavior.
+		PreserveDefaultVlan: true,
 	}
 	if err := json.Unmarshal(bytes, n); err != nil {
 		return nil, "", fmt.Errorf("failed to load netconf: %v", err)
@@ -299,7 +302,7 @@ func ensureBridge(brName string, mtu int, promiscMode, vlanFiltering bool) (*net
 	return br, nil
 }
 
-func ensureVlanInterface(br *netlink.Bridge, vlanID int) (netlink.Link, error) {
+func ensureVlanInterface(br *netlink.Bridge, vlanID int, preserveDefaultVlan bool) (netlink.Link, error) {
 	name := fmt.Sprintf("%s.%d", br.Name, vlanID)
 
 	brGatewayVeth, err := netlink.LinkByName(name)
@@ -313,7 +316,7 @@ func ensureVlanInterface(br *netlink.Bridge, vlanID int) (netlink.Link, error) {
 			return nil, fmt.Errorf("faild to find host namespace: %v", err)
 		}
 
-		_, brGatewayIface, err := setupVeth(hostNS, br, name, br.MTU, false, vlanID, "")
+		_, brGatewayIface, err := setupVeth(hostNS, br, name, br.MTU, false, vlanID, preserveDefaultVlan, "")
 		if err != nil {
 			return nil, fmt.Errorf("faild to create vlan gateway %q: %v", name, err)
 		}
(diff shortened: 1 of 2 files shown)


plugins/main/bridge/bridge.go
```diff
@@ -374,11 +375,47 @@ func setupVeth(netns ns.NetNS, br *netlink.Bridge, ifName string, mtu int, hairp
...
+func removeDefaultVlan(hostVeth netlink.Link, vlanID int) error {
+	vlanInfo, err := netlink.BridgeVlanList()
+	if err != nil {
+		return err
+	}
+
+	var defaultVlan uint16
+out:
+	for deviceID, brVlanInfo := range vlanInfo {
+		if deviceID == int32(hostVeth.Attrs().Index) {
```
maiqueb, line 399: why do we iterate the map when we can just get the corresponding slice from it ? 
Something like: 
```golang
brVlanInfo, isDeviceIDPresent := vlanInfo[int32(hostVeth.Attrs().Index)]
if !isDeviceIDPresent {
    return nil
}
...
```

Am I missing something ?

plugins/main/bridge/bridge.go
```diff
@@ -46,17 +46,18 @@ const defaultBrName = "cni0"
...
+	BrName             string `json:"bridge"`
+	IsGW               bool   `json:"isGateway"`
+	IsDefaultGW        bool   `json:"isDefaultGateway"`
+	ForceAddress       bool   `json:"forceAddress"`
+	IPMasq             bool   `json:"ipMasq"`
+	MTU                int    `json:"mtu"`
+	HairpinMode        bool   `json:"hairpinMode"`
+	PromiscMode        bool   `json:"promiscMode"`
+	Vlan               int    `json:"vlan"`
+	DisableDefaultVlan bool   `json:"disableDefaultVlan"`
```
maiqueb, line 58: could we have a `preserveDefaultVlan` attribute instead ? 

That would allow a simpler configuration when people want to override the default VLAN (... I'm taking an educated guess, but I would say that's the most common use-case) , and I also think (generically speaking) a boolean argument should have "positive" logic - it should be about "enabling" something - enabling a flag to disable a feature is more confusing than the alternative.

@mlguerrero12

plugins/main/bridge/bridge.go
mlguerrero12: Of course, thanks. I'll change it.

plugins/main/bridge/bridge.go
mlguerrero12: yes, I agree. My first choice was enableDefaultVlan, but I was having issues setting the default value equal to true (since the idea is to preserve the current behavior). Today, I tried again and it worked (another reason to not work late on Fridays).

So, you want me to change the name to preserveDefaultVlan, right?

maiqueb: Yes. Maybe someone has a better name for it, but I think that represents the user's intentions. 

Thanks for listening / taking into account this suggestion.

plugins/main/bridge/bridge.go
mlguerrero12: done

plugins/main/bridge/bridge.go
mlguerrero12: done

plugins/main/bridge/bridge_test.go
```diff
@@ -62,24 +62,26 @@ type Net struct {
 	PrevResult    types100.Result        `json:"-"`
 }
 
-// testCase defines the CNI network configuration and the expected
+// testCase defi
```
phoracek, line 65: This seems unintentional

plugins/main/bridge/bridge_test.go
```diff
@@ -62,24 +62,26 @@ type Net struct {
...
+	cniVersion        string      // CNI Version
+	subnet            string      // Single subnet config: Subnet CIDR
+	gateway           string      // Single subnet config: Gateway
+	ranges            []rangeInfo // Ranges list (multiple subnets config)
+	resolvConf        string      // host-local resolvConf file path
+	isGW              bool
+	isLayer2          bool
+	expGWCIDRs        []string // Expected gateway addresses in CIDR form
+	vlan              int
+	removeDefaultVlan bool
```
phoracek, line 78: Outdated?

plugins/main/bridge/bridge.go
```diff
@@ -374,11 +376,45 @@ func setupVeth(netns ns.NetNS, br *netlink.Bridge, ifName string, mtu int, hairp
 		if err != nil {
 			return nil, nil, fmt.Errorf("failed to setup vlan tag on interface %q: %v", hostIface.Name, err)
 		}
+
+		if !preserveDefaultVlan {
```
phoracek, line 380: You could move this above the VLAN add and simply remove all VLAN configuration, without the condition of `info.Vid != uint16(vlanID)`

phoracek (commented):
Minor notes. I like the new API and its defaults. I think this is the right way forward - fixing the behavior so it works as one would expect, while still allowing people who may rely on the faulty side-effect to use it

plugins/main/bridge/bridge_test.go
```diff
@@ -527,6 +535,9 @@ func (tester *testerV10x) cmdAddTest(tc testCase, dataDir string) (types.Result,
 			vlans, isExist := interfaceMap[int32(peerLink.Attrs().Index)]
 			Expect(isExist).To(BeTrue())
 			Expect(checkVlan(tc.vlan, vlans)).To(BeTrue())
+			if tc.removeDefaultVlan {
+				Expect(len(vlans)).To(Equal(1))
```
maiqueb, line 539: nit: it is more idiomatic to do `Expect(vlans).To(HaveLen(1))`.

plugins/main/bridge/bridge.go
```diff
@@ -379,6 +388,25 @@ func setupVeth(netns ns.NetNS, br *netlink.Bridge, ifName string, mtu int, hairp
...
+func removeDefaultVlan(hostVeth netlink.Link) error {
+	vlanInfo, err := netlink.BridgeVlanList()
+	if err != nil {
+		return err
+	}
+
+	brVlanInfo, ok := vlanInfo[int32(hostVeth.Attrs().Index)]
+	if ok {
+		for _, info := range brVlanInfo {
+			err = netlink.BridgeVlanDel(hostVeth, info.Vid, false, true, false, true)
```
maiqueb, line 400: question @mlguerrero12: do any of these boolean attributes even matter for deleting the VLAN ?

mlguerrero12: The pvid and untagged flags are ignored

   bridge vlan delete - delete a vlan filter entry
       This command removes an existing vlan filter entry.

       The arguments are the same as with bridge vlan add.  The pvid and
       untagged flags are ignored.

maiqueb:
Thank you @mlguerrero12. The only think holding back the lgtm from my part is the README update.

plugins/main/bridge/bridge.go
```diff
@@ -93,7 +94,8 @@ func init() {
 
 func loadNetConf(bytes []byte, envArgs string) (*NetConf, string, error) {
 	n := &NetConf{
-		BrName: defaultBrName,
+		BrName:              defaultBrName,
+		PreserveDefaultVlan: true,
 	}
```
s1061123, line 99: How about to add some comments to mention the defalt value, true, comes from backward compatibility, such as 'default is true for compatibility for previous version'

plugins/main/bridge/bridge.go
```diff
@@ -46,17 +46,18 @@ const defaultBrName = "cni0"
...
+	BrName              string `json:"bridge"`
+	IsGW                bool   `json:"isGateway"`
+	IsDefaultGW         bool   `json:"isDefaultGateway"`
+	ForceAddress        bool   `json:"forceAddress"`
+	IPMasq              bool   `json:"ipMasq"`
+	MTU                 int    `json:"mtu"`
+	HairpinMode         bool   `json:"hairpinMode"`
+	PromiscMode         bool   `json:"promiscMode"`
+	Vlan                int    `json:"vlan"`
+	PreserveDefaultVlan bool   `json:"preserveDefaultVlan"`
```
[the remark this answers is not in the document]
squeed: Should this be a `*bool` to preserve the existing behavior?

maiqueb (approved):
@mlguerrero12 I had forgotten the README is defined in a separate project.

This looks good to me. Please address the other reviewers comments.

plugins/main/bridge/bridge_test.go
phoracek: I see, thanks for explaining

mlguerrero12:
@phoracek, @maiqueb, please have a look at https://github.com/containernetworking/cni.dev/pull/119

squeed (approved):

PR #887, opened 2023-04-20
Fix ValidateExpectedRoute with non default routes and nil GW

### LINKED ISSUE #886: CHECK fails with ptp plugin and non default routes
I'm trying to switch from containerd to cri-o, and one big difference seems to be that cri-o call CheckNetworkList() https://github.com/cri-o/ocicni/blob/4fcbe80ee01fbc4c0401ae96537b3ee807c92c2c/pkg/ocicni/ocicni.go#L731

Here an example config that fails with `Error checking network: Expected Route {Dst:{IP:198.18.128.0 Mask:ffff8000} GW:<nil>} not found in routing table`

```
{
  "cniVersion": "0.4.0",
  "name": "mgmt",
  "type": "ptp",
  "ipMasq": true,
  "ipam": {
    "type": "host-local",
    "ranges": [
      [{"subnet": "198.18.0.0/17"}],
      [{"subnet": "fd61:7465:6d65:1000::/112"}]
    ],
    "routes": [
      { "dst": "0.0.0.0/0" },
      { "dst": "198.18.128.0/17" },
      { "dst": "198.19.254.254/32" },
      { "dst": "::/0" },
      { "dst": "fd61:7465:6d65:2000::/112" },
      { "dst": "fd61:7465:6d65:ffff::/128" }
    ]
  }
}
```

Using `"cniVersion": "0.3.1"` to bypass the check, the routes are created but with a gateway
```
# ip r
default via 198.18.0.1 dev eth0 
198.18.0.0/17 via 198.18.0.1 dev eth0 src 198.18.66.190 
198.18.0.1 dev eth0 scope link src 198.18.66.190 
198.18.128.0/17 via 198.18.0.1 dev eth0 
198.19.254.254 via 198.18.0.1 dev eth0 
```

If I keep only the default routes (v4 & v6) it works, but I actually need the more precise routes as my full use case is with multus and being able to configure some extra interfaces with any routes (/1 included)
Looking at the code I fail to see why default routes are a special case: https://github.com/containernetworking/plugins/blob/9f1f9a588b1cce419007830c6301a894ff09d535/pkg/ip/utils_linux.go#L82
related review: https://github.com/containernetworking/plugins/pull/264#discussion_r263033455

champtar:
Using ptp plugin with non default routes, we get the following error when cri-o call CheckNetworkList():
```
Expected Route {Dst:{IP:198.18.128.0 Mask:ffff8000} GW:<nil>} not found in routing table
```
Using cniVersion 0.3.1 to bypass the check, we can see that the route is added with a gateway
```
$ ip r
198.18.0.0/17 via 198.18.0.1 dev eth0 src 198.18.3.102
198.18.0.1 dev eth0 scope link src 198.18.3.102
198.18.128.0/17 via 198.18.0.1 dev eth0
```

If GW is nil only check if we have a route with a DST that matches, and ignore the GW.

Fixes #886

### DIFF
--- a/pkg/ip/utils_linux.go
+++ b/pkg/ip/utils_linux.go
@@ -83,7 +83,10 @@ func ValidateExpectedRoute(resultRoutes []*types.Route) error {
 	// Ensure that each static route in prevResults is found in the routing table
 	for _, route := range resultRoutes {
 		find := &netlink.Route{Dst: &route.Dst, Gw: route.GW}
-		routeFilter := netlink.RT_FILTER_DST | netlink.RT_FILTER_GW
+		routeFilter := netlink.RT_FILTER_DST
+		if route.GW != nil {
+			routeFilter |= netlink.RT_FILTER_GW
+		}
 		var family int
 
 		switch {
--- a/plugins/main/ptp/ptp_test.go
+++ b/plugins/main/ptp/ptp_test.go
@@ -368,7 +368,7 @@ var _ = Describe("ptp Operations", func() {
 			doTest(conf, ver, 1, dnsConf, targetNS)
 		})
 
-		It(fmt.Sprintf("[%s] configures and deconfigures a dual-stack ptp link with ADD/DEL", ver), func() {
+		It(fmt.Sprintf("[%s] configures and deconfigures a dual-stack ptp link + routes with ADD/DEL", ver), func() {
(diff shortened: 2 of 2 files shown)


s1061123:
lgtm!

s1061123 (approved):

PR #902, opened 2023-06-02
spoofcheck: Make use of go-nft's ApplyConfigEcho()

SirPhuttel:
Store the relevant applied config part for later to extract the rule to delete from there instead of having to list the ruleset. This is much faster especially with large rulesets.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -14,7 +14,7 @@ require (
 	github.com/d2g/dhcp4server v0.0.0-20181031114812-7d4a0a7f59a5
 	github.com/godbus/dbus/v5 v5.1.0
 	github.com/mattn/go-shellwords v1.0.12
-	github.com/networkplumbing/go-nft v0.3.0
+	github.com/networkplumbing/go-nft v0.4.0
 	github.com/onsi/ginkgo/v2 v2.11.0
 	github.com/onsi/gomega v1.27.8
 	github.com/opencontainers/selinux v1.11.0
--- a/pkg/link/spoofcheck.go
+++ b/pkg/link/spoofcheck.go
@@ -30,7 +30,7 @@ const (
 )
 
 type NftConfigurer interface {
-	Apply(*nft.Config) error
+	Apply(*nft.Config) (*nft.Config, error)
 	Read(filterCommands ...string) (*nft.Config, error)
 }
 
@@ -39,12 +39,16 @@ type SpoofChecker struct {
 	macAddress string
 	refID      string
 	configurer NftConfigurer
+	rulestore  *nft.Config
 }
 
 type defaultNftConfigurer struct{}
 
-func (dnc defaultNftConfigurer) Apply(cfg *nft.Config) error {
-	return nft.ApplyConfig(cfg)
+func (dnc defaultNftConfigurer) Apply(cfg *nft.Config) (*nft.Config, error) {
+	const timeout = 55 * time.Second
+	ctxWithTimeout, cancelFunc := context.WithTimeout(context.Background(), timeout)
+	defer cancelFunc()
+	return nft.ApplyConfigEcho(ctxWithTimeout, cfg)
 }
 
 func (dnc defaultNftConfigurer) Read(filterCommands ...string) (*nft.Config, error) {
@@ -59,7 +63,7 @@ func NewSpoofChecker(iface, macAddress, refID string) *SpoofChecker {
 }
 
 func NewSpoofCheckerWithConfigurer(iface, macAddress, refID string, configurer NftConfigurer) *SpoofChecker {
-	return &SpoofChecker{iface, macAddress, refID, configurer}
+	return &SpoofChecker{iface, macAddress, refID, configurer, nil}
 }
 
 // Setup applies nftables configuration to restrict traffic
@@ -88,7 +92,7 @@ func (sc *SpoofChecker) Setup() error {
 	macChain := sc.macChain(ifaceChain.Name)
 	baseConfig.AddChain(macChain)
 
-	if err := sc.configurer.Apply(baseConfig); err != nil {
+	if _, err := sc.configurer.Apply(baseConfig); err != nil {
 		return fmt.Errorf("failed to setup spoof-check: %v", err)
 	}
 
@@ -102,11 +106,25 @@ func (sc *SpoofChecker) Setup() error {
 	rulesConfig.AddRule(sc.matchMacRule(macChain.Name))
 	rulesConfig.AddRule(sc.dropRule(macChain.Name))
 
-	if err := sc.configurer.Apply(rulesConfig); err != nil {
+	rulestore, err := sc.configurer.Apply(rulesConfig)
+	if err != nil {
 		return fmt.Errorf("failed to setup spoof-check: %v", err)
 	}
+	sc.rulestore = rulestore
 
 	return nil
+}
+
+func (sc *SpoofChecker) findPreroutingRule(ruleToFind *schema.Rule) ([]*schema.Rule, error) {
+	ruleset := sc.rulestore
+	if ruleset == nil {
+		chain, err := sc.configurer.Read(listChainBridgeNatPrerouting()...)
+		if err != nil {
+			return nil, err
+		}
+		ruleset = chain
+	}
+	return ruleset.LookupRule(ruleToFind), nil
 }
 
 // Teardown removes the interface and mac-address specific chains and their rules.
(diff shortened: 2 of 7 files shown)


pkg/link/spoofcheck.go
```diff
@@ -102,20 +106,31 @@ func (sc *SpoofChecker) Setup() error {
...
+
 // Teardown removes the interface and mac-address specific chains and their rules.
 // The table and base-chain are expected to survive while the base-chain rule that matches the
 // interface is removed.
 func (sc *SpoofChecker) Teardown() error {
 	ifaceChain := sc.ifaceChain()
-	currentConfig, ifaceMatchRuleErr := sc.configurer.Read(listChainBridgeNatPrerouting()...)
-	if ifaceMatchRuleErr == nil {
+	currentConfig, ifaceMatchRuleErr := sc.StoredRules()
+	if currentConfig != nil {
```
EdDev, line 133: In Go, it is common to assume the return value is undefined when an error returns.
It will be better to keep this here as well, i.e. check against the error and only then proceed to check against the value.

EdDev (commented):
Looks very nice, thank you!

I have added a few points inline.

I noticed the new logic is not covered by unit tests, which may be a bit tricky now. I think we can improve on that in a follow up.

pkg/link/spoofcheck.go
```diff
@@ -102,20 +106,31 @@ func (sc *SpoofChecker) Setup() error {
...
+	rulestore, err := sc.configurer.Apply(rulesConfig)
+	if err != nil {
 		return fmt.Errorf("failed to setup spoof-check: %v", err)
 	}
+	sc.rulestore = rulestore
 
 	return nil
 }
 
+func (sc *SpoofChecker) StoredRules() (*nft.Config, error) {
```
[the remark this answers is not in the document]
SirPhuttel: > This seems to be acting as a cache. Assuming I understand correctly, here are a few observations about this:

Correct, it caches the previously applied ruleset snippet.

> * As it returns from the store or from an actual full read, I would just name the method `Rules`/`Ruleset`. The caller does not know from where the data came from.

Fine with me! I'd go with 'appliedRules' instead, because of the following:

> * A call to this method seems to invalidate the "cache" as a side effect. Is there a reason for this behavior?

Since Teardown uses the stored rules for a delete command, once used the cache is no longer valid. I wanted to make sure it is not reused by accident. Maybe introduce a second method to invalidate the cache?

>   From a regular cache, I would expect invalidation and refresh to occur when the ruleset is updated (e.g. when `Setup` and `Teardown` is called), but not when we fetch it.

OK, so Teardown() should call sc.appliedRules() && sc.dropCache() or something.

>   Also, when reading the (actual non-cached) ruleset, the cache can be updated so it can be used later again.

That's an interesting point. My original StoredRules() is tailored to the use-case, I merely introduced it to simplify Teardown()'s body a bit. Depending on whether a cache is present or not, it will return different results: The cache will contain the flush commands and the three rule add commands captured in Setup(). If there's no cache (for whatever reason), the whole nat prerouting chain is returned instead.

Maybe I should move the rule searching logic into it, too. Then it would return
either the prerouting rule to delete.

> If the intention was different, sorry in advance.

Sorry for the unclear code. I'm entirely unfamiliar with typical coding style in Go.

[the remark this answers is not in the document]
EdDev: Sounds good to me.

I can share that in some projects, we created cache wrappers to the backed data (in this case, the `sc.configurer`). So the `Read` and `Apply` got wrapped by the cache object and the logic of when to save the data in the cache has been part of its logic. I guess it is like a cache adapter.
But I think the approach here is good enough, it does the same without wrapping the backend data source, but by operating alongside it.

SirPhuttel:
Next round: I introduced SpoofChecker::findPreroutingRule() which is a wrapper around the LookupRule() method, taking care of gathering the relevant ruleset snippet if not cached already.

Cache dropping happens in Teardown() which knows if it is no longer valid or not.

pkg/link/spoofcheck.go
```diff
@@ -39,12 +39,16 @@ type SpoofChecker struct {
...
 	configurer NftConfigurer
+	rulestore  *nft.Config
 }
 
 type defaultNftConfigurer struct{}
 
-func (dnc defaultNftConfigurer) Apply(cfg *nft.Config) error {
-	return nft.ApplyConfig(cfg)
+func (dnc defaultNftConfigurer) Apply(cfg *nft.Config) (*nft.Config, error) {
+	const timeout = 55 * time.Second
```
squeed, line 48: Why does this need a separate timeout?

SirPhuttel: Since nft.ApplyConfigEcho() requires a context, I had to create one so I just copied the respective code from Read(). Not sure if it is possible to create a context without timeout, but looking at commit 135292e0509f6 ("bridge, del: timeout after 55 secs of trying to list rules"), which introduced the 55s timeout to Read(), it seems sensible to make ApplyConfigEcho() behave the same.
Or are you suggesting to introduce a common timeout value for use in both methods?

SirPhuttel:
Two updates:

- Dropped a leftover line from the commit message, no functional changes.
- Extended spoofcheck_test.go to cover the fallback to Read() if Apply() should not return anything.

@EdDev do you think the testsuite part is sufficient now? Any idea what to test for in addition?

EdDev:
> Two updates:
> 
>     * Dropped a leftover line from the commit message, no functional changes.
> 
>     * Extended spoofcheck_test.go to cover the fallback to Read() if Apply() should not return anything.
> 
> 
> @EdDev do you think the testsuite part is sufficient now? Any idea what to test for in addition?

I am good with the current change, thank you.

I need now to release go-nft to add the new API in a formal release.
There may be another change I want to include, so I will update soon about it.

go.mod
```diff
@@ -14,7 +14,7 @@ require (
 	github.com/d2g/dhcp4server v0.0.0-20181031114812-7d4a0a7f59a5
 	github.com/godbus/dbus/v5 v5.1.0
 	github.com/mattn/go-shellwords v1.0.12
-	github.com/networkplumbing/go-nft v0.3.0
+	github.com/networkplumbing/go-nft v0.3.1-0.20230602131639-734b9ba38567
```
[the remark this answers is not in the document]
EdDev: Release is out: https://github.com/networkplumbing/go-nft/releases/tag/v0.4.0

pkg/link/spoofcheck.go
squeed: It's not a big problem, but it's certainly not necessary. A CNI plugin binary is executed for every call, so if the runtime (read: containerd / cri-o) decides to cancel the CNI execution, it will just kill the binary. That will certainly stop any nft calls :-). In that case, you can just use `context.Background()` which will never be cancelled.

That  said, 55 seconds seems like more than enough for nft to apply (I hope!), so this is probably academic.

squeed (approved):
Looks good, just two things
- a few minor lint issues
- can you pick up the latest go-nft release?

Thanks

pkg/link/spoofcheck.go
SirPhuttel: Well, frequent updates to an oversized nftables ruleset have a realistic chance of starving readers. Each time the ruleset's generation ID bumps, the reader discards the already fetched cache and restarts. Older nft versions always fetched the full ruleset irrespective of the actual command (so listing a small chain would still take long if a large other one was present).
I don't think this is possible to happen with echo mode, but I don't see why having the timeout should hurt.
Your explanation sounds like the context with timeout is pointless, though I assume @maiqueb fixed a real issue when introducing the Read() timeout.

EdDev: > A CNI plugin binary is executed for every call, so if the runtime (read: containerd / cri-o) decides to cancel the CNI execution, it will just kill the binary. That will certainly stop any nft calls :-). In that case, you can just use `context.Background()` which will never be cancelled.

We saw in k8s based systems that there is no timeout that eventually kills the call to the CNI.
We identified thousands of `nft` instances hanging with no one canceling them.

I think it is not safe to assume someone will cancel the CNI call after a timeout, it depends on the runtime or even the one who calls the runtime in the first place.

EdDev (approved):
Thank you very much!

s1061123 (commented):
/lgtm

squeed:
@SirPhuttel we merged some dep updates, can you rebase? Then we'll merge this.

SirPhuttel:
> @SirPhuttel we merged some dep updates, can you rebase? Then we'll merge this.

DONE! Applied with just a context conflict in go.mod.

SirPhuttel:
And a (final?) whitespace fix to make the linter happy.

squeed:
Hmm, the tests are failing for no clear reason. I'm taking a look.

squeed:
tests are broken; I'm investigating. Huh

squeed:
@SirPhuttel can you rebase on `main`? There was an issue with CI, sorry.

SirPhuttel:
@squeed DONE, thanks for investigating.

Issue #901, opened 2023-05-26
macvlan: support enslaving an interface by chainning


firemiles:
can we support macvlan get master from chainning just like ipvlan #88

PR #903, opened 2023-06-04
macvlan: support chaining for master interface

firemiles:
this patch support macvlan get master from previous Result link ipvlan.
 
Closes: #901


plugins/main/macvlan/macvlan.go
```diff
@@ -107,12 +107,35 @@ func loadConf(args *skel.CmdArgs, envArgs string) (*NetConf, string, error) {
...
+	if n.Master == "" {
+		if result == nil {
+			var defaultRouteInterface string
+			defaultRouteInterface, err = getNamespacedDefaultRouteInterfaceName(args.Netns, n.LinkContNs)
+			if err != nil {
+				return nil, "", err
+			}
+			n.Master = defaultRouteInterface
+		} else {
+			if len(result.Interfaces) == 1 && result.Interfaces[0].Name != "" {
```
squeed, line 133: Rather than just picking the first interface, what if we were a bit cleverer?

Right now, the macvlan plugin has the option to pick from the host namespace or container namespace. Let's preserve that when it is chained. Specifically, if `n.LinkContNS` is true, then pick the first value in `prevResult` where `sandbox == true`. Even better would be to pick the first result with routes or an IP address. Likewise, if `n.LinkContNS` is false, then do the same but for host-side interfaces.

Makes sense?

squeed: @mmirecki any thoughts here?

firemiles: > Rather than just picking the first interface, what if we were a bit cleverer?
> 
> Right now, the macvlan plugin has the option to pick from the host namespace or container namespace. Let's preserve that when it is chained. Specifically, if `n.LinkContNS` is true, then pick the first value in `prevResult` where `sandbox == true`. Even better would be to pick the first result with routes or an IP address. Likewise, if `n.LinkContNS` is false, then do the same but for host-side interfaces.
> 
> Makes sense?

We have a scenario where we need to dynamically select the host master to create macvlan, and I don't know much about the scenario where we need to configure macvlan in the container.

s1061123:
Thank you for the contribution. I agreed your concepts. 

I suppose we need to clarify how macvlan selects master interface from all combination of the options.
Currently macvlan uses 'master' and 'linkInContainer' parameters in CNI conf. Your PR introduces anther way to inject options, prevResults and it may conflicts with 'master' and 'linkInContainer' parameters.

We also need to describe somewhere in code otherwise we may get confused later.

plugins/main/macvlan/macvlan.go
```diff
@@ -107,12 +107,35 @@ func loadConf(args *skel.CmdArgs, envArgs string) (*NetConf, string, error) {
...
+				n.Master = result.Interfaces[0].Name
+			} else {
+				return nil, "", fmt.Errorf("chained master failure. PrevResult lacks a single named interface")
```
s1061123, line 136: I am thinking that we could provide a way to pick interface from prevResult by 'master' option.
I'm just wondering how about to pick interface by 'master' option name if prevResult != nil and multiple interfaces in result.

firemiles: `master` option is a static option in cni config.  how to dynamic config for different master? Give a special master option value to tell maclvan plugin to get master from prevResult ?

PR #914, opened 2023-06-21
[tuning]add ability to set tx queue len

tariq1890:
Add support to set tx queue length to network interfaces. This is useful when we have to configure net devs of SRIOV Virtual functions in the CNI layer


plugins/meta/tuning/tuning_test.go
```diff
@@ -780,14 +923,15 @@ var _ = Describe("tuning plugin", func() {
...
-		It(fmt.Sprintf("[%s] configures and deconfigures mac address, promisc mode and MTU (from conf file) with custom dataDir", ver), func() {
+		It(fmt.Sprintf("[%s] configures and deconfigures mac address, promisc mode, MTU and tx queue len (from conf file) with custom dataDir", ver), func() {
 			conf := []byte(fmt.Sprintf(`{
 				"name": "test",
 				"type": "iplink",
 				"cniVersion": "%s",
 				"mac": "c2:11:22:33:44:77",
 				"promisc": true,
 				"mtu": 4000,
+                "txQLen": 20000,
```
moshe010, line 934: fix indentation

moshe010 (changes requested):

plugins/meta/tuning/tuning_test.go
tariq1890: Done. Thanks for catching that!

plugins/meta/tuning/tuning.go
```diff
@@ -224,6 +239,9 @@ func createBackup(ifName, containerID, backupPath string, tuningConf *TuningConf
 		config.Allmulti = new(bool)
 		*config.Allmulti = (link.Attrs().RawFlags&unix.IFF_ALLMULTI != 0)
 	}
+	if tuningConf.TxQLen != 0 {
```
maiqueb, line 242: nit: this is redundant, right ?
`0` is the default value of the `configToRestore` struct; I would always set it, which would allow you to remove this if.

plugins/meta/tuning/tuning_test.go
```diff
@@ -487,6 +488,148 @@ var _ = Describe("tuning plugin", func() {
...
+				Expect(result.IPs[0].Address.String()).To(Equal("10.0.0.2/24"))
+
+				link, err := netlink.LinkByName(IFNAME)
+				Expect(err).NotTo(HaveOccurred())
+				Expect(link.Attrs().TxQLen).To(Equal(20000))
+
+				if testutils.SpecVersionHasCHECK(ver) {
+					n := &TuningConf{}
+					err = json.Unmarshal(conf, &n)
+					Expect(err).NotTo(HaveOccurred())
```
maiqueb, line 542: nit: you could inline this, which would allow to use positive logic.

```golang
Expect(json.Unmarshal(conf, &n).To(Succeed())
```

plugins/meta/tuning/tuning_test.go
```diff
@@ -487,6 +488,148 @@ var _ = Describe("tuning plugin", func() {
...
+
+					confString, err := buildOneConfig(ver, n, r)
+					Expect(err).NotTo(HaveOccurred())
+
+					args.StdinData = confString
+
+					err = testutils.CmdCheckWithArgs(args, func() error {
+						return cmdCheck(args)
+					})
+					Expect(err).NotTo(HaveOccurred())
```
maiqueb, line 552: nit: this could also be inlined.

plugins/meta/tuning/tuning_test.go
```diff
@@ -487,6 +488,148 @@ var _ = Describe("tuning plugin", func() {
...
+				})
+				Expect(err).NotTo(HaveOccurred())
+
+				result, err := types100.GetResult(r)
+				Expect(err).NotTo(HaveOccurred())
+
+				Expect(result.Interfaces).To(HaveLen(1))
+				Expect(result.Interfaces[0].Name).To(Equal(IFNAME))
+				Expect(result.IPs).To(HaveLen(1))
+				Expect(result.IPs[0].Address.String()).To(Equal("10.0.0.2/24"))
```
maiqueb, line 614: nit: if this test is about txQueueLen, why are you checking these results ? 

I am guessing this is checked in other tests; IMO, it would make sense to make the test as small and focused as it can - i.e. just check the txQueueLen.

Also applies to the previous test .

plugins/meta/tuning/tuning_test.go
```diff
@@ -487,6 +488,148 @@ var _ = Describe("tuning plugin", func() {
 			Expect(err).NotTo(HaveOccurred())
 		})
 
+		It(fmt.Sprintf("[%s] configures and deconfigures tx queue len with ADD/DEL", ver), func() {
+			conf := []byte(fmt.Sprintf(`{
+				"name": "test",
+				"type": "iplink",
+				"cniVersion": "%s",
+				"txQLen": 20000,
```
maiqueb, line 496: nit: make make this 20000 value a const, which can be re-used throughout the test ?
It is IMO better than consistently repeating this "magic" number.

maiqueb (approved):
All I have are opinionated nits. 

The code looks good, and the test coverage also looks solid.

Thank you.

plugins/meta/tuning/tuning_test.go
tariq1890: Removed

plugins/meta/tuning/tuning.go
tariq1890: I had  to use an int pointer to make sure the tx queue len was being set and unset properly

plugins/meta/tuning/tuning_test.go
tariq1890: Done

plugins/meta/tuning/tuning_test.go
tariq1890: Done

moshe010 (approved):

squeed (approved):
really nice, thanks!

PR #925, opened 2023-07-08
Update linux plugins in doc

astraw99:
To make the doc better.


plugins/linux_only.txt
```diff
@@ -1,14 +1,18 @@
 plugins/ipam/dhcp
+plugins/ipam/host-local
```
squeed, line 2: I think these plugins work on windows, so they should not be in this list.

squeed (changes requested):

plugins/linux_only.txt
astraw99: But there are plugins like `plugins/ipam/host-local` work both on linux and windows.
What about changing the file name from `linux_only.txt` to `linux.txt`, and `windows_only.txt` to `windows.txt`?
Then add the plugins into the list accordingly, it will be more clarify.

squeed: if you'd like, feel free. But right now this PR is removing test coverage it shouldn't.

astraw99: Updated with filename changed.
PTAL thanks.

plugins/linux.txt
mlguerrero12, line 1: where is this file used?

astraw99: Yes, currently the `linux.txt` file is just for doc purposes.
Updated, PTAL thanks.

mlguerrero12: but now you are removing host-local from windows which is not ok

astraw99: As host-local is both working in linux and windows, changed the filename from `linux_only.txt` to `linux.txt`, and `windows_only.txt` to `windows.txt`.
PTAL thanks.

astraw99:
@mlguerrero12 ping, PTAL thanks.

PR #928, opened 2023-07-19
noop pr to test CI

squeed:
something seems wrong.


squeed:
Aha, we're installing `cnitool@main`, which has a bug. Easy fix.

PR #935, opened 2023-07-31
nftables support for ipmasq and portmap

danwinship:
This adds nftables backends to the two remaining iptables-only bits in containernetworking/plugins; ipmasq (used by `ptp` and `bridge`) and `portmap`. For now, they are both set up to use iptables by default, unless you explicitly request nftables via the network config, or you run the plugin on a host that has nft installed but not iptables (eg, future RHEL 10 hosts).

The `firewall` plugin currently has "iptables" and "firewalld" backends. In theory it could have a separate explicit "nftables" backend, but I didn't do that here.

The ipmasq and portmap code implement nftables via https://github.com/danwinship/nftables, which is a library I started for the planned kube-proxy nftables backend, and also plan to use for various other nftables ports (eg, cri-o). It might move from `danwinship/` to somewhere more generic at some point, but it might not.

After I started working on this branch, I discovered that the `bridge` plugin was actually already using nftables via `pkg/link/spoofcheck.go`, using the https://github.com/networkplumbing/go-nft library. Using two separate nftables libraries is probably not great, so we may want to fix that eventually. The libraries have somewhat similar APIs up to a point, then diverge completely because `go-nft` uses the JSON API to `/sbin/nft` and therefore has to use the (poorly-documented) JSON rule schema, while `danwinship/nftables` uses the plain text API to `/sbin/nft` and so uses "normal" nft rules. eg, compare:

```golang
func (sc *SpoofChecker) matchIfaceJumpToChainRule(chain, toChain string) *schema.Rule {
        return &schema.Rule{
                Family: schema.FamilyBridge,
                Table:  natTableName,
                Chain:  chain,
                Expr: []schema.Statement{
                        {Match: &schema.Match{
                                Op:    schema.OperEQ,
                                Left:  schema.Expression{RowData: []byte(`{"meta":{"key":"iifname"}}`)
},
                                Right: schema.Expression{String: &sc.iface},
                        }},
                        {Verdict: schema.Verdict{Jump: &schema.ToTarget{Target: toChain}}},
                },
                Comment: ruleComment(sc.refID),
        }
}
```

which in `danwinship/nftables` syntax would become:

```golang
func (sc *SpoofChecker) matchIfaceJumpToChainRule(chain, toChain string) *nftables.Rule {
        return &nftables.Rule{
                Chain: chain,
                Rule: nftables.Concat(
                        "iifname", sc.iface,
                        "jump", toChain,
                ),
                Comment: ruleComment(sc.refID),
        }
}
```

("Draft" PR because not fully-tested yet...)


pkg/utils/netfilter.go
```diff
@@ -0,0 +1,45 @@
+// Copyright 2017 CNI authors
```
aojea, line 1: I never know how important is to have the right date on these headers

pkg/utils/netfilter.go
```diff
@@ -0,0 +1,45 @@
...
+	"github.com/coreos/go-iptables/iptables"
+	"github.com/danwinship/nftables"
+)
+
+// SupportsIPTables tests whether the system supports using netfilter via the iptables API
+// (whether via "iptables-legacy" or "iptables-nft"). (Note that this returns true if it
+// is *possible* to use iptables; it does not test whether any other components on the
+// system are *actually* using iptables.)
+func SupportsIPTables() bool {
+	ipt, err := iptables.NewWithProtocol(iptables.ProtocolIPv4)
```
aojea, line 27: is going to be important the protocol?
as if  `SupportsIP6Tables`?

pkg/ip/ipmasq_linux.go
```diff
@@ -23,7 +23,9 @@ import (
 
 // SetupIPMasq installs iptables rules to masquerade traffic
 // coming from ip of ipn and going outside of ipn
-func SetupIPMasq(ipn *net.IPNet, chain string, comment string) error {
+func SetupIPMasq(ipn *net.IPNet, pluginName, containerID string) error {
```
aojea: not a lot but seems this is used externally too https://grep.app/search?q=SetupIPMasq

pkg/utils/netfilter.go
danwinship: (So first off, we only have to worry about "`iptables` works but `ip6tables` doesn't". IMO there's no plausible scenario where the reverse is true.)

In kube-proxy we have the problem that we don't really know for sure at startup if we're going to be dealing with IPv4, IPv6, or dual-stack Services. So it's plausible that we'll try to use `ip6tables` and then discover it doesn't work, because the administrator hadn't expected us to do anything with IPv6.

The CNI plugins don't need to worry about this: they don't have "startup time"; they only run when there's a specific pod IP that they need to work with. So if `iptables` is available and `ip6tables` isn't, then either (a) it doesn't matter because we'll never see an IPv6 IP anyway, or (b) the host is broken and we _should_ return errors.

So I think it's fine as-is?

pkg/ip/ipmasq_linux.go
danwinship: ah, right. Do I have to add a new API rather than changing the old one? I don't see anything about API stability guarantees...

aojea:
linter fail, I see is still in draft,  let me know if you want a review,

danwinship:
OK, finally updated and tested:
- `ip.SetupIPMasq()` and `ip.TeardownIPMasq()` are now unchanged (and still always use iptables). Instead, I added a new `SetupIPMasqForPlugin` and `TeardownIPMasqForPlugin` that support both iptables and nftables.
- The `bridge` and `ptp` plugins now use iptables by default, or nftables if iptables isn't installed or doesn't work. Or you can force the backend via `"ipMasqBackend": "iptables"` or `"ipMasqBackend": "nftables"` in the config.
- Likewise, `portmap` now supports nftables, but still uses iptables by default, unless it's not available/usable, or you pass `"backend": "nftables"`, or you set `"conditionsV4"` or `"conditionsV6"` to something that looks like nftables syntax rather than iptables.

(I guess this will need a docs update to the cni.dev repo?)

danwinship:
(I don't seem to be able to rerun CI?)

danwinship:
oh, and I could port this to use the same nftables library that spoofcheck uses, if you preferred.

or port the spoofcheck code to use knftables like kube-proxy (and Calico and soon some other stuff)

go.mod
```diff
@@ -23,6 +23,7 @@ require (
 	github.com/safchain/ethtool v0.4.1
 	github.com/vishvananda/netlink v1.2.1-beta.2
 	golang.org/x/sys v0.21.0
+	sigs.k8s.io/knftables v0.0.16
```
aojea, line 26: you. just cut 0.0.17 two days ago https://github.com/kubernetes-sigs/knftables/releases/tag/v0.0.17 , but it does not seem required here

pkg/utils/netfilter.go
aojea: fair

go.mod
danwinship: yeah, I did mean to update that, but then I forgot, and as you say, it doesn't really matter

danwinship:
> or port the spoofcheck code to use knftables like kube-proxy

https://github.com/danwinship/plugins/commit/a0578027f6c988c8a38772ce72eb613c34730a4f

go.mod
```diff
@@ -1,13 +1,15 @@
...
+
+toolchain go1.22.0
 
 require (
 	github.com/Microsoft/hcsshim v0.12.4
 	github.com/alexflint/go-filemutex v1.3.0
 	github.com/buger/jsonparser v1.1.1
 	github.com/containernetworking/cni v1.1.2
-	github.com/coreos/go-iptables v0.7.0
+	github.com/coreos/go-iptables v0.7.1-0.20240112124308-65c67c9f46e6
```
squeed, line 12: I just tagged go-iptables v0.8.0 :-)

pkg/ip/ipmasq_nftables_linux.go
```diff
@@ -0,0 +1,171 @@
...
+// to delete the rule for container A, and not the rule for container B.)
+//
+// It would be more nftables-y to have a chain with a single rule doing a lookup against a
+// set with an element per mapping, rather than having a chain with a rule per mapping.
+// But there's no easy, non-racy way to say "delete the element 192.168.1.3 from the set,
+// but only if it was added for container A, not if it was added for container B".
+
+// hashForContainer returns a unique hash identifying the rules for this container with
+// this plugin
+func hashForContainer(pluginName, containerID string) string {
```
squeed, line 46: There is an error in the iptables implementation of this -- this needs to hash (network, ifname, containerid).

pkg/ip/ipmasq_nftables_linux.go
```diff
@@ -0,0 +1,171 @@
...
+func setupIPMasqNFTables(ipn *net.IPNet, pluginName, containerID string) error {
+	nft, err := knftables.New(knftables.InetFamily, ipMasqTableName)
+	if err != nil {
+		return err
+	}
+	return setupIPMasqNFTablesWithInterface(nft, ipn, pluginName, containerID)
+}
+
+func setupIPMasqNFTablesWithInterface(nft knftables.Interface, ipn *net.IPNet, pluginName, containerID string) error {
+	comment := hashForContainer(pluginName, containerID)
```
squeed, line 60: The iptables implementation is wrong here -- we need to be able to find all masquerade rules for a given network. Can the comment include network name in plain text?

squeed:
Looks basically good. I'd like to see GC support if possible, since it informs the design the API. Could you add that, even if the top-end API glue isn't there yet.

pkg/ip/ipmasq_nftables_linux.go
danwinship: `pluginName` here is `NetConf.Name`, which is what you mean by `network`, right? (Yeah, `pluginName` is a bad variable name... I was assuming it meant `NetConf.Type`...)

Why do we need `ifname`?

plugins/meta/portmap/main.go
```diff
@@ -279,3 +302,58 @@ func parseConfig(stdin []byte, ifName string) (*PortMapConf, *current.Result, er
...
+
+	// If backend wasn't requested explicitly, default to iptables, unless it is not
+	// available (and nftables is). FIXME: flip this default at some point.
+	if conf.Backend == nil {
+		if !utils.SupportsIPTables() && utils.SupportsNFTables() {
+			conf.Backend = &nftablesBackend
+		} else {
+			conf.Backend = &iptablesBackend
+		}
+	}
```
aojea, line 328: maybe is better to not have validation and defaulting in the same place, despite you already  indicate that in the comment? 

maybe just move this into a `defaultBackend(conf *PortMapConf)` or just name this function to make clear it also defaults values?

plugins/meta/portmap/portmap_test.go
```diff
@@ -139,7 +215,7 @@ var _ = Describe("portmapping configuration", func() {
 							{ "hostPort": 8080, "containerPort": 80, "protocol": "tcp"}
 						]
 					},
-					"conditionsV4": ["a", "b"],
+					"conditionsV4": ["-a", "b"],
```
aojea, line 218: what does the `-` minus symbol means?

plugins/meta/portmap/main.go
danwinship: Yeah, I could rename it. I think it's useful to do validation and defaulting together, because it broadens the set of cases where we'll pick the right backend without the user needing to specify anything.

danwinship:
@squeed this is ready for re-review btw; it includes an implementation of GC for nftables ipmasq, which isn't actually plumbed through, but at least gets unit tested.

squeed (approved):

champtar:
I've started to play with this, found a regression (#1114 / PR #1117), and 2 bugs ( #1115 - PR #1116 / #1118 - PR todo)

PR #936, opened 2023-08-04
bridge: add support for setting vlans as pvid and untagged

bersoare:
This PR adds support for specifying a PVID and untagged vlans for a trunk. Added new checks for the presence of these attributes in the tests, as well as a new test case that uses the newly introduced values.

### DIFF
--- a/plugins/main/bridge/bridge.go
+++ b/plugins/main/bridge/bridge.go
@@ -56,6 +56,8 @@ type NetConf struct {
 	HairpinMode         bool         `json:"hairpinMode"`
 	PromiscMode         bool         `json:"promiscMode"`
 	Vlan                int          `json:"vlan"`
+	PVID                int          `json:"pvid,omitempty"`
+	UntaggedIDs         []*VlanTrunk `json:"untaggedIDs,omitempty"`
 	VlanTrunk           []*VlanTrunk `json:"vlanTrunk,omitempty"`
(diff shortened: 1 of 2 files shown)


plugins/main/bridge/bridge.go
```diff
@@ -391,7 +403,7 @@ func ensureVlanInterface(br *netlink.Bridge, vlanID int, preserveDefaultVlan boo
 			return nil, fmt.Errorf("faild to find host namespace: %v", err)
 		}
 
-		_, brGatewayIface, err := setupVeth(hostNS, br, name, br.MTU, false, vlanID, nil, preserveDefaultVlan, "")
+		_, brGatewayIface, err := setupVeth(hostNS, br, name, br.MTU, false, vlanID, trunkConfig{allowedVlans: nil}, preserveDefaultVlan, "")
```
s1061123, line 406: `trunkConfig{allowedVlans: nil}` can be changed as `trunkConfig{}`

```
$ cat test.go 
package main

import "fmt"
type trunkConfig struct {
	pvid         int
	untaggedVids map[int]struct{}
	allowedVlans []int
}

func main() {
	fmt.Printf("%v\n", trunkConfig{}.allowedVlans)
}
$ go run test.go 
[]

```

bersoare:
hello @s1061123 ,

thanks for the suggestions, indeed changing it to be a []VlanTrunk makes sense to keep it consistent.
just changed it, all tests passed locally. let me know what you think.

thank you!

Issue #953, opened 2023-10-10
macvlan: error killing pod: Link not found


cyclinder:
```
Events:
  Type     Reason          Age                   From               Message
  ----     ------          ----                  ----               -------
  Normal   Scheduled       5m8s                  default-scheduler  Successfully assigned ns4740-192363920/ds-4bfa5c3fb5-8pwcf to spiderpool1008202421-control-plane
  Normal   AddedInterface  5m8s                  multus             Add eth0 [fc00:f853:ccd:e793:f::c7/64 172.18.40.6/16] from kube-system/macvlan-vlan0
  Normal   AddedInterface  5m6s                  multus             Add net1 [fd00:95b2::3/120 10.185.192.3/24] from ns4740-192363920/test-multus-6d8e8be07a
  Normal   Pulled          5m5s                  kubelet            Container image "alpine" already present on machine
  Normal   Created         5m5s                  kubelet            Created container samplepod
  Normal   Started         5m5s                  kubelet            Started container samplepod
  Normal   Killing         5m3s                  kubelet            Stopping container samplepod
  Warning  FailedKillPod   32s (x21 over 4m33s)  kubelet            error killing pod: failed to "KillPodSandbox" for "5c8fe27b-afa4-4ff2-9399-93773a4e1f5e" with KillPodSandboxError: "rpc error: code = Unknown desc = failed to destroy network for sandbox \"402c1fae47b140592e12cd11f3f5f1927788e6196e6f8d313ab09eee9cd4fedc\": delegateDel: error invoking ConflistDel - \"test-multus-6d8e8be07a\": conflistDel: error in getting result from DelNetworkList: Link not found"
```

I checked the code, this may have happened in https://github.com/containernetworking/plugins/blob/f95505231a8209b1b41eac2d5c2399c6ab444785/plugins/main/macvlan/macvlan.go#L407

Link not found != link not found?

cyclinder:
```
root@cyclinder3:~/cyclinder/test/linknofound# cat main.go
package main

import (

	"fmt"
	"log"

	"github.com/containernetworking/plugins/pkg/ip"

)

func main() {

	if err := ip.DelLinkByName("eth0"); err != nil {
		fmt.Println("del err: ",err)
		if err != ip.ErrLinkNotFound {
			log.Fatalln(err)
		}
	}

	fmt.Println("Del")
}
root@cyclinder3:~/cyclinder/test/linknofound# go run main.go
del err:  link not found
Del
```

[...]

PR #954, opened 2023-10-11
macvlan cmdDel: replace the loadConf function with json.unmarshal

cyclinder:
When the master interface on the node has been deleted, and loadConf tries 
to get the MTU, This causes cmdDel to return a linkNotFound error to the
runtime. The cmdDel only needs to unmarshal the net config. No need to
get the MTU. So we just replaced the loadConf function with
json.unmarshal in cmdDel.

Fixes https://github.com/containernetworking/plugins/issues/953

### DIFF
--- a/plugins/main/macvlan/macvlan.go
+++ b/plugins/main/macvlan/macvlan.go
@@ -382,13 +382,13 @@ func cmdAdd(args *skel.CmdArgs) error {
 }
 
 func cmdDel(args *skel.CmdArgs) error {
-	n, _, err := loadConf(args, args.Args)
+	var n NetConf
+	err := json.Unmarshal(args.StdinData, &n)
 	if err != nil {
-		return err
+		return fmt.Errorf("failed to load netConf: %v", err)
 	}
 
 	isLayer3 := n.IPAM.Type != ""
-
 	if isLayer3 {
 		err = ipam.ExecDel(n.IPAM.Type, args.StdinData)
 		if err != nil {


plugins/main/macvlan/macvlan.go
```diff
@@ -382,13 +382,13 @@ func cmdAdd(args *skel.CmdArgs) error {
 }
 
 func cmdDel(args *skel.CmdArgs) error {
-	n, _, err := loadConf(args, args.Args)
-	if err != nil {
-		return err
+	var n NetConf
```
mlguerrero12, line 385: I guest you needed the external err variable because the rest of the code is assuming there is one already declared, right?

You could simply have 
err := json.Unmarshal(args.StdinData, &n)

and then check the error variable. err will be created and used by the rest of the code.

plugins/main/macvlan/macvlan.go
```diff
@@ -382,13 +382,13 @@ func cmdAdd(args *skel.CmdArgs) error {
...
+	var err error
+	if err := json.Unmarshal(args.StdinData, &n); err != nil {
+		return fmt.Errorf("failed to unmarshal net conf: %v", err)
```
mlguerrero12, line 388: I think "failed to load netconf" was better and is consistent with other parts.

plugins/main/macvlan/macvlan.go
cyclinder: Actually, I want to write in this way, We just add one line:

```
+       var err error
	if err = json.Unmarshal(args.StdinData, &n); err != nil {
		return fmt.Errorf("failed to unmarshal net conf: %v", err)
        }
       ....
```

If write in your way:

```
    	err := json.Unmarshal(args.StdinData, &n)
+       if err != nil {
		return fmt.Errorf("failed to unmarshal net conf: %v", err)
        }
```

Both are ok for me.

mlguerrero12:
looks good now, please improve the commit title and message. See https://cbea.ms/git-commit/ as reference.

cyclinder:
thanks @mlguerrero12, This is my first time contributing to CNI, Thank you for helping me.

mlguerrero12:
In the description, you mention that the problem is when the pod's nic is deleted, but in reality the bug you are seeing is because the master interface is deleted and loadConf tries to get the MTU. The delete command only needs to unmarshall the net config. No need to get the MTU. Please update the description and wrap the body of the commit around 72 characters as mentioned in bullet 6 of the link I shared above.

cyclinder:
thanks a lot @mlguerrero12. How about the latest description?

mlguerrero12:
LGTM

Just a small nit: please correct the grammar mistakes in the body of the commit.

@squeed, could you please have a look at this?

squeed (approved):

cyclinder:
Updated, Thanks @mlguerrero12 @squeed

mlguerrero12 (approved):

cyclinder:
Hey @mlguerrero12 @squeed, Is time to merge this?

squeed (approved):

PR #961, opened 2023-10-23
dependabot: batch updates

squeed:
Rather than endless rebases, just batch all go updates to once a week.


austinvazquez (approved):
Neat! I need to steal this for my projects.

PR #962, opened 2023-10-23
bridge: remove useless code

arthur-zhang:
gws.defaultRouteFound is always false.
firstV4Addr it is declared but not used anywhere in the code.

### DIFF
--- a/plugins/main/bridge/bridge.go
+++ b/plugins/main/bridge/bridge.go
@@ -729,7 +729,7 @@ func cmdAdd(args *skel.CmdArgs) error {
 	}
 
 	// Use incoming DNS settings if provided, otherwise use the
-	// settings that were already configued by the IPAM plugin
+	// settings that were already configured by the IPAM plugin
 	if dnsConfSet(n.DNS) {
 		result.DNS = n.DNS
 	}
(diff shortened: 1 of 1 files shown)


arthur-zhang:
help wanted, seems ci is broken.

maiqueb:
> help wanted, seems ci is broken.

The issue you're complaining about was fixed in https://github.com/containernetworking/plugins/pull/967

Please rebase latest main branch, it should be OK then.

maiqueb (approved):
Good catch.

Thanks for cleaning up the code base.
You'll need to rebase the latest main branch; that should take care of the lint errors causing CI to fail.

maiqueb (approved):

squeed (approved):

plugins/main/bridge/bridge.go
```diff
@@ -240,7 +240,7 @@ func calcGateways(result *current.Result, n *NetConf) (*gwInfo, *gwInfo, error)
 
 		// Add a default route for this family using the current
 		// gateway address if necessary.
-		if n.IsDefaultGW && !gws.defaultRouteFound {
```
dcbw, line 243: Are we sure this is NOP? At this point `gws` refers to a global defined at the top of the function:

```
	gwsV4 := &gwInfo{}
	gwsV6 := &gwInfo{}
```

So if there are multiple v4/v6 addresses, I think defaultRouteFound would carry across the IP addressing loop. I think we need to keep this?

s1061123:
@dcbw I understood that. That is right that gws is used in `calcGateways()`. Let me double check that in #980

Issue #985, opened 2023-11-27
feature: allow configuration of num_queues for macvlan and ipvlan


cyclinder:
allow configuration of num_queues for macvlan and ipvlan, it may enhance network performance.

```
	linkAttrs := netlink.LinkAttrs{
		MTU:         conf.MTU,
		Name:        tmpName,
		ParentIndex: m.Attrs().Index,
		Namespace:   netlink.NsFd(int(netns.Fd())),
		NumTxQueues: 20,   // numtxqueue
		NumRxQueues: 20,  // numrxqueue
	}
```

PR #986, opened 2023-11-27
allow configuring numTxQueues and numRxQueues for macvlan and ipvlan

cyclinder:
Adding a new configuration option: 'queues,' which is used to configure the "numtxqueues" and "numrxqueues" properties of the network card. It can only be a non-negative integer or 0. The default is 0, indicating that netlink API will configure the default numqueues as 1. In scenarios where network bandwidth sensitivity is crucial, allowing the configuration of numqueues will enhance network bandwidth.

Fixes https://github.com/containernetworking/plugins/issues/985

related-link: https://github.com/projectcalico/cni-plugin/pull/1116

### DIFF
--- a/plugins/main/ipvlan/ipvlan.go
+++ b/plugins/main/ipvlan/ipvlan.go
@@ -38,6 +38,7 @@ type NetConf struct {
 	Master     string `json:"master"`
 	Mode       string `json:"mode"`
 	MTU        int    `json:"mtu"`
+	Queues     int    `json:"queues,omitempty"`
 	LinkContNs bool   `json:"linkInContainer,omitempty"`
 }
 
@@ -87,6 +88,10 @@ func loadConf(args *skel.CmdArgs, cmdCheck bool) (*NetConf, string, error) {
 			}
 		}
 	}
+	if n.Queues < 0 {
+		return nil, "", fmt.Errorf("invalid Queues %d, must be greater than 0", n.Queues)
+	}
+
 	return n, n.CNIVersion, nil
 }
 
@@ -149,6 +154,8 @@ func createIpvlan(conf *NetConf, ifName string, netns ns.NetNS) (*current.Interf
 			MTU:         conf.MTU,
 			Name:        tmpName,
 			ParentIndex: m.Attrs().Index,
+			NumTxQueues: conf.Queues,
+			NumRxQueues: conf.Queues,
 			Namespace:   netlink.NsFd(int(netns.Fd())),
 		},
 		Mode: mode,
--- a/plugins/main/ipvlan/ipvlan_test.go
+++ b/plugins/main/ipvlan/ipvlan_test.go
@@ -343,6 +343,7 @@ var _ = Describe("ipvlan Operations", func() {
 					Master:     masterInterface,
 					Mode:       "l2",
 					MTU:        1500,
+					Queues:     20,
 					LinkContNs: isInContainer,
 				}
 
--- a/plugins/main/macvlan/macvlan.go
+++ b/plugins/main/macvlan/macvlan.go
@@ -39,6 +39,7 @@ type NetConf struct {
 	Master     string `json:"master"`
 	Mode       string `json:"mode"`
 	MTU        int    `json:"mtu"`
+	Queues     int    `json:"queues,omitempty"`
 	Mac        string `json:"mac,omitempty"`
 	LinkContNs bool   `json:"linkInContainer,omitempty"`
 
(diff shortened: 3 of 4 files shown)


s1061123:
- Tx/Rx queue parameter should be different params because user want to configure separately (for example, macvlan/ipvlan does not have special tx queue len, hence user may want to config Tx for 0).
- Could you please add unit test to verify that tx/rx queue parameters are correctly configured to the created macvlan/ipvlan? (as https://github.com/cyclinder/plugins/blob/2a3835bd2cc8bf3a9e0d72f32b46814de877942a/plugins/main/macvlan/macvlan_test.go#L766-L768 does for MAC)

s1061123 (commented):
- Tx/Rx queue parameter should be different params because user want to configure separately (for example, macvlan/ipvlan does not have special tx queue len, hence user may want to config Tx for 0).
- Could you please add unit test to verify that tx/rx queue parameters are correctly configured to the created macvlan/ipvlan? (as https://github.com/cyclinder/plugins/blob/2a3835bd2cc8bf3a9e0d72f32b46814de877942a/plugins/main/macvlan/macvlan_test.go#L766-L768 does for MAC)

cyclinder:
thanks @s1061123 

If configuring tx/rx separately, it may increase complexity. In most scenarios, the queue lengths for tx/rx are the same(please correct me if I'm wrong), but I consider it okay whether configured separately or not.

> Could you please add unit test to verify that tx/rx queue parameters are correctly configured to the created macvlan/ipvlan? 

yes, I'm willing to do this.

s1061123:
> If configuring tx/rx separately, it may increase complexity. In most scenarios(refer to https://github.com/projectcalico/cni-plugin/pull/1116), the queue lengths for tx/rx are the same(please correct me if I'm wrong), but I consider it okay whether configured separately or not.

Please show why you conclude 'most scenarios the queue lengths for tx/rx are the same ' from above PR, especially from the description, 'vpp has more than one workers to process packet in our situation'. You may need to know that vpp is not linux kernel network stack. In addition, linux kernel and netlink provide both as separate parameter. This is your PR, so please explain why you decide tx/rx queue are the same in almost case to consistent to linux kernel desgin, otherwise your change is not reasonable.

In addition, you may need to show who much performance is improved by multi rx/tx queue in ipvlan/macvlan because ipvlan/macvlan is software forwarding, hence the performance improvement is not the same as vpp/hardware NIC.

In #985, you mention 'it may enhance network performance.'. Could you please show 'tx/rx queue enahnce improves network performance in macvlan/ipvlan'? I'm not clear why it improves because macvlan/ipvlan is linux kernel network stack feature and it may not have multiple worker. So you need to explain why enhance macvlan/ipvlan tx/rx queue improves network performance or need to show 'how much performance is improved by tx/rx queue enhancement.

cyclinder:
> We are talking about num of queue, not queue length.

Sorry for my bad words.

> Please show why you conclude 'most scenarios the queue lengths for tx/rx are the same ' from above PR.

From the user's perspective, they may not be aware of how the kernel is designed. Typically, when using it, queues of tx/rx are configured together. Setting only one of them may be confusing🤔.

> So you need to explain why enhance macvlan/ipvlan tx/rx queue improves network performance or need to show 'how much performance is improved by tx/rx queue enhancement.

Alright, I will make some performance benchmarks based on the current changes to see if it improves performance.

s1061123:
Based on above conversation, I suspect you just files PR without any clear reason. Like "I found interesting parameter (numTxQueues and numRxQueues). I don't why, but these parameters look like increasing more performance, so let's file a PR. Of coruse, I don't know  recommendation parameters too because I don't know how linux kernel use these parameters." Simply to say, it is not a good PR because PR submitter can not explain what your code changes. PR submitter should explain how your code changes and what is happen actually because you actually write a code. Reviewer does NOT have responsible to your code. Reviewer is just responsible to review your code and reviewer does not explain what your code does. YOU, are the only one who explain because YOU write a code you should have a reason why you change and you must understand what you changes. Please be responsible to your code and please explain what the change makes.

> From the user's perspective, they may not be aware of how the kernel is designed. Typically, when using it, queues of tx/rx are configured together. Setting only one of them may be confusing🤔.

This comment seems to be improvement feedback to linux kernel. Because numTxQueue and numRxQueue is defined linux kernel and its netlink API and you conclude that both should be consolidated to one parameter (numQueues?). This shoud be done in linux kernel and netlink API, not CNI because these are defined in linux kernel. You believe that almost user want to configure numTxQueue/numRxQeueue as same number without any reason (you suppose they don't have a reason why, but user want to change these parameters, right?). So don't ask CNI, and please ask Linux kernel community (https://lkml.org/). numTxQueue/numRxQueue are defined in linux kernel. 

At least, I believe that if someone want to change some of linux kernel parameter, they understand what the parameter is (and they also understand it should be increased or decreased).

s1061123:
I'd recommend you to close this PR and issue #985 once and file new PR again when you're ready.
Closing PR and closing issue is not penalty to you also us ;)

PR #989, opened 2023-12-08
Support chaining of VLAN plugin

nvetel:
# Introduction

This feature adds the ability to create multiple vlan interfaces out of a container namespace interface created by another CNI.
It is greatly inspired from [PR #903](https://github.com/containernetworking/plugins/pull/903) proposing a similar feature on macvlan.

# Problem

I need to have an interface in a pod, and create as many vlan interfaces as I want on that master interface, with a different IPAM configuration on each one.

image

# Proposal

The idea is to relay on the vlan plugin to create such configuration, since creating a vlan interface and using IPAM on is exactly what it is for. 

An obvious solution is to chain vlan plugin after the plugin creating master interface, and to use the linkInContainer option to target a master interface in the container namespace.
But the limitation is that it's hard to know the master interface in advance to reuse it in the chained vlan plugin.

The proposal is then use an empty master in the vlan plugin netconf combined with the linkInContainer parameter to ask the plugin to use the CNI_IF_NAME argument as master interface, and to name the new interface as the master with a dot vlanID notation at the end.

Here is a netconf example with a master interface created by the "my-main-plugin", and then the VLAN interfaces 3 and 4 created out of it :

```
apiVersion: "k8s.cni.cncf.io/v1"
kind: NetworkAttachmentDefinition
metadata:
  name: vlan-chaining-example
spec:
  config: '{
        "cniVersion": "0.3.1",
        "name": "vlan-chaining-example",
        "plugins": [{
            "type": "my-main-plugin",
        },{
            "type": "vlan",
            "linkInContainer": true,
            "vlanId": 3,
            "ipam": {
                    "type": "static",
                    "addresses": [{
                          "address": "192.168.3.10",
                          "gateway": "192.168.3.1"
                    }]
            }
        },{
            "type": "vlan",
            "linkInContainer": true,
            "vlanId": 4,
            "ipam": {
                    "type": "static",
                    "addresses": [{
                          "address": "192.168.4.10",
                          "gateway": "192.168.4.1"
                    }]
            }
        }]
    }'
```

### DIFF
--- a/plugins/main/vlan/vlan.go
+++ b/plugins/main/vlan/vlan.go
@@ -52,9 +52,31 @@ func loadConf(args *skel.CmdArgs) (*NetConf, string, error) {
 	if err := json.Unmarshal(args.StdinData, n); err != nil {
 		return nil, "", fmt.Errorf("failed to load netconf: %v", err)
 	}
-	if n.Master == "" {
-		return nil, "", fmt.Errorf("\"master\" field is required. It specifies the host interface name to create the VLAN for")
+
+	// Parse previous result
+	var result *current.Result
+	var err error
+	if n.NetConf.RawPrevResult != nil {
+		if err = version.ParsePrevResult(&n.NetConf); err != nil {
+			return nil, "", fmt.Errorf("could not parse prevResult: %v", err)
+		}
+
+		result, err = current.NewResultFromResult(n.PrevResult)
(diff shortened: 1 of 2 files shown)


nvetel:
PR duplicated with  https://github.com/containernetworking/plugins/pull/993

Issue #951, opened 2023-10-02
Add "activateInterface" option to bridge plugin CNI


AlonaKaplan:
Add "activateInterface" option to bridge plugin CNI.
The option will control whether the defined interface will be brought `up` by the plugin. The default will be true.
This is required in case some other CNI in the plugin chain wants to modify the interface before it is brought up and made visible to the network.

For example, in case I want to prevent my interface to send any IPv6 traffic. I can use a another CNI to achieve this by setting `net.ipv6.conf.all.disable_ipv6` to `1`.
But since the bridge CNI will first activate the interface, it may send IPv6 data (NDP) on activation, before the next CNI disables it.
Having `activateInterface=false` will solve this issue. The bridge CNI will create and configure the interface and the next CNI in the chain will apply the `sysctl` and activate the interface,

maiqueb:
Without going into if the ask makes sense or not, I'd say the flag should be the opposite - i.e. by default have the interface enabled, thus having the existing current behavior as "the" default.

AlonaKaplan:
> Without going into if the ask makes sense or not, I'd say the flag should be the opposite - i.e. by default have the interface enabled, thus having the existing current behavior as "the" default.

The ticket says `the default will be true`.

maiqueb:
> > Without going into if the ask makes sense or not, I'd say the flag should be the opposite - i.e. by default have the interface enabled, thus having the existing current behavior as "the" default.
> 
> The ticket says `the default will be true`.

Oh right ! 

FWIW, you don't say what's the name of the attribute, and thus: "The option will control whether the defined interface will be brought up by the plugin. The default will be true." can pretty much mean anything - i.e. a `disableInterface` knob controls whether the defined interface will be brought up by the plugin, and if it defaults to true, you'd get a disabled interface.

I'm happy we share the default should be the "current behavior", but I still think the issue could be a bit more explicit; Near the end of the description you do propose a name for the knob; could you maybe introduce it it the beginning ? IMO that's more explicit / less confusing.

PR #997, opened 2024-01-09
bridge: Enable disabling bridge interface

ormergi:
The new `disableContainerInterface` parameter is added to the bridge plugin to
enable setting the container interface state down.

When the parameter is enabled, the container interface (veth peer that is placed
at the container ns) remain down (i.e: disabled).
The bridge and host peer interfaces state are not affected by the parameter.

Notes for reviewers:
- Since IPAM logic involve various configurations including waiting for addresses
to be realized and setting the interface state UP, the new parameter cannot work
with IPAM.
In case both IPAM and DisableContainerInterface parameters are set, the bridge
plugin will raise an error.
- CNI versions 0.1.0 and 0.2.0 result expect single IP addresses, where newer versions expect multiple addresses (slice).
  When the new parameter is set, there will be no IP to report, thus it cannot work with older versions.
  In newer versions it works and an empty IP addresses slice will return.

Related Issue: https://github.com/containernetworking/plugins/issues/951

### DIFF
--- a/plugins/main/bridge/bridge.go
+++ b/plugins/main/bridge/bridge.go
@@ -47,19 +47,20 @@ const defaultBrName = "cni0"
 
 type NetConf struct {
 	types.NetConf
-	BrName              string       `json:"bridge"`
-	IsGW                bool         `json:"isGateway"`
-	IsDefaultGW         bool         `json:"isDefaultGateway"`
-	ForceAddress        bool         `json:"forceAddress"`
-	IPMasq              bool         `json:"ipMasq"`
-	MTU                 int          `json:"mtu"`
-	HairpinMode         bool         `json:"hairpinMode"`
-	PromiscMode         bool         `json:"promiscMode"`
-	Vlan                int          `json:"vlan"`
-	VlanTrunk           []*VlanTrunk `json:"vlanTrunk,omitempty"`
-	PreserveDefaultVlan bool         `json:"preserveDefaultVlan"`
-	MacSpoofChk         bool         `json:"macspoofchk,omitempty"`
-	EnableDad           bool         `json:"enabledad,omitempty"`
+	BrName                    string       `json:"bridge"`
+	IsGW                      bool         `json:"isGateway"`
+	IsDefaultGW               bool         `json:"isDefaultGateway"`
+	ForceAddress              bool         `json:"forceAddress"`
+	IPMasq                    bool         `json:"ipMasq"`
+	MTU                       int          `json:"mtu"`
+	HairpinMode               bool         `json:"hairpinMode"`
+	PromiscMode               bool         `json:"promiscMode"`
+	Vlan                      int          `json:"vlan"`
+	VlanTrunk                 []*VlanTrunk `json:"vlanTrunk,omitempty"`
+	PreserveDefaultVlan       bool         `json:"preserveDefaultVlan"`
+	MacSpoofChk               bool         `json:"macspoofchk,omitempty"`
+	EnableDad                 bool         `json:"enabledad,omitempty"`
+	DisableContainerInterface bool         `json:"disableContainerInterface,omitempty"`
 
 	Args struct {
 		Cni BridgeArgs `json:"cni,omitempty"`
@@ -530,6 +531,10 @@ func cmdAdd(args *skel.CmdArgs) error {
 
 	isLayer3 := n.IPAM.Type != ""
 
+	if isLayer3 && n.DisableContainerInterface {
+		return fmt.Errorf("cannot use IPAM when DisableContainerInterface flag is set")
+	}
+
 	if n.IsDefaultGW {
 		n.IsGW = true
 	}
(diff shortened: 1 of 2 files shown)


plugins/main/bridge/bridge.go
```diff
@@ -708,7 +715,9 @@ func cmdAdd(args *skel.CmdArgs) error {
 		}
 
 		if idx == len(retries)-1 {
-			return fmt.Errorf("bridge port in error state: %s", hostVeth.Attrs().OperState)
+			if !n.DisableContainerInterface {
```
EdDev, line 718: It does not look to be relevant here.
The "check bridge port state" needs to be bypass if the flag is set to true.

EdDev (commented):
Thanks!

plugins/main/bridge/bridge.go
```diff
@@ -530,6 +531,10 @@ func cmdAdd(args *skel.CmdArgs) error {
 
 	isLayer3 := n.IPAM.Type != ""
 
+	if isLayer3 && n.DisableContainerInterface {
+		return fmt.Errorf("cannot use IPAM when container interface should be disabled")
```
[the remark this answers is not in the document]
ormergi: Done

plugins/main/bridge/bridge.go
ormergi: Done

ormergi:
[Changes](https://github.com/containernetworking/plugins/compare/034585e9352b00514e771ac69b0ab7b426720764..7b9327957804b3fd2523027771054e94d8bdc39b): fix linter issues

plugins/main/bridge/bridge.go
```diff
@@ -530,6 +531,10 @@ func cmdAdd(args *skel.CmdArgs) error {
 
 	isLayer3 := n.IPAM.Type != ""
 
+	if isLayer3 && n.DisableContainerInterface {
+		return fmt.Errorf("cannot use IPAM when DisableContainerInterface flag is set")
```
EdDev, line 535: Can you check this scenario in the tests?

ormergi: Done

ormergi:
[Changes](https://github.com/containernetworking/plugins/compare/a12ba15b964a342e8d9537a6e82f8197585ea9f9..7e131a0076de6bc69ed0ebc54c5903604ed940fa): failing test when both IPAM and disable-interface parameters are set

plugins/main/bridge/bridge_test.go
```diff
@@ -2461,6 +2490,36 @@ var _ = Describe("bridge Operations", func() {
...
+
+		It(fmt.Sprintf("[%s] should set the container veth peer state down", ver), func() {
+			Expect(originalNS.Do(func(ns.NetNS) error {
+				defer GinkgoRecover()
+				tc := testCase{
+					cniVersion:       ver,
+					disableContIface: true,
+					isLayer2:         true,
+					AddErr020:        "cannot convert: no valid IP addresses",
+					AddErr010:        "cannot convert: no valid IP addresses",
```
AlonaKaplan, line 2517: Why are those errors expected?

ormergi: CNI versions 0.1.0 and 0.2.0 result expect single IP addresses, where newer versions expect multiple addresses (slice).
When the new parameter is set, there will be no IP to report, thus it cannot work with older versions.
In newer versions it works and an empty IP addresses slice will return.

AlonaKaplan: Can you please add this explanation to the PR message?

ormergi: Done

mlguerrero12 (commented):
have you checked if there is something to do for the check command? Does it verify that the link is up?

plugins/main/bridge/bridge.go
```diff
@@ -692,23 +698,28 @@ func cmdAdd(args *skel.CmdArgs) error {
...
+				break
+			}
 
-		if idx == len(retries)-1 {
-			return fmt.Errorf("bridge port in error state: %s", hostVeth.Attrs().OperState)
+			if idx == len(retries)-1 {
+				return fmt.Errorf("bridge port in error state: %s", hostVeth.Attrs().OperState)
+			}
 		}
 	}
```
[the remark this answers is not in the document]
ormergi: I have posted a PR for small refactoring around the places this PR change https://github.com/containernetworking/plugins/pull/998
Can we please address what you suggest on a follow up PR?

ormergi:
> have you checked if there is something to do for the check command? Does it verify that the link is up?

Hi @mlguerrero12 thanks for the review, from what I saw the check cmd does not check the interface state at all.

plugins/main/bridge/bridge.go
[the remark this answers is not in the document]
mlguerrero12: sure

mlguerrero12 (commented):
LGTM, thanks @ormergi

ormergi:
Hey @squeed please take another look, let me know if there is anything else we should do :slightly_smiling_face:

ormergi:
Updating docs https://github.com/containernetworking/cni.dev/pull/137

squeed (approved):

squeed:
Thanks for your contribution!

PR #998, opened 2024-01-14
bridge: Extract some netlink operations code to file

ormergi:
This is a small refactoring around netlink operations logic.
Setting link state up and waiting for link oper state code is moved to link package.


pkg/link/opstate.go
```diff
@@ -0,0 +1,55 @@
...
+			return nil, fmt.Errorf("timeout waiting for %q state up", linkName)
+		}
+	}
+	return link, nil
+}
+
+func SetUp(linkName string) error {
+	link, err := netlink.LinkByName(linkName)
+	if err != nil {
+		return fmt.Errorf("failed to retrieve link: %v", err)
```
squeed, line 49: nit: please use `%w` to wrap errors everywhere.

squeed (changes requested):
A few small nits, then this looks good!

pkg/link/opstate.go
ormergi: I didn't saw error being unwrapped in the bridge plugin code, also we didnt warped them before this change :slightly_smiling_face: 
Why is it necessary to wrap them here?

pkg/link/opstate.go
```diff
@@ -0,0 +1,55 @@
...
+		link, err = netlink.LinkByName(linkName)
+		if err != nil {
+			return nil, err
+		}
+		if link.Attrs().OperState == netlink.OperUp {
+			break
+		}
+
+		if idx == len(retries)-1 {
+			return nil, fmt.Errorf("timeout waiting for %q state up", linkName)
```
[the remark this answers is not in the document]
ormergi: Done

ormergi:
@squeed thanks for the review :slightly_smiling_face: , please take another look.

pkg/link/opstate.go
```diff
@@ -0,0 +1,56 @@
...
+func WaitForOperStateUp(linkName string) (netlink.Link, error) {
+	var link netlink.Link
+	var err error
+	retries := []int{0, 50, 500, 1000, 1000}
+	for idx, sleep := range retries {
+		time.Sleep(time.Duration(sleep) * time.Millisecond)
+
+		link, err = netlink.LinkByName(linkName)
+		if err != nil {
+			return nil, err
```
mlguerrero12, line 33: let's be consistent and return here fmt.Errorf("failed to retrieve link: %v", err) as in SetUp

plugins/main/bridge/bridge.go
```diff
@@ -677,39 +676,18 @@ func cmdAdd(args *skel.CmdArgs) error {
 			}
 		}
 	} else {
+		// If layer 2 we still need to set the container veth to up
 		if err := netns.Do(func(_ ns.NetNS) error {
-			link, err := netlink.LinkByName(args.IfName)
```
mlguerrero12, line 681: Why do you want a function for this? I understand the one for retries which includes some logic with the for loop but this is simple and not used anywhere else.

ormergi: My intent was to improve readability, even though there is no much of logic its less code to read, and may reduce some cognitive load (I dont need to know we need to fetch the link in order to bring it down).

Also, with #997 changes, it makes the code read better: 
``` 
...
else if !disableContIface {
        link.SetUp(ifaceName)
}
...
```
WDYT?

pkg/link/opstate.go
squeed: because all of this code was written before `%w` was a thing :-).

ormergi: Done

plugins/main/bridge/bridge.go
mlguerrero12: no, I think this is simple enough. If you want to improve readability, then abstract all the logic inside if layer3 branch, then in your other pr, you can do
`
if !disableContIface {
    if layer {
        layer3func()
    } else {
         thiscode
    }
    WaitOperStateUp()
}
`

PR #1003, opened 2024-01-24
Support DeviceID on Auxiliary Bus

adrianchiris:
Device plugins may allocate network device on a bus different than PCI.

sriov-network-device-plugin supports the allocation of network devices over Auxiliary bus[1][2][3].

extend host-device CNI to support such devices if provided through runtime config.

- Check if device provided by DeviceID runtime config is present on either PCI bus or Auxiliary bus
- extend getLink method to support getting netdev link obj from auxiliary bus
- add unit-test to cover the new flow

[1] https://github.com/k8snetworkplumbingwg/sriov-network-device-plugin/tree/master?tab=readme-ov-file#auxiliary-network-devices-selectors
[2] https://github.com/k8snetworkplumbingwg/sriov-network-device-plugin/tree/master/docs/subfunctions
[3] https://docs.kernel.org/networking/devlink/devlink-port.html

### DIFF
--- a/plugins/main/host-device/host-device.go
+++ b/plugins/main/host-device/host-device.go
@@ -37,7 +37,10 @@ import (
 	bv "github.com/containernetworking/plugins/pkg/utils/buildversion"
 )
 
-var sysBusPCI = "/sys/bus/pci/devices"
+var (
+	sysBusPCI       = "/sys/bus/pci/devices"
+	sysBusAuxiliary = "/sys/bus/auxiliary/devices"
+)
 
 // Array of different linux drivers bound to network device needed for DPDK
 var userspaceDrivers = []string{"vfio-pci", "uio_pci_generic", "igb_uio"}
@@ -53,6 +56,9 @@ type NetConf struct {
 	RuntimeConfig struct {
 		DeviceID string `json:"deviceID,omitempty"`
 	} `json:"runtimeConfig,omitempty"`
+
+	// for internal use
+	auxDevice string `json:"-"` // Auxiliary device name as appears on Auxiliary bus (/sys/bus/auxiliary)
 }
 
 func init() {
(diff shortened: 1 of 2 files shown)


plugins/main/host-device/host-device.go
```diff
@@ -53,6 +56,9 @@ type NetConf struct {
 	RuntimeConfig struct {
 		DeviceID string `json:"deviceID,omitempty"`
 	} `json:"runtimeConfig,omitempty"`
+
+	// for internal use
+	AuxDevice string `json:"-"` // Auxiliary device name as appears on Auxiliary bus (/sys/bus/auxiliary)
```
squeed, line 61: If this is internal, it the first word should be lower-cased. Yay golang.

adrianchiris: sure, will align.

noticed that `NetConf.DPDKMode` may also need to be internal as its inferred from the system[1].
but anyhow, not related to this PR.

[1] https://github.com/containernetworking/plugins/blob/d1aada912d434472ac3f709a1685cd9acfbb617a/plugins/main/host-device/host-device.go#L82

adrianchiris:
@squeed golangci-lint version bumped in gh actions which causes lint failure (not related to this PR)

submitted #1009 to keep version fixed so that CI is stable (version bumps IMO should occur periodically in controlled manner, i.e submitting PR)

s1061123 (approved):
Looks good to me, thanks @adrianchiris !

s1061123:
@squeed PTAL again?

s1061123:
@adrianchiris hi, CI (Lint) has some error. Could you please take a look into it?

adrianchiris:
> @adrianchiris hi, CI (Lint) has some error. Could you please take a look into it?

@s1061123  we should merge : #1009 then ill rebase this one and we should be good.

the lint issues not related to this PR but rather because linter version bumped in CI.

adrianchiris:
Rebased on top of main. CI is green.

@s1061123 @s1061123 PTAL on this one :)

PR #921, opened 2023-07-06
Dev/exclude subnets from traffic shaping2

oOraph:
bandwidth: possibility to exclude some subnets from traffic shapping

what changed:

we had to refactor the bandwidth plugin and switch from a classless qdisc (tbf)
to a classful qdisc (htb).

subnets are to be provided in config or runtimeconfig just like other parameters

unit and integration tests were also adapted in consequence

unrelated changes:

test fixes: the most important tests were just silently skipped due to ginkgo Measure deprecation
(the ones actually checking the effectiveness of the traffic control)

PR #1025, opened 2024-04-06
bandwidth: introduce configurable queue length

middaywords:
Currently, the bandwidth plugin uses tc tbf and set default latency to 25ms, which easily causes Head-Of-Line blocking and bufferbloat. This commit proposes to make it a configurable value.

### DIFF
--- a/plugins/meta/bandwidth/bandwidth_linux_test.go
+++ b/plugins/meta/bandwidth/bandwidth_linux_test.go
@@ -295,8 +295,10 @@ var _ = Describe("bandwidth test", func() {
 					"type": "bandwidth",
 					"egressRate": 0,
 					"egressBurst": 0,
+					"egressLatency": 0,
 					"ingressRate": 8000,
 					"ingressBurst": 80,
+					"ingressLatency": 25000,
 					"prevResult": {
 						"interfaces": [
 							{
--- a/plugins/meta/bandwidth/ifb_creator.go
+++ b/plugins/meta/bandwidth/ifb_creator.go
@@ -24,8 +24,6 @@ import (
 	"github.com/containernetworking/plugins/pkg/ip"
 )
 
-const latencyInMillis = 25
-
 func CreateIfb(ifbDeviceName string, mtu int) error {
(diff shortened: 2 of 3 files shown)


s1061123:
@middaywords thank you for the PR. But I suppose that it could be implemented by another CNI plugin as different purpose.
Currently `bandwidth` has lots of features and code and your feature is different (i.e. bandwidth is slightly different from latency, even though `tc` is used). `bandwidth` is not 'general TC CNI plugin', just for 'bandwidth'.  Having another plugin makes code simple and feature centric.

middaywords:
@s1061123 the title `latency` may be misleading, it means queue depth in tc tbf, i've changed it.
The key problem here is that, the bandwidth plugin uses a deep queue by default, and it's not configurable, which causes head-of-line blocking.
Specifically, when using the bandwidth plugin to limit the bandwidth, it brings extra latency about tens of milliseconds, which means it cannot go into real production env.

middaywords:
here is the test results when I test the plugin bandwidth using netperf, in local kind env. ping latency of server and client is about 50us, but the measured latency with traffic is about >10ms
| Data flow direction                                   | Throughput(Mbps) for TCP_STREAM flow | Latency(us) for TCP_RR flow |
| ----------------------------------------------------- | ------------------------------------ | --------------------------- |
| netperf-client -> netperf-server (egress rate limit)  | 95.6                                 | 10760.05                    |
| netperf-server -> netperf-client (ingress rate limit) | 95.6                                 | 226120.05                   |

s1061123:
Understand your comment. BTW, we'll merge this PR soon and you must need to rebase (it is pretty big).

https://github.com/containernetworking/plugins/pull/921

So please check the repo tomorrow and please rebase once repo is updated with above PR. Appreciated your help.

middaywords:
https://github.com/containernetworking/plugins/pull/921 uses HTB class for QoS and rate limiting, it can avoid close Head-Of-LIne blocking using many different subclasses, though its queue is not configurable.
Close this PR because it's no longer needed.

PR #1040, opened 2024-05-16
dependabot: don't automatically update cni packages

squeed:
We should manage their lifecycle separately.


s1061123 (approved):
Strongly agreed! LGTM!

PR #1050, opened 2024-05-27
Pass status along ipam

maiqueb:
This PR bumps the plugins to use CNI 1.2.0 and allows delegating the STATUS verb to IPAM plugins.

This way, other CNI plugins implementing STATUS can delegate the STATUS calls to IPAM CNIs.


LionelJouin (commented):
I looked at the Bandwidth tests. Since https://github.com/containernetworking/cni/pull/1039, the behavior of json.Marshall seems to have changed for struct that embeds `types.NetConf`.

With CNI v1.2.0: https://go.dev/play/p/L4feVRtl5S4
With CNI v1.1.2: https://go.dev/play/p/4RrrgKJUEN3

So, [here in bandwidth tests](https://github.com/containernetworking/plugins/blob/v1.5.0/plugins/meta/bandwidth/bandwidth_measure_linux_test.go#L161), all parameters are missing when calling ADD.

squeed:
@LionelJouin very good point. Do you think this affects more than just tests? I'm not sure if there's anything that embeds `types.NetConf` that also marshal to disk. But, we probably need to revert https://github.com/containernetworking/cni/pull/1039

plugins/main/loopback/loopback_test.go
```diff
@@ -74,7 +74,7 @@ var _ = Describe("Loopback", func() {
 				session, err := gexec.Start(command, GinkgoWriter, GinkgoWriter)
 				Expect(err).NotTo(HaveOccurred())
 
-				Eventually(session).Should(gbytes.Say(`{.*}`))
```
aojea, line 77: was this working before?

LionelJouin (approved):

squeed:
Is this ready to merge, or will you be implementing STATUS passthrough in this ?R

maiqueb:
> Is this ready to merge, or will you be implementing STATUS passthrough in this PR?

@squeed did you mean GC ? This PR does just that (i.e. status passthrough).

I'd rather defer GC to another PR.

squeed:
I don't see anything providing an implementation of Status, so this PR lays the groundwork but doesn't, AFAICT, pass STATUS to a delegated plugin.

maiqueb:
I see ! 

Yeah, let's defer that as well :)

squeed:
This was done in #1082

PR #1073, opened 2024-08-14
host-device: use temp network namespace for rename

### LINKED ISSUE #1072: host-device temp interface name causes race conditions
Since https://github.com/containernetworking/plugins/pull/1002 the host-device plugin renames the interface before moving it.

This doesn't play too well with udev and NetworkManager, see https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/issues/1599

Would it be possible to use a temporary netns, ie move to a temp netns, rename, move to the dest ns ?
@adrianchiris FYI

champtar:
Using a temporary name / doing a fast rename causes some race conditions with udev and NetworkManager: https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/issues/1599

Fixes #1072

### DIFF
--- a/pkg/ns/ns_linux.go
+++ b/pkg/ns/ns_linux.go
@@ -31,6 +31,10 @@ func GetCurrentNS() (NetNS, error) {
 	// return an unexpected network namespace.
 	runtime.LockOSThread()
 	defer runtime.UnlockOSThread()
+	return getCurrentNSNoLock()
+}
+
+func getCurrentNSNoLock() (NetNS, error) {
 	return GetNS(getCurrentThreadNetNSPath())
 }
 
@@ -152,6 +156,54 @@ func GetNS(nspath string) (NetNS, error) {
 	return &netNS{file: fd}, nil
 }
 
+// Returns a new empty NetNS.
+// Calling Close() let the kernel garbage collect the network namespace.
+func TempNetNS() (NetNS, error) {
+	var tempNS NetNS
+	var err error
+	var wg sync.WaitGroup
+	wg.Add(1)
+
+	// Create the new namespace in a new goroutine so that if we later fail
+	// to switch the namespace back to the original one, we can safely
+	// leave the thread locked to die without a risk of the current thread
+	// left lingering with incorrect namespace.
+	go func() {
+		defer wg.Done()
+		runtime.LockOSThread()
+
+		var threadNS NetNS
+		// save a handle to current network namespace
+		threadNS, err = getCurrentNSNoLock()
+		if err != nil {
+			err = fmt.Errorf("failed to open current namespace: %v", err)
+			return
+		}
+		defer threadNS.Close()
+
+		// create the temporary network namespace
+		err = unix.Unshare(unix.CLONE_NEWNET)
+		if err != nil {
+			return
+		}
+
+		// get a handle to the temporary network namespace
+		tempNS, err = getCurrentNSNoLock()
+
+		err2 := threadNS.Set()
+		if err2 == nil {
+			// Unlock the current thread only when we successfully switched back
+			// to the original namespace; otherwise leave the thread locked which
+			// will force the runtime to scrap the current thread, that is maybe
+			// not as optimal but at least always safe to do.
+			runtime.UnlockOSThread()
+		}
+	}()
+
+	wg.Wait()
+	return tempNS, err
+}
+
 func (ns *netNS) Path() string {
 	return ns.file.Name()
 }
@@ -173,7 +225,7 @@ func (ns *netNS) Do(toRun func(NetNS) error) error {
 	}
 
 	containedCall := func(hostNS NetNS) error {
-		threadNS, err := GetCurrentNS()
+		threadNS, err := getCurrentNSNoLock()
 		if err != nil {
 			return fmt.Errorf("failed to open current netns: %v", err)
 		}
--- a/plugins/main/host-device/host-device.go
+++ b/plugins/main/host-device/host-device.go
@@ -230,118 +230,121 @@ func cmdDel(args *skel.CmdArgs) error {
 	return nil
 }
 
-// setTempName sets a temporary name for netdevice, returns updated Link object or error
-// if occurred.
-func setTempName(dev netlink.Link) (netlink.Link, error) {
-	tempName := fmt.Sprintf("%s%d", "temp_", dev.Attrs().Index)
-
-	// rename to tempName
-	if err := netlink.LinkSetName(dev, tempName); err != nil {
-		return nil, fmt.Errorf("failed to rename device %q to %q: %v", dev.Attrs().Name, tempName, err)
-	}
-
-	// Get updated Link obj
-	tempDev, err := netlink.LinkByName(tempName)
-	if err != nil {
-		return nil, fmt.Errorf("failed to find %q after rename to %q: %v", dev.Attrs().Name, tempName, err)
-	}
-
-	return tempDev, nil
-}
-
-func moveLinkIn(hostDev netlink.Link, containerNs ns.NetNS, ifName string) (netlink.Link, error) {
-	origLinkFlags := hostDev.Attrs().Flags
+func moveLinkIn(hostDev netlink.Link, containerNs ns.NetNS, containerIfName string) (netlink.Link, error) {
 	hostDevName := hostDev.Attrs().Name
-	defaultNs, err := ns.GetCurrentNS()
+
+	// With recent kernels we could do all changes in a single netlink call,
+	// but on failure the device is left in a partially modified state.
+	// Doing changes one by one allow us to (try to) rollback to the initial state.
+
+	// Create a temporary namespace to rename (and modify) the device in.
+	// We were previously using a temporary name, but rapid rename leads to
+	// race condition with udev and NetworkManager.
+	tempNS, err := ns.TempNetNS()
 	if err != nil {
-		return nil, fmt.Errorf("failed to get host namespace: %v", err)
+		return nil, fmt.Errorf("failed to create tempNS: %v", err)
 	}
+	defer tempNS.Close()
 
-	// Devices can be renamed only when down
-	if err = netlink.LinkSetDown(hostDev); err != nil {
-		return nil, fmt.Errorf("failed to set %q down: %v", hostDev.Attrs().Name, err)
-	}
-
-	// restore original link state in case of error
-	defer func() {
-		if err != nil {
-			if origLinkFlags&net.FlagUp == net.FlagUp && hostDev != nil {
-				_ = netlink.LinkSetUp(hostDev)
+	// Restore original up state in case of error
+	// This must be done in the hostNS as moving
+	// device between namespaces sets the link down
+	if hostDev.Attrs().Flags&net.FlagUp == net.FlagUp {
+		defer func() {
+			if err != nil {
+				// lookup the device again (index might have changed)
+				if hostDev, err := netlink.LinkByName(hostDevName); err == nil {
+					_ = netlink.LinkSetUp(hostDev)
+				}
 			}
-		}
-	}()
-
-	hostDev, err = setTempName(hostDev)
-	if err != nil {
-		return nil, fmt.Errorf("failed to rename device %q to temporary name: %v", hostDevName, err)
+		}()
 	}
 
-	// restore original netdev name in case of error
-	defer func() {
-		if err != nil && hostDev != nil {
-			_ = netlink.LinkSetName(hostDev, hostDevName)
-		}
-	}()
-
-	if err = netlink.LinkSetNsFd(hostDev, int(containerNs.Fd())); err != nil {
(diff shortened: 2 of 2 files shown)


pkg/ns/ns_linux.go
```diff
@@ -152,6 +152,41 @@ func GetNS(nspath string) (NetNS, error) {
...
+		return nil, fmt.Errorf("failed to open current namespace: %v", err)
+	}
+	defer hostNS.Close()
+
+	var wg sync.WaitGroup
+	wg.Add(1)
+
+	var ns NetNS
+
+	go func() {
```
aojea, line 169: why spawn a goroutine if you are blocking until it finish and no other code runs?

champtar: Reworked a bit the function and added comments, it's the same as Do(), if we fail to switch the namespace back we leave the thread to die

champtar:
Bumping go to 1.21 so CI pass

champtar:
Rebased on main, removed go version bump

plugins/main/host-device/host-device.go
```diff
@@ -230,117 +230,93 @@ func cmdDel(args *skel.CmdArgs) error {
...
+				}
 			}
 		}()
 
-		// Save host device name into the container device's alias property
-		if err = netlink.LinkSetAlias(contDev, hostDevName); err != nil {
-			return fmt.Errorf("failed to set alias to %q: %v", tempDevName, err)
-		}
 		// Rename container device to respect args.IfName
-		if err = netlink.LinkSetName(contDev, ifName); err != nil {
```
aojea, line 316: I can't find where you set the device to `contDev` now, that should be the final name

champtar: `contDev` is `netlink.Link`, the interface name is `ifName`, see new line 270

champtar: reworked a bit more the PR with way more comments, hope it helps

champtar:
Rebased on main
@aojea do you have time to do a full review ?

champtar:
Maybe @squeed you have some time to review this ?

champtar:
Rebased

s1061123 (changes requested):
I don't have not so much time to review whole code, but I'm wondering that when temp network namespace is deleted, created by Unshare()?

`tempNS.Close()` just closes the netns handle(i.e. inode) in process, not remove netns. (currently `pkg/ns/ns_linux.go` is not designed to create/delete network ns, actually. Just manipulate 'already created' netns).

pkg/ns/ns_linux.go
```diff
@@ -152,6 +156,53 @@ func GetNS(nspath string) (NetNS, error) {
...
+		// create the temporary network namespace
+		err = unix.Unshare(unix.CLONE_NEWNET)
+		if err != nil {
+			return
+		}
+
+		// get a handle to the temporary network namespace
+		tempNS, err = getCurrentNSNoLock()
+
+		err2 := threadNS.Set()
```
champtar, line 192: > I don't have not so much time to review whole code, but I'm wondering that when temp network namespace is deleted, created by Unshare()?

> tempNS.Close() just closes the netns handle(i.e. inode) in process, not remove netns. (currently pkg/ns/ns_linux.go is not designed to create/delete network ns, actually. Just manipulate 'already created' netns).

@s1061123 
Just after Unshare(), I get the netns handle and revert back the thread to his original netns, so tempNS has no process. There is no call to delete a netns, as soon as there is no reference to a namespace (either a process  using it or a file handle), it'll be deleted.
https://man7.org/linux/man-pages/man7/namespaces.7.html
```
Opening one of the files in this directory (or a file that is
bind mounted to one of these files) returns a file handle for the
 corresponding namespace of the process specified by pid.  As long
as this file descriptor remains open, the namespace will remain
alive, even if all processes in the namespace terminate.  The
file descriptor can be passed to setns(2).
```

EdDev:
I would expect CNI plugins to execute on servers and not desktops. The auto-detection of new interfaces (i.e. detecting and managing them) is supposed to be disabled on servers that use NM.

Can that be an alternative solution?

pkg/ns/ns_linux.go
s1061123: > Opening one of the files in this directory (or a file that is
bind mounted to one of these files) returns a file handle for the
 corresponding namespace of the process specified by pid.  As long
as this file descriptor remains open, the namespace will remain
alive, even if all processes in the namespace terminate.  The
file descriptor can be passed to setns(2).

This does only mentions 'namespace is not deleted until namespace is open', not 'once all file descriptor is closed, the namespace is automatically removed'. 

> There is no call to delete a netns, as soon as there is no reference to a namespace (either a process using it or a file handle), it'll be deleted.

So please make it clear, what (which component/code) removes netns.

BTW, please add unit-test to guarantee the netns is removed after the close because resource leak is serious issue so we'd like to prevent it as mush as we could.

champtar:
> I would expect CNI plugins to execute on servers and not desktops. The auto-detection of new interfaces (i.e. detecting and managing them) is supposed to be disabled on servers that use NM.
> 
> Can that be an alternative solution?

I'm already using `no-auto-default=*` in NM config, but I use cockpit and want some interfaces (identified by mac pattern) not to be configurable at all (`NM_UNMANAGED`), and this race condition makes it impossible

EdDev:
> I'm already using `no-auto-default=*` in NM config, but I use cockpit and want some interfaces (identified by mac pattern) not to be configurable at all (`NM_UNMANAGED`), and this race condition makes it impossible

I think the basic is setting the `no-auto-default` and `ignore-carrier` as set with the `NetworkManager-config-server` package. But I think NM has more options in this regard.
If you already exhausted this direction with the NM experts, then I guess that is it.

The proposed solution you provided to overcome the problem is innovative, but at the same time feels a bit hacky. It attempts to overcome a problem of NM & udev, which a general CNI should not care about.
I guess NM had not faced this problem before because this is not a common action, but if we think about it, the same operation could have been done by a human or automation, regardless of a CNI, and one could expect NM to still behave correctly. Solving this at the core, would solve the CNI scenario as well.

If the workaround is accepted here, then at the minimum I would suggest to have an open bug on the NM project, asking to solve it there and ref in the code. I.e. making the current solution temporary.

pkg/ns/ns_linux.go
champtar: > This does only mentions 'namespace is not deleted until namespace is open', not 'once all file descriptor is closed, the namespace is automatically removed'.

Here a better part of the same man page
```
Namespace lifetime
       Absent any other factors, a namespace is automatically torn down
       when the last process in the namespace terminates or leaves the
       namespace.  However, there are a number of other factors that may
       pin a namespace into existence even though it has no member
       processes.  These factors include the following:

       •  An open file descriptor or a bind mount exists for the
          corresponding /proc/pid/ns/* file.
```

> So please make it clear, what (which component/code) removes netns.

Will improve the comment before the function

> BTW, please add unit-test to guarantee the netns is removed after the close because resource leak is serious issue so we'd like to prevent it as mush as we could.

There is no syscall to list namespaces, you might be confused by the /var/run/netns folder, this is just a convention for "named netns" and if you just create a new netns it'll not appear magically in /var/run/netns

BTW cni repo uses vishvananda/netns
https://github.com/containernetworking/cni/blob/main/pkg/ns/ns_linux.go#L20

Basic example for a tempNetNS (New + Close)
https://github.com/vishvananda/netns/blob/4c46424d73b556b3ea4bc5a7cec9e7376dcb2a73/README.md?plain=1#L40

New is just Unshare + Get
https://github.com/vishvananda/netns/blob/4c46424d73b556b3ea4bc5a7cec9e7376dcb2a73/netns_linux.go#L42
Close just closes the file descriptor
https://github.com/vishvananda/netns/blob/4c46424d73b556b3ea4bc5a7cec9e7376dcb2a73/nshandle_linux.go#L63

NewNamed() if you are curious
https://github.com/vishvananda/netns/blob/4c46424d73b556b3ea4bc5a7cec9e7376dcb2a73/netns_linux.go#L50

In our code we are a bit safer because we are using os.File (so even if we we forget to call Close golang will do it for us)

champtar:
> > I'm already using `no-auto-default=*` in NM config, but I use cockpit and want some interfaces (identified by mac pattern) not to be configurable at all (`NM_UNMANAGED`), and this race condition makes it impossible
> 
> I think the basic is setting the `no-auto-default` and `ignore-carrier` as set with the `NetworkManager-config-server` package. But I think NM has more options in this regard. If you already exhausted this direction with the NM experts, then I guess that is it.
> 
> The proposed solution you provided to overcome the problem is innovative, but at the same time feels a bit hacky. It attempts to overcome a problem of NM & udev, which a general CNI should not care about. I guess NM had not faced this problem before because this is not a common action, but if we think about it, the same operation could have been done by a human or automation, regardless of a CNI, and one could expect NM to still behave correctly. Solving this at the core, would solve the CNI scenario as well.

Agreed that NM should do better, but the bug is 3 weeks old and still waiting for triage.
Also having the NM fix backported will take a long time.

> If the workaround is accepted here, then at the minimum I would suggest to have an open bug on the NM project, asking to solve it there and ref in the code. I.e. making the current solution temporary.

My fix also leads to less udev events / rules execution, so I really don't see it as a workaround
And maybe some other software handling network configuration are impacted, each distro uses its own ...

pkg/ns/ns_linux.go
s1061123: I stopped to review the code, because it is still unclear to me  and you don't answer my question: what (which component/code) removes netns.  I leave it to other reviwers to review.

champtar: Calling Close() will remove the netns, more precisely it will let the kernel garbage collect it as this is the only reference we have to it.
We can only have a resource leak if we don't close the open file handle to the namespace, and we use os.File so this can't happen (os.File call close in his finalizer if you forgot to)

Is it clear enough ?

champtar: (I've added `// Calling Close() let the kernel garbage collect the network namespace.`)

champtar:
>  Fixing it at the "edge" is not scaling well

Depending on how you write your udev rules they might be impacted as well, for example if you do negative matches `ATTR{address}!="00:11:22:*"` this will match because the device doesn't exists, so `ATTR{address}` is empty

Fast rename is also problematic for udev alone, and race condition are always fun to debug

> and burdens this CNI logic.

it's the same number of steps

With temp name:
- link down
- rename 1
- change namespace
- rename 2

With temp namespace:
- create a temp namespace
- change namespace (this also down the link)
- rename
- change namespace

pkg/ns/ns_linux.go
champtar: @s1061123 are my latest explanations clear enough ?

champtar: Hi @s1061123,
this is marked a "change requested", but don't see any changes to do

s1061123: > Hi @s1061123,
this is marked a "change requested", but don't see any changes to do

Here is what I mentioned, and as above, I stopped to review the code, because it is still unclear to me , you don't answer my question: what (which component/code) removes netns and I don't have enough time to review. I leave it to other reviwers to review. So please find someone else to review the PR.

---
I don't have not so much time to review whole code, but I'm wondering that when temp network namespace is deleted, created by Unshare()?

tempNS.Close() just closes the netns handle(i.e. inode) in process, not remove netns. (currently pkg/ns/ns_linux.go is not designed to create/delete network ns, actually. Just manipulate 'already created' netns).

champtar: I've answered multiple times that calling Close() is enough to remove the netns, I've shared documentation to support that and other pieces of go code doing the same thing.
I totally understand if you don't have time, and I'm fine having someone else review (any volunteers ?),
simply if the change requested is "please do not leak ressources / netns", then please remove the change requested status as this code does not leak a netns.

s1061123: But let me leave it for further reviewers (still I am not conclude 'we should have temp network namespace').

squeed:
Honestly, this makes sense to me. I understand that NM changes can take far too long to roll out, especially when there is no appreciable workaround.

Personally, I'm in favor of this PR. It's not like namespaces are so expensive. IIRC there have been other issues regarding "churn" as the host-network device is reconfigured, so this seems like a reasonable change.

champtar:
@squeed as it make sense to you, any chance you can review this PR ?

EdDev:
> > and burdens this CNI logic.
> 
> it's the same number of steps
> 
> With temp name:
>     * link down
>     * rename 1
>     * change namespace
>     * rename 2
> 
> 
> With temp namespace:
>     * create a temp namespace
>     * change namespace (this also down the link)
>     * rename
>     * change namespace

Ok, this argument is convincing.

squeed (approved):

champtar:
This as been approved by @squeed for a week, ready for review for a good month, how can we move forward ?

squeed:
@champtar we’ll be cutting a release in a week or so, this will get in.

PR #1083, opened 2024-08-28
update deps (go-iptables, cni)

danwinship:
Splitting this out from #935 for clarity / ease of debugging.

Closes #1068

cc @squeed


plugins/main/loopback/loopback_test.go
```diff
@@ -74,7 +74,8 @@ var _ = Describe("Loopback", func() {
 				session, err := gexec.Start(command, GinkgoWriter, GinkgoWriter)
 				Expect(err).NotTo(HaveOccurred())
 
-				Eventually(session).Should(gbytes.Say(`{.*}`))
+				// "(?s)" turns on the "s" flag, making "." match newlines too.
+				Eventually(session).Should(gbytes.Say(`(?s){.*}`))
```
danwinship, line 78: Clearly this was intended to match the result as a whole, but actually, it was only matching

    "dns": {}

because that was the only place where there was a `{` followed by a `}` _on the same line_. And then https://github.com/containernetworking/cni/pull/1039 broke that...

squeed (approved):

PR #1091, opened 2024-09-14
dhcp: Add priority option to dhcp.

lisongmin:
Currently, we can not set the metric of routes in dhcp. It's ok if there is only a network interface.

But if there are multiple network interfaces, and both have a default route, We need to set the metric of the route to make the traffic go through the correct network interface.

For host-local and static, we can set the metric with the route.priority option. But there is no such option for dhcp.


pkg/ipam/ipam_linux.go
```diff
@@ -117,6 +117,7 @@ func ConfigureIface(ifName string, res *current.Result) error {
 			Dst:       &r.Dst,
 			LinkIndex: link.Attrs().Index,
 			Gw:        gw,
+			Priority:  r.Priority,
```
LionelJouin, line 120: The error could return the priority `return fmt.Errorf("failed to add route '%v via...`

LionelJouin (approved):

squeed (approved):

PR #1100, opened 2024-09-23
Fix inadvertent txqueuelen being set to zero

gudmundur:
When diagnosing networking issues, I noticed that `txqueuelen` on `veth` pairs and `tap` devices was set to zero. In investigating the issue, I found that most usages of `netlink.LinkAttrs` in this repository uses the struct literal, which in the `netlink` docs is advised against unless `TxQLen` is explicitly set.

> Note `NewLinkAttrs` constructor, it sets default values in structure. For now it sets only `TxQLen` to `-1`, so kernel will set default by itself. If you're using simple initialization(`LinkAttrs{Name: "foo"}`) `TxQLen` will be set to `0` unless you specify it like `LinkAttrs{Name: "foo", TxQLen: 1000}`.
> ([link](https://pkg.go.dev/github.com/vishvananda/netlink#section-readme))

When using the `ip` command directly without specifying `txqueuelen`, Linux will automatically to 1000:

```sh
$ ip tuntap add tap0 mode tap
$ ip link show tap0
4: tap0: <BROADCAST,MULTICAST> mtu 1500 qdisc noop state DOWN mode DEFAULT group default qlen 1000
    link/ether 8a:06:b2:30:4d:1e brd ff:ff:ff:ff:ff:ff
```

This change goes through every single usage of `netlink.LinkAttrs{` and replaces it with the `netlink.NewLinkAttrs` constructor, and then overrides the individual fields in the struct.

I even noticed that in the `bridge` plugin, this is [somewhat called out](https://github.com/containernetworking/plugins/blob/fa737f82b2e0268e67bdf6e1ca18ad4ae7393469/plugins/main/bridge/bridge.go#L342-L346) when using the literal. I've replaced this explicitly in d376c8df5088e62c3a993ce14e5653b7d37407a0 though.

### DIFF
--- a/pkg/ip/link_linux.go
+++ b/pkg/ip/link_linux.go
@@ -32,11 +32,12 @@ var ErrLinkNotFound = errors.New("link not found")
 
 // makeVethPair is called from within the container's network namespace
 func makeVethPair(name, peer string, mtu int, mac string, hostNS ns.NetNS) (netlink.Link, error) {
+	linkAttrs := netlink.NewLinkAttrs()
+	linkAttrs.Name = name
+	linkAttrs.MTU = mtu
+
 	veth := &netlink.Veth{
-		LinkAttrs: netlink.LinkAttrs{
-			Name: name,
-			MTU:  mtu,
-		},
+		LinkAttrs:     linkAttrs,
 		PeerName:      peer,
 		PeerNamespace: netlink.NsFd(int(hostNS.Fd())),
 	}
--- a/pkg/ipam/ipam_linux_test.go
+++ b/pkg/ipam/ipam_linux_test.go
@@ -56,11 +56,12 @@ var _ = Describe("ConfigureIface", func() {
 		err = originalNS.Do(func(ns.NetNS) error {
 			defer GinkgoRecover()
 
+			linkAttrs := netlink.NewLinkAttrs()
+			linkAttrs.Name = LINK_NAME
+
 			// Add master
 			err = netlink.LinkAdd(&netlink.Dummy{
-				LinkAttrs: netlink.LinkAttrs{
-					Name: LINK_NAME,
-				},
+				LinkAttrs: linkAttrs,
 			})
 			Expect(err).NotTo(HaveOccurred())
 			_, err = netlink.LinkByName(LINK_NAME)
--- a/pkg/ns/README.md
+++ b/pkg/ns/README.md
@@ -13,10 +13,10 @@ The `ns.Do()` method provides **partial** control over network namespaces for yo
 
 ```go
 err = targetNs.Do(func(hostNs ns.NetNS) error {
+	linkAttrs := netlink.NewLinkAttrs()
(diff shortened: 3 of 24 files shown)


gudmundur:
Been trying to sort out what's up with the failing `sbr` test and have been unable to reproduce. I see that it's also failing on another recent pull request (https://github.com/containernetworking/plugins/pull/1099, [failing test](https://github.com/containernetworking/plugins/actions/runs/10986451243/job/30499889878#step:8:397)). Happy to provide a fix if I had a way to reproduce the failure. 😭

squeed:
Thanks for the PR. Have you seen #1097? Are these two PRs duplicate or complementary?

gudmundur:
@squeed The second issue #1097 describes is somewhat related, but it misses the mark on setting it to zero meaning that the kernel will pick a default.

This can be illustrated as such:

```sh
$ ip link add dev vm1 type veth
$ ip link show vm1
... snip snip snip ...
20: vm1@vm2: <BROADCAST,MULTICAST,M-DOWN> mtu 1500 qdisc noop state DOWN mode DEFAULT group default qlen 1000
    link/ether 76:8b:12:5d:e9:c0 brd ff:ff:ff:ff:ff:ff
```

Note the `qlen 1000` on the interface. Now if we run the same command, but set `txqueuelen 0` the following happens:

```sh
$ ip link add dev vm1 txqueuelen 0 type veth
$ ip link show vm1
24: vm1@veth0: <BROADCAST,MULTICAST,M-DOWN> mtu 1500 qdisc noop state DOWN mode DEFAULT group default 
    link/ether 6a:19:4e:0b:3a:5e brd ff:ff:ff:ff:ff:ff
```

Note the missing `qlen 1000`. If you look at any regular device that is created by `ip` they will always have a `qlen 1000`. The `netlink` library [propagates the 0 to the syscall when creating the link](https://github.com/vishvananda/netlink/blob/6f5713947556a0288c5cb71f036f9e91924ebcaa/link_linux.go#L1608-L1611). I remember before fixing this, that we would [see a handful of these log lines in the kernel logs](https://github.com/torvalds/linux/blob/075dbe9f6e3c21596c5245826a4ee1f1c1676eb8/net/sched/sch_api.c#L1338). In digging a little through the Linux codebase it seems that a value of zero is [caught in some places](https://github.com/search?q=repo%3Atorvalds%2Flinux%20DEFAULT_TX_QUEUE_LEN&type=code). In Googling this a little more, I also found that [the Cilium folks had this issue in the past](https://github.com/cilium/cilium/issues/17703).

What I did notice on our end when we worked around this issue, was that we saw less tx packet drops. Now I can't remember the specifics on whether that was on `veth` or `tap`.

squeed:
(@LionelJouin found the cause of the fake, he should have a fix soon)

squeed:
Rebased to pick up SBR fixes.

gudmundur:
@squeed Took care of the squashing for you. Reworded the commit message for a single commit as well.

PR #1120, opened 2024-11-11
ipmasq: fix nftables backend

### LINKED ISSUE #1118: ipmasq nftables doesn't support multiple ips (dual stack)
dual stack setup
```
{
  "type": "ptp",
  "ipMasq": true,
  "ipMasqBackend": "nftables",
  "ipam": {
    "type": "host-local",
    "ranges": [
      [{"subnet": "198.18.0.0/17"}],
      [{"subnet": "fd61:7465:6d65:1000::/112"}]
    ],
    "routes": [
      { "dst": "0.0.0.0/0" },
      { "dst": "::/0" },
    ]
  }
},
```

Looking at `nft list ruleset`, only the `ip6` rules are present in `cni_plugins_masquerade` table
Looking at `nft monitor rules`, we see that the `ip` rules are added then deleted

https://github.com/containernetworking/plugins/blob/fec2d62676cbe4f2fd587b4840c7fc021bead3f9/pkg/ip/ipmasq_nftables_linux.go#L84-L85
In `setupIPMasqNFTablesWithInterface` the stale rule logic is incorrect

champtar:
Add SetupIPMasqForNetworks and TeardownIPMasqForNetworks functions that take []*net.IPNet instead of *net.IPNet.

This allow the nftables backend to cleanup stale rules and recreate all needed rules in a single transaction, where previously the stale rules cleanup was breaking all but the last IPNet.

Fixes 61d078645a6d2a2391a1555ecda3d0a080a45831
Fixes #1118

Comments for reviewer:
I kept `SetupIPMasqForNetwork` (without the `s`) but it's broken for nftables if you call it in loop, so might be better to just break the API and get rid of it, as it was just released in 1.6.0.


champtar:
Tested with
```
{
  "type": "ptp",
  "ipMasq": true,
  "ipMasqBackend": "nftables",
  "ipam": {
    "type": "host-local",
    "ranges": [
      [{"subnet": "198.18.0.0/17"}],
      [{"subnet": "fd61:7465:6d65:1000::/112"}]
    ],
    "routes": [
      { "dst": "0.0.0.0/0" },
      { "dst": "198.18.128.0/17" },
      { "dst": "198.19.254.254/32" },
      { "dst": "::/0" },
      { "dst": "fd61:7465:6d65:2000::/112" },
      { "dst": "fd61:7465:6d65:ffff::/128" }
    ]
  }
},
```

champtar:
@danwinship if you can review this one

danwinship:
> I kept `SetupIPMasqForNetwork` (without the `s`) but it's broken for nftables if you call it in loop, so might be better to just break the API and get rid of it, as it was just released in 1.6.0.

I don't like keeping the broken function, but I don't know what @squeed will think about removing it, API-compatibility-wise.

It's also possible to avoid needing a new function, by including the CIDR in the comment hash so each one will get a different comment.

danwinship:
needs a unit test that fails with the old code and passes with the new code

champtar:
> > I kept `SetupIPMasqForNetwork` (without the `s`) but it's broken for nftables if you call it in loop, so might be better to just break the API and get rid of it, as it was just released in 1.6.0.
> 
> I don't like keeping the broken function, but I don't know what @squeed will think about removing it, API-compatibility-wise.
> 
> It's also possible to avoid needing a new function, by including the CIDR in the comment hash so each one will get a different comment.

Right now `teardownIPMasqNFTablesWithInterface` completely ignores the IPNets and clean everything, which makes a lot of sense to me

> needs a unit test that fails with the old code and passes with the new code

will do

danwinship:
OK, Casey says `containernetworking/plugins` has no API guarantees, so you can just drop the old function.

But you should still add a unit test

danwinship (approved):
lgtm

danwinship:
@squeed this one should be ready to merge too

squeed (approved):

PR #1123, opened 2024-11-20
Bump Go version to 1.23 in scripts/release.sh

jingyuanliang:


MrHohn (approved):
LGTM. Seems right to me to match up the version bumped in https://github.com/containernetworking/plugins/pull/1094.
/cc @squeed

s1061123 (approved):
It looks good to me. Merge it soon. 

BTW, we no longer uses this script file because it is for 'manual release process that is used previously'. Now we use github action to automate process. So this file (`scripts/release.sh`) will be removed soon. Just FYI.

PR #1133, opened 2025-01-06
Split CI and go.mod versions

squeed:
Downstream users would like to lower the minimum required go version, but it would be nice to test and release with the latest go. So, use a placeholder go version file for CI.


.github/go-version
```diff
@@ -0,0 +1 @@
+go 1.23
```
s1061123, line 1: Need to newline at the end of file.

s1061123 (changes requested):

.github/go-version
```diff
@@ -0,0 +1 @@
+1.23
```
s1061123, line 1: Need to newline at the end of file. (you can see red circle mark)
https://gist.github.com/camh-/1bebfcff1b0f814e9b191edc60d5206b

s1061123 (changes requested):

.github/go-version
squeed: oops, fixed.

PR #1134, opened 2025-01-06
Remove scripts/release.sh because of no longer used

s1061123:
scripts/release.sh is used for release plugins manually (by maintainer's hand), previously. Now we introduced automated release process by github action, hence it is no longer used and no longer maintained. This change removes this file. Thanks, release.sh for a long time!


s1061123:
Need to fix github CI, close it and file another one.

PR #1138, opened 2025-01-14
test: enable unpriv user namespaces

squeed:
These are disabled by default in some distros; we would like to test rootless, however.

### DIFF
--- a/test_linux.sh
+++ b/test_linux.sh
@@ -1,4 +1,4 @@
-#!/usr/bin/env sh
+#!/usr/bin/env bash
 #
 # Run CNI plugin tests.
 # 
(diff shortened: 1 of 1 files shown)


squeed:
Aha, that fixes it @s1061123 :-)

test_linux.sh
```diff
@@ -18,6 +18,23 @@ testrun() {
...
+    local existing
+
+    key="$1"
+    val="$2"
+    existing="$(sysctl -ben "$key")"
+
+    sysctl -r 
+
+    if [ "$val" -ne "$existing" ]; then
+        echo "sudo sysctl -w '$key'='$val'"
```
s1061123, line 33: suppose we should make consistent between code and message. `s/-w/-we/`

s1061123 (changes requested):
Thank you so much for the fix, @squeed !
I have only just minor comment. Almost good. Once changed, I approve it.

s1061123 (approved):
LGTM, thank you for quick fix!

Issue #1107, opened 2024-10-15
Does bridge plugin support disabling inter container connectivity?


swagatbora90:
Hi team,

I have a use-case where I need to disable intercontainer connectivity between containers attached to the same bridge network. With docker, I can pass `com.docker.network.bridge.enable_icc=false` option during `network create` which disables inter container connectivity. So I can do something like:
```
% docker network create --driver bridge --opt com.docker.network.bridge.enable_icc=false test_icc_disabled
1a0fc02fa29869f0301e2727defa1e9f90ba5c46825d67c74ec9973d1a42dbea

% docker run -d --name container1 --network test_icc_disabled busybox sleep 3600
8afe7c27e3757850c1583844ec12053febf9ab174fc3e1ec9ab2284db81ce002

% docker run -d --name container2 --network test_icc_disabled busybox sleep 3600
453c08e5fed68357a98668a4c96a8a02a57606869ef5aba7a043af27c02f7c73

% docker exec container2 ping -c 4 container1
PING container1 (240.11.2.2): 56 data bytes

--- container1 ping statistics ---
4 packets transmitted, 0 packets received, 100% packet loss
```

I am trying to find out if the bridge plugin provides a similar mechanism to disable the default behavior through config?

henry118:
The `com.docker.network.bridge.enable_icc=false` docker config will result in the following iptable rule in the host's forward chain (where 'docker0' is the bridge name):

```
-A FORWARD -i docker0 -o docker0 -j DROP
```

I'm not aware of a similar functionality in CNI's bridge plugin. Is it something the community would consider to add as a new feature? If so we we will be happy to contribute a patch.

cc @squeed

swagatbora90:
Had a discussion about this issue during the last community meeting. Overall, no major concerns were raised regarding the proposed feature for the bridge plugin. The maintainers suggested that we move forward and open a PR with the proposed changes. I'll go ahead and assign this to myself.

Thanks for the feedback!

PR #1140, opened 2025-01-15
Add a new firewall ingress-policy "isolated"

swagatbora90:
IngressPolicyIsolated ("isolated") behaves  similar to ingress policy "same-bridge" with the exception that connections from the same bridge are also blocked. This is meant to be functionally equivalent to Docker network option "enable_icc" when set to false.

Resolves: #1107


swagatbora90:
PTAL @squeed @henry118

PR #1143, opened 2025-01-21
DHCP lease maintenance should terminate when interface no longer exists.

dougbtv:
Due to oberservations that threads can grow and the dhcp daemon uses an increasing amount of memory.

This situation can happen organically when using say, bridge CNI, and the bridge has been removed outside of the bridge CNI lifecycle, and an interface no longer exists on a pod.

Also, for reproduction steps (admittedly, openshift biased), see my steps in https://gist.github.com/dougbtv/803316016c1e96a529725013402d5f0d


plugins/ipam/dhcp/lease.go
```diff
@@ -292,6 +292,11 @@ func (l *DHCPLease) maintain() {
 	for {
 		var sleepDur time.Duration
 
+		if _, err := netlink.LinkByName(l.link.Attrs().Name); err != nil {
+			log.Printf("%v: interface %s no longer exists, terminating lease maintenance", l.clientID, l.link.Attrs().Name)
+			return
```
dougbtv, line 297: I was going to release at this point, but, in my testing, if the interface no longer exists -- the call to release fails.

squeed: We had a problem in Cilium with LinkByName returning ErrDumpInterrupted on churny nodes. I wonder if we should wrap that here in a retry too.

plugins/ipam/dhcp/lease.go
```diff
@@ -344,6 +358,23 @@ func (l *DHCPLease) maintain() {
...
+		// Returns the error to trigger a retry.
+        if _, err := netlink.LinkByName(l.link.Attrs().Name); err != nil {
+            return nil, err
+        }
+        return nil, nil
+    }
+
+    ctx, cancel := context.WithTimeout(ctx, linkCheckTotalTimeout)
+    defer cancel()
+    _, err := backoffRetry(ctx, linkCheckRetryMax, checkFunc)
```
dougbtv, line 373: the only downside here is that `backoffRetry()` sets a "base delay" using the `resendDelay0` const which was from the DHCP RFC stuff, so, it's 4 seconds the first time. This could probably be quicker? Will happily rejig backoffRetry to take a base delay parameter.

dougbtv: oops missed a major part of this... fix incoming.

dougbtv:
Soooo.... yeah what I have currently pushed has some things I totally whiffed on and I'm working on correcting it, but, in an ideal world, the netlink library release v1.3.0 doesn't include [the commit to detect the ErrDumpInterrupted](https://github.com/vishvananda/netlink/commit/084abd93d350e97ee5410b5b6311bcc211f7ea05)

I'm looking into working around it until that can get released, which I also requested in https://github.com/vishvananda/netlink/issues/1019

plugins/ipam/dhcp/lease.go
```diff
@@ -344,6 +362,51 @@ func (l *DHCPLease) maintain() {
...
+	}
+
+	ctx, cancel := context.WithTimeout(ctx, linkCheckTotalTimeout)
+	defer cancel()
+
+	exists, err := backoffRetryLinkExists(ctx, linkCheckRetryMax, checkFunc)
+	return exists, err
+}
+
+func backoffRetryLinkExists(ctx context.Context, maxDelay time.Duration, f func() (bool, error)) (bool, error) {
```
dougbtv, line 389: I wound up creating another method because I needed to have the retry method ` f func() (bool, error)` return a bool and an error, so, I additionally added `linkCheckDelay0`

plugins/ipam/dhcp/lease.go
```diff
@@ -344,6 +362,51 @@ func (l *DHCPLease) maintain() {
 	}
 }
 
+func (l *DHCPLease) checkLinkExistsWithBackoff(ctx context.Context) (bool, error) {
+	checkFunc := func() (bool, error) {
+		_, err := netlink.LinkByName(l.link.Attrs().Name)
+		if err != nil {
+			log.Printf("!bang: Error checking network link '%s': %v, type: %v", l.link.Attrs().Name, err, reflect.TypeOf(err))
+			var linkNotFoundErr *netlink.LinkNotFoundError
+			if errors.As(err, &linkNotFoundErr) {
```
dougbtv, line 371: This wasn't evaluating to true even though I was seeing logs as:

```
2025/01/21 21:36:46 Error checking network link 'net1': Link not found, type: netlink.LinkNotFoundError
```

I'll remove the reflect package and also the lines marked with `!bang` once I get it sorted.

dougbtv:
I'm going to change to checking for the string `Link Not Found` which I'm not particularly stoked on, but, happy to take any input on why this didn't work:

```
var linkNotFoundErr *netlink.LinkNotFoundError
if errors.As(err, &linkNotFoundErr) {
	log.Printf("!bang Link not found error detected! this is what I want to see.")
	return false, nil
}
```

That would never evaluate as true, even when I could find the error as `netlink.LinkNotFoundError` when using a reflect output like this: `log.Printf("!bang: Error checking network link '%s': %v, type: %v", l.link.Attrs().Name, err, reflect.TypeOf(err))`

zeeke:
> I'm going to change to checking for the string `Link Not Found` which I'm not particularly stoked on, but, happy to take any input on why this didn't work:
> 
> ```
> var linkNotFoundErr *netlink.LinkNotFoundError
> if errors.As(err, &linkNotFoundErr) {
> 	log.Printf("!bang Link not found error detected! this is what I want to see.")
> 	return false, nil
> }
> ```
> 
> That would never evaluate as true, even when I could find the error as `netlink.LinkNotFoundError` when using a reflect output like this: `log.Printf("!bang: Error checking network link '%s': %v, type: %v", l.link.Attrs().Name, err, reflect.TypeOf(err))`

You're passing the address of an empty pointer (pointer of pointer) to `errors.As(...)`.

This should do the job

```
var linkNotFoundErr *netlink.LinkNotFoundError = &netlink.LinkNotFoundError{}
if errors.As(err, linkNotFoundErr) {
	log.Printf("!bang Link not found error detected! this is what I want to see.")
	return false, nil
}
```

Try it in https://go.dev/play/p/GZJO1RdVR8g

plugins/ipam/dhcp/lease.go
```diff
@@ -344,6 +360,45 @@ func (l *DHCPLease) maintain() {
 	}
 }
 
+func (l *DHCPLease) checkLinkExistsWithBackoff(ctx context.Context) (bool, error) {
+	checkFunc := func() (bool, error) {
+		_, err := netlink.LinkByName(l.link.Attrs().Name)
+		if err != nil {
+			var linkNotFoundErr *netlink.LinkNotFoundError = &netlink.LinkNotFoundError{}
+			if errors.As(err, linkNotFoundErr) {
```
dougbtv, line 368: Owe you one @zeeke !!! Thanks. Appreciate that, it works a charm and much cleaner than the string check, thanks.

plugins/ipam/dhcp/lease.go
```diff
@@ -344,6 +360,45 @@ func (l *DHCPLease) maintain() {
...
+			}
+			return false, err
+		}
+		return true, nil
+	}
+
+	ctx, cancel := context.WithTimeout(ctx, linkCheckTotalTimeout)
+	defer cancel()
+
+	exists, err := backoffRetryLinkExists(ctx, linkCheckRetryMax, checkFunc)
```
s1061123, line 379: `linkCheckRetryMax` is constant variable, defined as global scope in this file. So we may remove it from function argument. Otherwise, remove constant variable, `linkCheckRetryMax` and define it in function variable.

plugins/ipam/dhcp/lease.go
```diff
@@ -344,6 +360,45 @@ func (l *DHCPLease) maintain() {
...
+	return exists, err
+}
+
+func backoffRetryLinkExists(ctx context.Context, maxDelay time.Duration, f func() (bool, error)) (bool, error) {
+	baseDelay := linkCheckDelay0
+	for {
+		exists, err := f()
```
s1061123, line 386: Currently, this function is pretty for `backoffRetryLinkExists`, not for common retry, hence we may not required to have ` f func() (bool, error))` as argument, I guess. Instead of that, having `linkName` in function argument makes it simple, I feel.

```
func backoffRetryLinkExists(ctx context.Context, maxDelay time.Duration, linkName string) (bool, error) {
  checkFunc := func() (bool, error) {
        _, err := netlink.LinkByName(linkName)
        if err != nil {
            var linkNotFoundErr *netlink.LinkNotFoundError = &netlink.LinkNotFoundError{}
            if errors.As(err, linkNotFoundErr) {
                return false, nil
            }
            return false, err
        }
        return true, nil
    }
```

plugins/ipam/dhcp/lease.go
```diff
@@ -344,6 +360,45 @@ func (l *DHCPLease) maintain() {
 	}
 }
 
+func (l *DHCPLease) checkLinkExistsWithBackoff(ctx context.Context) (bool, error) {
+	checkFunc := func() (bool, error) {
+		_, err := netlink.LinkByName(l.link.Attrs().Name)
```
s1061123, line 365: If we assume that this link is already removed, then `l.link.Attrs()` may be invalid. We should store interface name ahead.

s1061123 (changes requested):

plugins/ipam/dhcp/lease.go
dougbtv: thanks, simplified so that it isn't passed through the method param and just used directly there, good call

squeed (approved):

zeeke (approved):

SchSeba (approved):
LGTM!

s1061123 (approved):
LGTM!

SchSeba:
Hi @s1061123 @squeed,
can we merge this one?

s1061123:
I could not secure the time to review this week.
 Could you ask someone else?

s1061123:
@squeed Could you check merge criteria and merge that it is okey? This PR already has two maingainer approval.

squeed (approved):

PR #1144, opened 2025-01-26
Add rule for outbound interface when there is a single interface IP

neilcook:
When using the sbr plugin with a CNI plugin such as multus, applications have to bind to the source IP address of the interface for the str plugin to work correctly. They cannot bind to the interface *name*, because the sbr plugin doesn't currently add a rule for the interface name.

This is problematic because it is straightforward to configure the name of a multus interface, however the IP address is usually assigned from a range, so the exact interface IP address to use is not known in advance unlike the interface name. It would thus be much easier for applications if the sbr plugin added a rule with the interface name in addition to the rule with the interface address.

This PR does exactly that - adds an additional rule with the outbound interface name, so that applications can configure the name of the interface to bind to, rather than having to know the interface IP address.

Obviously, this approach does not work if there are multiple IP addresses for the interface, so this change only adds the rule for the outbound interface name if there is a single IP address on that interface.

I have added tests for the new rule, which all pass.


plugins/meta/sbr/main.go
```diff
@@ -253,12 +253,23 @@ func doRoutes(ipCfgs []*current.IPConfig, iface string) error {
...
+		srcRule.Src = &src
 
-		if err = netlink.RuleAdd(rule); err != nil {
-			return fmt.Errorf("Failed to add rule: %v", err)
+		if err = netlink.RuleAdd(srcRule); err != nil {
+			return fmt.Errorf("Failed to add src rule: %v", err)
 		}
 
+		// Only add an interface rule if there is 1 IP address configured on the interface
+		if len(ipCfgs) == 1 {
```
squeed, line 263: This rule doesn't reference `ipCfg` at all, so why is this in the `for _, ipCfg := range ipCfgs {` loop?

neilcook: That is a very good question :) I can move it out of the loop.

neilcook:
Any more thoughts on this PR?

mlguerrero12:
So, this is intended for packets originating from applications that are bound to a device. After that, a table is selected to continue with the regular routing process. Is that right?

I believe this should be a new parameter instead of applying with there is a single IP (no provided table ID case). It can take precedence over the IPs. Something like  `useOriginatingIf`, when this is set to True by the users, then a rule is added using oif. No need to add rules for IPs (please correct me if I'm wrong here). And move all associated routes of the interface to the table.

Also, why not using the VRF plugin instead?

mlguerrero12:
as far as I know, for vrf, you only need to use SO_BINDTODEVICE which I guess you are already using

PR #1151, opened 2025-03-20
Fix addresses and routes reinserted to the VRF

mlguerrero12:
Fix addresses and routes reinserted to the VRF
    
These changes make sure that only IPV6 global addresses and non local/connected routes are reinserted to the VRF table after slaving the interface

Issue:

Consistent failure

```
plugin type="vrf" failed (add): cmdAdd failed: failed getting local/host addresses for net1 in table 1 with dst fe80::2204:fff:fef1:88a1/64
```

After slaving an interface to a VRF, global IPV6 addresses (unless a sysctl option is enabled) are lost. The plugin gets the IPV6 addresses after and before slaving the interface and adds the ones that are missing. It checks that those addresses are present in the kernel to prevent failures when adding missing global routes associated with these addresses (subsequent operation). There is a retrying mechanism for checking the presence of these addresses since the kernel takes some time to add them.

First approach was to increase the number of retries (from 10 retries to 100 retries - same sleep time 10ms). It works, however, as seen in the log, the IPV6 address is a link-local one. These are moved by the kernel automatically. For some reason, when combining SRIOV and VRF, the kernel takes longer to add these. So, instead, a logic was added to only reinsert global IPV6 addresses. The issue was fixed with this without having to increment the retries. There were some routes that don't need to be added. This PR includes a fix for this as well. This latter approach is preferred instead of blindly increasing the retries.

https://docs.kernel.org/networking/vrf.html

### DIFF
--- a/plugins/meta/vrf/vrf.go
+++ b/plugins/meta/vrf/vrf.go
@@ -99,12 +99,12 @@ func addInterface(vrf *netlink.Vrf, intf string) error {
 		return fmt.Errorf("interface %s has already a master set: %s", intf, master.Attrs().Name)
 	}
 
-	// IPV6 addresses are not maintained unless
+	// Global IPV6 addresses are not maintained unless
 	// sysctl -w net.ipv6.conf.all.keep_addr_on_down=1 is called
 	// so we save it, and restore it back.
-	beforeAddresses, err := netlink.AddrList(i, netlink.FAMILY_V6)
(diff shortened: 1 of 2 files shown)


mlguerrero12:
@squeed, please take a look at this pr when you get the chance

c.c. @dougbtv

squeed (approved):

PR #1175, opened 2025-04-28
Delete vlan default GW interface

mlguerrero12:
Del command cleans up default vlan interface


squeed:
It makes sense that we should be clearing up the host-side interface, but (as we discussed synchronously), what happens if there's a delayed DEL that collides with an existing container? Is there a chance we could take another container offline?

Is there an easy way to tell if the host-side vlan interfaces is still "owned" by the container being deleted (veth peer perhaps?)

mlguerrero12:
Yes, it's not safe. I'll explore some options. Thanks

PR #1187, opened 2025-06-09
Set value of gw to nil for opt121 routes in DHCP

omartin2010:
Default behavior of the plugin is to try and install routes without using SRC which prevents routes from being installed in the routing table in (some) cloud environment that assign /32 IPs to NICs. This change enables routes to be installed without this value (and making the scope to be set to SCOPE_LINK. Another way of fixing it would be if the plugin would be refactored to use the SRC object of the struct in github.com/vishvananda/netlink which would require more work and potentially impact other plugins.


omartin2010:
Done - lmk if this is helpful.

omartin2010:
I think we should be good.

PR #1195, opened 2025-08-06
bridge-cni: add uplinkInterface to configure VLANs

boebu:
Currently the documentation for bridge-cni L2 vlan-only setup suggests to manually add vlans to the corresponding uplink interface, if outbound connectivity is required.(https://github.com/containernetworking/cni.dev/blob/512744c6eedde79c53308e785c70d90bde7baa61/content/plugins/current/main/bridge.md?plain=1#L112)

This PR enhances the bridge-cni configuration to allow autoconfiguration of the uplink interface with specified vlans. This is especially useful for setups where the bridge and uplink in configured by the underlying operating system.

### DIFF
--- a/plugins/main/bridge/bridge.go
+++ b/plugins/main/bridge/bridge.go
@@ -63,6 +63,7 @@ type NetConf struct {
 	EnableDad                 bool         `json:"enabledad,omitempty"`
 	DisableContainerInterface bool         `json:"disableContainerInterface,omitempty"`
 	PortIsolation             bool         `json:"portIsolation,omitempty"`
+	UplinkInterface           string       `json:"uplinkInterface,omitempty"`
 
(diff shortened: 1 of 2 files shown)


mlguerrero12:
Could you please provide more details about the use case? Is using an external tool to configure the uplink interface not an option?

Also, how would this change be reverted? Even with cleanup code in the CNI DEL command, we could end up in a situation similar to the L3 VLAN gateway interface issue (https://github.com/containernetworking/plugins/pull/1175).

This logic seems intended for L2 scenarios, but it's missing a check to skip this behavior for L3. It would also be good to ensure that ensureUplinkVlan is only called when the vlan or vlans parameters are actually set.

I think this should be handled externally and not by the plugin.

PR #1217, opened 2025-11-27
sbr: Add options to configure static gateway IP and preserve default routes

dagrayvid:
This PR aims to solve two problems I faced using the source-based router CNI plugin for distributed LLM inference with vLLM / llm-d. 

1. Option to configure a statically defined gateway IP to set or override the gateway if not passed in from the CNI config chain. 

In my environment (IBM Cloud), the DHCP server my secondary high-speed networks does not provide the gateway IP, so I needed a way to override / statically define the gateway IPs to ensure that the custom routing tables included the default route to the gateway. 

2. Option to "preserve default routes" in the main routing table, so that packets without a source IP are still routed correctly. For example in my environment after enabling SBR, `ping -I 10.0.0.6 10.1.0.7` would work, but `ping 10.1.0.7`. With `"preserveDefaultRoutes": true`, both work.

Tested in my environment with this CNI config:
```
{
  "cniVersion": "0.3.1",
  "name": "dhcp-host-device-port-1",
  "plugins": [
    {
      "type": "host-device",
      "device": "enp163s0",
      "isRdma": true,
      "ipam": {
        "type": "dhcp"
      }
    },
    {
      "type": "sbr-custom",
      "gateway": "10.0.0.1",
      "preserveDefaultRoutes": true
    }
  ]
}
```

In a Pod, the result of the statically defined gateway:
```
bash-5.1# ip route show table 100
default via 10.0.0.1 dev net1
10.0.0.0/16 dev net1 proto kernel scope link src 10.0.0.5
```

and the result of preserve default route:
```
bash-5.1# ip route show table 254
default via 10.130.0.1 dev eth0
10.0.0.0/16 dev net1 scope link src 10.0.0.5
10.1.0.0/16 dev net2 scope link src 10.1.0.6
10.2.0.0/16 dev net3 scope link src 10.2.0.6
...
```

### DIFF
--- a/plugins/meta/sbr/main.go
+++ b/plugins/meta/sbr/main.go
@@ -23,6 +23,7 @@ import (
 
 	"github.com/alexflint/go-filemutex"
 	"github.com/vishvananda/netlink"
+	"golang.org/x/sys/unix"
 
 	"github.com/containernetworking/cni/pkg/skel"
 	"github.com/containernetworking/cni/pkg/types"
@@ -49,6 +50,16 @@ type PluginConf struct {
 
 	// Add plugin-specific flags here
 	Table *int `json:"table,omitempty"`
+	// Gateways allows specifying static/hardcoded gateway IP addresses
+	// If set, these will be used instead of the gateway from prevResult
+	// Supports dual-stack: provide one IPv4 and/or one IPv6 gateway
+	// Note: Currently applies the same gateway to all IPs of the same family.
+	// Per-subnet gateway mapping is not yet supported.
+	Gateways []string `json:"gateways,omitempty"`
+	// AddSourceHints adds source IP hints to subnet routes in the main table,
+	// enabling destination-based routing when no explicit source IP is specified.
+	// This allows both source-based and destination-based routing to work together.
+	AddSourceHints bool `json:"addSourceHints,omitempty"`
 }
 
 // Wrapper that does a lock before and unlock after operations to serialise
@@ -168,7 +179,7 @@ func cmdAdd(args *skel.CmdArgs) error {
 		if conf.Table != nil {
 			return doRoutesWithTable(ipCfgs, *conf.Table)
 		}
-		return doRoutes(ipCfgs, args.IfName)
+		return doRoutes(ipCfgs, args.IfName, conf.Gateways, conf.AddSourceHints)
 	})
 	if err != nil {
 		return err
@@ -207,7 +218,7 @@ func getNextTableID(rules []netlink.Rule, routes []netlink.Route, candidateID in
 }
 
 // doRoutes does all the work to set up routes and rules during an add.
-func doRoutes(ipCfgs []*current.IPConfig, iface string) error {
+func doRoutes(ipCfgs []*current.IPConfig, iface string, staticGateways []string, addSourceHints bool) error {
 	// Get a list of rules and routes ready.
 	rules, err := netlinksafe.RuleList(netlink.FAMILY_ALL)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


plugins/meta/sbr/main.go
```diff
@@ -49,6 +50,14 @@ type PluginConf struct {
 
 	// Add plugin-specific flags here
 	Table *int `json:"table,omitempty"`
+	// Gateway allows specifying static/hardcoded gateway IP addresses
+	// If set, these will be used instead of the gateway from prevResult
+	// Supports dual-stack: provide one IPv4 and/or one IPv6 gateway
+	Gateway []string `json:"gateway,omitempty"`
```
LionelJouin:
```suggestion
	Gateways []string `json:"gateways,omitempty"`
```

plugins/meta/sbr/main.go
```diff
@@ -49,6 +50,16 @@ type PluginConf struct {
...
+	// Gateways allows specifying static/hardcoded gateway IP addresses
+	// If set, these will be used instead of the gateway from prevResult
+	// Supports dual-stack: provide one IPv4 and/or one IPv6 gateway
+	// Note: Currently applies the same gateway to all IPs of the same family.
+	// Per-subnet gateway mapping is not yet supported.
+	Gateways []string `json:"gateways,omitempty"`
+	// PreserveDefaultRoutes keeps subnet routes in the main routing table
+	// with proper source IP hints for destination-based routing.
+	// This allows packets without an explicit source IP to be routed correctly.
+	PreserveDefaultRoutes bool `json:"preserveDefaultRoutes,omitempty"`
```
karampok: Preserve"DefaultRoutes" make me thinks default router (route add default via ).  just `PreserveRoutes` is cleaner

karampok (commented):
Some comments from my side

-  5 commits seem too many imo, maybe 2 (one for static gateway on for preserveDefaultRoutes) or even 1 (like atomic commits)
-  test look a bit thin, like only ipv4 not unhappy path

(Note: I am not maintainer of the repo)

squeed (approved):

squeed (changes requested):
Can you please squash your commits? And a v6 test would be good.

plugins/meta/sbr/main.go
dagrayvid: I think this is a good point. I renamed this option to `AddSourceHints`. Do you think that is more clear?

plugins/meta/sbr/main.go
```diff
@@ -238,6 +249,22 @@ func doRoutes(ipCfgs []*current.IPConfig, iface string) error {
 		return fmt.Errorf("Unable to list routes: %v", err)
 	}
 
+	// Parse static gateways if provided (supports dual-stack)
+	var staticGwV4, staticGwV6 net.IP
+	for _, gw := range staticGateways {
+		ip := net.ParseIP(gw)
```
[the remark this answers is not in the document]
dagrayvid: Yes, I will leave this as is for now because netlink is still using net.ParseIP.

karampok (commented):
one nit comment, but lgtm in general

karampok (approved):

dagrayvid:
Just following up on this PR
@squeed

plugins/meta/sbr/main.go
```diff
@@ -49,6 +50,13 @@ type PluginConf struct {
 
 	// Add plugin-specific flags here
 	Table *int `json:"table,omitempty"`
+	// Gateway allows specifying a static/hardcoded gateway IP address
+	// If set, this will be used instead of the gateway from prevResult
+	Gateway string `json:"gateway,omitempty"`
```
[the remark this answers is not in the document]
squeed: Having 2 gateways seems reasonable for now. Given that we're not the primary user of this plugin, I'll defer that decision to you. I'll resolve this, but now you know potential room for expansion.

squeed (approved):

Issue #1221, opened 2025-12-14
Segfault possibly because of invalid or nil bandwidth configuration


adelton:
In https://github.com/k3s-io/k3s/issues/13344 I've described a failure including SIGSEGV when using K3s with its CNI plugins with CRI-O, as opposed to containerd.

@brandond helped to unpack the stacktrace to
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x5f89cc]

goroutine 1 gp=0xc000002380 m=0 mp=0x8da300 [running, locked to thread]:
panic({0x6763e0?, 0x8cd930?})
	/usr/local/go/src/runtime/panic.go:811 +0x168 fp=0xc0000d3950 sp=0xc0000d38a0 pc=0x46e668
runtime.panicmem(...)
	/usr/local/go/src/runtime/panic.go:262
runtime.sigpanic()
	/usr/local/go/src/runtime/signal_unix.go:925 +0x359 fp=0xc0000d39b0 sp=0xc0000d3950 pc=0x4705d9
github.com/containernetworking/plugins/plugins/meta/bandwidth.cmdCheck(0xc000096980)
	/tmp/tmp.SfBAFPHij0/src/github.com/containernetworking/plugins/plugins/meta/bandwidth/main.go:302 +0x2ac fp=0xc0000d3b18 sp=0xc0000d39b0 pc=0x5f89cc
github.com/containernetworking/plugins/vendor/github.com/containernetworking/cni/pkg/skel.(*dispatcher).checkVersionAndCall(0xc0000a9db0, 0xc000096980, {0x72b288, 0xc000079d70}, 0x6de140)
	/tmp/tmp.SfBAFPHij0/src/github.com/containernetworking/plugins/vendor/github.com/containernetworking/cni/pkg/skel/skel.go:204 +0x116 fp=0xc0000d3bc0 sp=0xc0000d3b18 pc=0x576856
github.com/containernetworking/plugins/vendor/github.com/containernetworking/cni/pkg/skel.(*dispatcher).pluginMain(0xc0000a9db0, {0x6de138, 0x6de148, 0x6de140, 0x0, 0x0}, {0x72b288, 0xc000079d70}, {0xc0000924e0, 0x20})
[...]

PR #1222, opened 2025-12-15
bandwidth: handle nil bandwidth in CHECK

squeed:
Somehow we missed this case; if CHECK is called with no bandwidth configuration, we segfault. Oops.

Fixes: #1221


LionelJouin (approved):

squeed:
@brandond any chance you could verify in your environment that this fixes the issue before we cut a release?

adelton:
What is the canonical way to build from master, ideally in some containerized fashion so that the environment is the same as the one used to build binaries for releases?

brandond:
Fantastic. @squeed could we get v1.9.1?

PR #1235, opened 2026-02-26
fix(portmap): correct masquerading source address handling

l1b0k:
This commit fixes the portmap plugin's masquerading source address handling in the nftables backend.  #1234


BSWANG (approved):

squeed:
@danwinship would you mind giving this a sanity check?

squeed (approved):
Nice fix, thanks.

PR #1237, opened 2026-03-08
Simplify some code with slices and for-range.

atombrella:
I applied a few of the modernize rules to simplify the code slightly:

* [slices.Contains](https://pkg.go.dev/golang.org/x/tools/go/analysis/passes/modernize#hdr-Analyzer_slicescontains) removes some chunks of iterations that do exactly what can be leveraged to `slices.Contains`
* [for range N](https://pkg.go.dev/golang.org/x/tools/go/analysis/passes/modernize#hdr-Analyzer_slicescontains) which is syntactic sugar.
* A single `// +build` which definitely needs to go.
* `slices.Sort` instead of `sort.Slice`.

The golangci-lint version is a bit old, and I didn't want to tamper around with it in this PR.

### DIFF
--- a/pkg/ip/ipmasq_iptables_linux.go
+++ b/pkg/ip/ipmasq_iptables_linux.go
@@ -18,6 +18,7 @@ import (
 	"errors"
 	"fmt"
 	"net"
+	"slices"
 	"strings"
 
 	"github.com/coreos/go-iptables/iptables"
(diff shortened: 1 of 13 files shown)


squeed:
@atombrella this looks good to me. Can you rebase without a merge commit and we can merge this?

atombrella:
> @atombrella this looks good to me. Can you rebase without a merge commit and we can merge this?

@squeed Rebased :smile:

PR #1247, opened 2026-04-09
bridge: add group_fwd_mask support for multicast forwarding (PTP)

VaishnavSreekumar:
Adds support for configuring the group_fwd_mask attribute in the bridge CNI plugin. This allows the Linux bridge to forward specific link-local multicast frames (e.g., LLDP, PTP, or LACP) that are typically dropped by default.

### Why is this needed?
Currently, the Linux kernel allows overriding default bridge behavior via /sys/class/net/<bridge>/bridge/group_fwd_mask, but the CNI bridge plugin lacks an interface to expose this configuration. This is a blocker for high-precision networking environments, specifically:

PTP (IEEE 1588): Used for sub-microsecond clock synchronization.

Telecom Profiles: Environments utilizing G.8275.1 that require specific multicast forwarding.

## Changes

Schema Update: Added GroupFwdMask as an optional integer field to the NetConf struct.

Sysfs Integration: Implemented logic to write the user-defined mask to the bridge's sysfs path during setup.

Validation: Included checks to ensure the value is only applied when explicitly set, maintaining default kernel behavior otherwise.

## Proposed Configuration
Users can now specify the mask in their CNI configuration file:

JSON
```
{
  "type": "bridge",
  "name": "mynet",
  "bridge": "cni0",
  "groupFwdMask": 16384,
  "ipam": {
    "type": "host-local",
    "subnet": "10.10.0.0/16"
  }
}
```
## Technical Details

Backward Compatibility: Fully backward compatible; if the field is omitted, no changes are made to the bridge mask.

Target Version: Intended for integration into upcoming Kubernetes network driver releases.

### Testing Done
Verified that the group_fwd_mask file in /sys is correctly updated when the CNI is invoked.

Confirmed that PTP traffic (multicast MAC 01-80-C2-00-00-0E) is forwarded when the mask is set appropriately.

## Release note

```release-note
bridge: add group_fwd_mask configuration to control multicast forwarding behavior
```

### DIFF
--- a/plugins/meta/portmap/portmap_nftables.go
+++ b/plugins/meta/portmap/portmap_nftables.go
@@ -228,16 +228,32 @@ func (pmNFT *portMapperNFTables) forwardPorts(config *PortMapConf, containerNet 
 		// In theory we should validate that the original dst IP and port are as
 		// expected, but *any* traffic matching one of these patterns would need
 		// to be masqueraded to be able to work correctly anyway.
+
+		var masqSrcAddr string
+		if config.MasqAll {
+			// MasqAll: match traffic from any source IP
+			if isV6 {
+				masqSrcAddr = "::/0"
+			} else {
+				masqSrcAddr = "0.0.0.0/0"
+			}
+		} else {
+			// Default: only match traffic from container's own IP (hairpin)
+			masqSrcAddr = containerNet.IP.String()
+		}
+
 		tx.Add(&knftables.Rule{
 			Chain: masqueradingChain,
 			Rule: knftables.Concat(
-				ipX, "saddr", containerNet.IP,
+				ipX, "saddr", masqSrcAddr,
 				ipX, "daddr", containerNet.IP,
 				"masquerade",
 			),
 			Comment: &config.ContainerID,
 		})
-		if !isV6 {
+		if !isV6 && !config.MasqAll {
+			// Only add localhost rule when MasqAll is false
+			// (when MasqAll is true, 0.0.0.0/0 already covers 127.0.0.1)
 			tx.Add(&knftables.Rule{
 				Chain: masqueradingChain,
 				Rule: knftables.Concat(
@@ -275,8 +291,12 @@ func (pmNFT *portMapperNFTables) checkPorts(config *PortMapConf, containerNet ne
 	}
 	if *config.SNAT {
 		masqueradings = 1
-		if !isV6 {
-			masqueradings *= 2
+		// When MasqAll is false and IPv4, we have 2 rules:
+		// 1. hairpin rule (container IP -> container IP)
+		// 2. localhost rule (127.0.0.1 -> container IP)
+		// When MasqAll is true, we only have 1 rule (0.0.0.0/0 -> container IP)
+		if !isV6 && !config.MasqAll {
+			masqueradings = 2
 		}
 	}
 
--- a/plugins/meta/portmap/portmap_nftables_test.go
+++ b/plugins/meta/portmap/portmap_nftables_test.go
@@ -224,5 +224,253 @@ add rule ip6 cni_hostport prerouting c d fib daddr type local jump hostports_all
 				Expect(err).To(HaveOccurred())
 			})
 		})
+
+		Describe("MasqAll configuration", func() {
+			var pmNFT *portMapperNFTables
+			var ipv4Fake, ipv6Fake *knftables.Fake
+			BeforeEach(func() {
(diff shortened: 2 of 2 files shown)


s1061123:
Thank you so much for the PR! We also need document modification because we maintain the document side as well, which is in https://github.com/containernetworking/cni.dev

Could you please create doc PR as well?

.gitignore
```diff
@@ -22,7 +22,8 @@ _testmain.go
 *.exe
 *.test
 *.prof
-
+/bridge
```
squeed: This seems unnecessary? You can set `.git/info/exclude` for local files ([doc](https://nesbitt.io/2026/02/05/git-magic-files.html#gitignore))

plugins/main/bridge/bridge.go
```diff
@@ -513,11 +514,26 @@ func calcGatewayIP(ipn *net.IPNet) net.IP {
...
 	}
+	// groupFwdMask enables forwarding of link-local multicast MAC addresses.
+	// By default, Linux bridge drops addresses like 01-80-C2-00-00-0E (PTP).
+	// When set, this writes to /sys/class/net/<bridge>/bridge/group_fwd_mask.
+	if n.GroupFwdMask != 0 {
+		if n.GroupFwdMask < 0 || n.GroupFwdMask > 65535 {
+			return nil, nil, fmt.Errorf("groupFwdMask must be between 0 and 65535")
+		}
+
+		path := fmt.Sprintf("/sys/class/net/%s/bridge/group_fwd_mask", n.BrName)
```
squeed: Is `n.BrName` sanitized here? Could it contain `..`? That would be... bad.

squeed (changes requested):
Could you add tests? Otherwise, seems reasonable.

squeed:
Is this useful for other kinds of interfaces? Should we also add this setting to the tuning plugin?

VaishnavSreekumar:
I have 
* Added input validation via helper function
* Improved error message with bridge context
* Added comment clarifying default (0) behavior

Also added a unit test for validation logic
I am working on the corresponding documentation PR in `cni.dev` and will link it here shortly
Please let me know if you'd like additional tests around bridge setup behavior

VaishnavSreekumar:
Added documentation PR
This documents the groupFwdMask configuration option for the bridge plugin, including its usage and behavior for enabling forwarding of link-local multicast traffic
[(https://github.com/containernetworking/cni.dev/pull/152)]

Please let me know if those matches what u had in mind

VaishnavSreekumar:
> Is this useful for other kinds of interfaces? Should we also add this setting to the tuning plugin?

That’s a good point.

My understanding is that group_fwd_mask is specific to Linux bridge interfaces, which is why I implemented it in the bridge plugin. For other interface types, this setting wudnt apply in the same way

with that  if there are scenario where similar control is useful across different interfaces extending it via the tuning plugin could make sense as a followup enhancement.

Happy to explore that separately if needed.

plugins/main/bridge/bridge.go
```diff
@@ -513,11 +515,28 @@ func calcGatewayIP(ipn *net.IPNet) net.IP {
...
+	// By default, Linux bridge drops addresses like 01-80-C2-00-00-0E (PTP).
+	// When set, writes the value to /sys/class/net/<bridge>/bridge/group_fwd_mask.
+
+	// Skip default (0) to preserve kernel behavior unless explicitly set
+	if n.GroupFwdMask != 0 {
+		if err := validateGroupFwdMask(n.GroupFwdMask); err != nil {
+			return nil, nil, err
+		}
+
+		safeName := filepath.Base(n.BrName)
```
squeed: Ah, good idea. Let's also exclude the name of `..` just to be paranoid.

squeed (changes requested):
It looks like you checked in some extra files. Please remove them from the commit.

plugins/main/bridge/groupfwdmask_test.go
```diff
@@ -0,0 +1,26 @@
...
+		wantErr bool
+	}{
+		{"valid", 16384, false},
+		{"zero", 0, false},
+		{"negative", -1, true},
+		{"too large", 70000, true},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
```
mlguerrero12: I think it's better to add tests in `bridge_test.go` following the existing pattern. You can verify that an add command with this config succeeds, or fails when the value is invalid.

mlguerrero12 (changes requested):

Issue #1249, opened 2026-04-19
Support ECMP route configuration in Static IPAM plugin


yushoyamaguchi:
# Background
The current CNI Static IPAM plugin does not explicitly define or document support for ECMP (Equal-Cost Multi-Path) routing when multiple routes with the same destination prefix are provided.
I would like to clarify and formally support ECMP route configuration as part of the plugin behavior.

# Proposal
I think two possible approaches:
## Option 1: Implicit ECMP via multiple routes
Allow users to define multiple routes with the same destination prefix but different gateways, and install them as ECMP routes inside the Pod.
```
      "routes": [
        { "dst": "0.0.0.0/0", "gw": "10.0.0.1" },
        { "dst": "0.0.0.0/0", "gw": "10.0.0.2" }
      ]
```
Note:
- This behavior already works for IPv6 in practice. (ref: https://github.com/torvalds/linux/blob/v6.19/net/ipv6/ip6_fib.c#L1152-L1162)
- We propose to extend and standardize this behavior for IPv4 as well.
- I think this should be clearly documented as supported behavior.

## Option 2: Explicit ECMP syntax
Introduce a dedicated syntax in the `routes` field to explicitly define ECMP routes.
This could improve clarity and avoid ambiguity in interpretation.
However, this approach would require extending the current API/schema.

# Use case
Our use case is as follows:
- In an OpenStack environment, we create multiple routers connected to the same L2 network.
- This network is attached to worker VMs.
- Inside the VM, we attach an additional interface to Pods using ipvlan (as a secondary interface).
- For this interface, we want to configure ECMP routes where multiple OpenStack routers act as gateways.
This requires the ability to define multiple equal-cost default routes in the CNI configuration.

<img width="807" height="609" alt="Image" src="https://github.com/user-attachments/assets/6ff3559b-7789-4aa3-94af-2fe82c88c01a" />

yushoyamaguchi:
https://github.com/containernetworking/plugins/pull/1248
Previously, I implemented a test assuming that ECMP would be achieved with a configuration like the following:
```
      "routes": [
        { "dst": "0.0.0.0/0", "gw": "10.0.0.1" },
        { "dst": "0.0.0.0/0", "gw": "10.0.0.2" }
      ]
```

However, after further verification, I found that for IPv4 this does not actually result in ECMP routes being installed.
```
default via 10.0.0.1 dev eth1 
default via 10.0.0.2 dev eth1
```
[...]

PR #1251, opened 2026-04-21
Revert "ipam: test ECMP support"

yushoyamaguchi:
This reverts commit c1069bdc27cc375027f90567200d4543a80f4fb6.

ref: https://github.com/containernetworking/plugins/issues/1249#issuecomment-4276254035

I misunderstood the conditions required for ECMP to be established, and the criteria used in this test case to determine whether ECMP was established were also incorrect.


s1061123 (approved):
Looks good to me.

PR #1257, opened 2026-05-04
host-device: copy host interface IP addresses and routes into container

SchSeba:
Add a new configuration option `useInterfaceNetwork` that instructs the host-device plugin to capture the interface's IP addresses and routes from the host before moving the device into the container namespace, and then apply them inside the container.

This is critical for virtual environments (AWS, IBM Cloud, GPC) where the cloud provider configures IP addresses and routes directly on the network device. In these environments, there is no traditional IPAM source; the ground truth for L3 configuration lives on the host interface itself.

When `useInterfaceNetwork` is enabled, the plugin:
  - Captures all global-scope addresses and non-local routes from the host device before moving it into the container namespace.
  - Applies the captured addresses and routes to the interface inside the container.
  - Reports the addresses and routes in the CNI result (merged with any IPAM result if an IPAM plugin is also configured).

NOTE: The interface configuration on the host node must be persistent. When the device is moved back to the host (via DEL) and renamed to its original name, the system's network management service (e.g. NetworkManager, systemd-networkd, cloud-init, or cloud-specific agents) is expected to detect the device and re-apply the IP addresses and routes. This plugin does NOT re-configure the host interface on DEL; it relies on the node's network configuration being declarative and reconciled by the platform's networking stack.

Also implements the STATUS command to verify the host device exists, replacing the previous TODO stub.


SchSeba:
Hi @s1061123 @squeed @LionelJouin if you have time please take a look on the PR.
This is critical for us to support virtual clusters running on clouds where the VFs are pass into the cluster VMs nodes with network configuration.

plugins/main/host-device/host-network-state.go
```diff
@@ -0,0 +1,228 @@
...
+	"github.com/containernetworking/plugins/pkg/netlinksafe"
+	"github.com/containernetworking/plugins/pkg/ns"
+)
+
+const (
+	// localRouteTable is the Linux "local" routing table.
+	localRouteTable = 255
+)
+
+// HostNetworkStateFile holds the captured host-side L3 configuration
```
karampok: personal preference no comment unless exported

plugins/main/host-device/host-network-state.go
```diff
@@ -0,0 +1,228 @@
...
+// (addresses, routes, and rules) that should be applied to the container interface.
+type HostNetworkStateFile struct {
```
karampok: prefix ^File$ not very nice not really a file. May InterfaceInfo, InterfaceConfig, or just Interface

plugins/main/host-device/host-network-state.go
```diff
@@ -0,0 +1,228 @@
...
+	HostIfName    string       `json:"hostIfName"`
+	HostLinkWasUp bool         `json:"hostLinkWasUp"`
+	Addresses     []string     `json:"addresses,omitempty"`
```
karampok: can we use netlink.Addr, netlink.routes, rule

plugins/main/host-device/host-network-state.go
```diff
@@ -0,0 +1,228 @@
...
+			Priority: rule.Priority,
+		})
+	}
+
+	return nil
+}
+
+// applyNetworkStateToPod applies captured state to the moved interface inside the pod namespace.
+func applyNetworkStateToPod(containerNs ns.NetNS, contDev netlink.Link, state *HostNetworkStateFile) error {
+	if state == nil {
```
karampok, line 130: having that  check, (imo) indicates that this function maybe should be method to the struct.

plugins/main/host-device/host-device.go
```diff
@@ -496,6 +528,82 @@ func printLink(dev netlink.Link, cniVersion string, containerNs ns.NetNS) error
...
+		CNIVersion: current.ImplementedSpecVersion,
+		Interfaces: []*current.Interface{
+			{
+				Name:    dev.Attrs().Name,
+				Mac:     dev.Attrs().HardwareAddr.String(),
+				Sandbox: containerNs.Path(),
+			},
+		},
+	}
+	mergeNetworkStateIntoResult(result, state)
```
karampok, line 555: imo if user decides to keep network config from the host then we should ignore IPAM or block the combination witt IPAM in the config. I think is either IPAM, or host network config (no ip is also valid config)

plugins/main/host-device/host-network-state_test.go
```diff
@@ -0,0 +1,205 @@
...
+import (
+	"encoding/json"
+	"net"
+	"testing"
+
+	current "github.com/containernetworking/cni/pkg/types/100"
+)
+
+// TestUseInterfaceNetwork verifies useInterfaceNetwork boolean behavior.
+func TestUseInterfaceNetwork(t *testing.T) {
```
karampok: what exactly this test is doing?

plugins/main/host-device/host-network-state_test.go
```diff
@@ -0,0 +1,205 @@
...
+	if useInterfaceNetwork(&NetConf{}) {
+		t.Fatalf("expected default config to disable interface network copy")
+	}
+	if !useInterfaceNetwork(&NetConf{UseInterfaceNetwork: true}) {
+		t.Fatalf("expected useInterfaceNetwork=true to enable copy")
+	}
+}
+
+// TestStateJSONHasNoNeighbors verifies state serialization excludes neighbors.
+func TestStateJSONHasNoNeighbors(t *testing.T) {
```
karampok: what exactly this test is doing? Not sure the tests in the file really add value

plugins/main/host-device/host-network-state.go
SchSeba: Done - removed doc comments from all unexported symbols.

plugins/main/host-device/host-network-state.go
SchSeba: Good call - renamed `HostNetworkStateFile` to `HostNetworkState` throughout.

plugins/main/host-device/host-network-state.go
SchSeba: The string-based representation is intentional here - `netlink.Addr`, `netlink.Route` and `netlink.Rule` contain `net.IP` / `*net.IPNet` fields that don't round-trip cleanly through JSON (`net.IP` marshals as a base64 byte array, `*net.IPNet` isn't directly marshalable). Using strings gives us portable, human-readable JSON and avoids coupling the serialization format to the netlink library's internal types.

We do convert back to netlink types when applying (`applyOnLink`), so the actual netlink interaction is the same.

plugins/main/host-device/host-network-state.go
SchSeba: Agreed - converted `applyNetworkStateToPod` and `applyNetworkStateOnLink` to methods on `*HostNetworkState` (`applyToPod` and `applyOnLink`). The nil-receiver check now reads more naturally.

plugins/main/host-device/host-device.go
SchSeba: I see the point, but there are valid use cases for the combination: the host interface provides the base L3 config (addresses/routes from the cloud provider), and IPAM adds additional addresses on top (e.g. secondary IPs, service IPs). Blocking the combination would reduce flexibility for users who need both.

The current merge behavior is additive - IPAM addresses are appended alongside host-captured ones. If you feel strongly, we could add a validation warning instead of an error, but I'd prefer to keep the flexibility. WDYT?

plugins/main/host-device/host-network-state_test.go
SchSeba: Removed this test - it was only validating the trivial boolean guard function which is already covered implicitly by the integration tests.

plugins/main/host-device/host-network-state_test.go
SchSeba: Removed this test as well. Kept the `TestMergeNetworkState*` and `TestLoadConf*` tests since those exercise actual logic (result merging, config parsing, DPDK rejection).

plugins/main/host-device/host-device.go
```diff
@@ -57,6 +57,7 @@ type NetConf struct {
 	RuntimeConfig struct {
 		DeviceID string `json:"deviceID,omitempty"`
 	} `json:"runtimeConfig,omitempty"`
+	UseInterfaceNetwork bool `json:"useInterfaceNetwork,omitempty"`
```
s1061123, line 64: Recommend to add comment to quickly mention what is 'UseInterfaceNetwork' because the option name is not intuitive (what 'useInterfaceNetwork' is used?)

s1061123 (dismissed):
This PR introduces new parameter, 'useInterfaceNetwork', so could you please create another PR in https://github.com/containernetworking/cni.dev/pulls to modify host-device CNI document as well?

plugins/main/host-device/host-device.go
SchSeba: Done - added an inline comment: `// When true, copy the host interface's IP addresses and routes into the container before IPAM runs.`

plugins/main/host-device/host-device.go
karampok: What about conflicting default routes? IPs on the same prefix?

karampok:
I am a bit unsure if we capture everything that need to be re-applied or if capture something we should not.
For example, when IPv6, when SLAAC, routes from RA, nothing should be copied imo. (IPv6 test are missing)

Some AI generated list which seems possible
```
1. Neighbours not captured — GCP assigns /32 to NICs; all egress needs ARP entry for gateway; LinkSetNsFd wipes neighbour table; traffic blackholes.
  proof
  2. rp_filter not captured — AWS multi-ENI requires loose (2); container gets netns default strict (1); asymmetric return traffic silently dropped.
  proof
  3. RTPROT_RA routes copied — RA routes have expiry (expires Nsec); re-applied without lifetime; no RA daemon in container to refresh; stale routes
  persist forever. proof
  4. SLAAC addresses copied without IFA_F_PERMANENT filter — dynamic addrs (ip addr show → dynamic) become permanent in container; no renewal; never
  expire. [proof](/usr/include/linux/if_addr.h:54 — IFA_F_PERMANENT 0x80]

```

SchSeba:
Thanks for the thorough review @karampok - addressing your IPv6/SLAAC concerns:

**Addressed in the latest push:**
- **RTPROT_RA routes** — now filtered out in `captureHostNetworkState`. RA-learned routes carry kernel-managed lifetimes and there's no RA daemon inside the container to refresh them, so copying them would create stale permanent routes.
- **SLAAC addresses** — now filtered via `IFA_F_PERMANENT` flag check. Only permanently configured addresses are captured; dynamic (SLAAC/DHCPv6) addresses are skipped since they won't be renewed inside the container.

**Out of scope for this PR (can be follow-ups):**
- **Neighbours** — On GCP with /32 NICs, the gateway ARP entry is indeed wiped when the device moves namespaces. However, once traffic flows in the container, the kernel will re-do ARP resolution (the gateway is reachable via the copied routes). If specific cloud setups require pre-populated neighbour entries, that can be added as a follow-up with a new `captureNeighbours` option. Adding it here would increase scope and complexity significantly.
- **rp_filter** — This is a per-interface sysctl, not part of the netlink L3 configuration. Managing sysctls is the responsibility of the [tuning CNI plugin](https://www.cni.dev/plugins/current/meta/tuning/), which is the standard approach in CNI. Users on AWS multi-ENI setups can chain `tuning` after `host-device` to set `rp_filter=2`.

Also added IPv6-related filtering (link-local unicast routes were already skipped).

plugins/main/host-device/host-device.go
SchSeba: Good question. Both cases are handled gracefully:

- **Conflicting IPs on the same prefix**: `applyOnLink` wraps every `AddrAdd` / `RouteAdd` call with `isAlreadyExistsErr` - if the host-captured address or route already exists (because IPAM configured the same one), the duplicate is silently ignored.
- **Conflicting default routes**: Linux allows multiple default routes with different metrics (ECMP). If both the host config and IPAM supply a default route with the *same* gateway+metric, the second `RouteAdd` returns `EEXIST` which we handle. If they differ in metric, both coexist as expected.

The order of operations is: host addresses/routes are applied first (in `applyToPod`), then IPAM runs and calls `ConfigureIface`. So IPAM is always the "last writer" for identical entries, and duplicates are harmlessly skipped.

In practice, the combination is intended for cases like: host provides the primary IP/routes from the cloud provider, and IPAM adds secondary IPs or service-specific routes on top - they typically won't conflict.

plugins/main/host-device/host-device.go
SchSeba: I also open the PR for the documentation site https://github.com/containernetworking/cni.dev/pull/156

plugins/main/host-device/host-device.go
```diff
@@ -138,12 +149,24 @@ func cmdAdd(args *skel.CmdArgs) error {
...
 
 	var contDev netlink.Link
+	var networkState *HostNetworkState
 	if !cfg.DPDKMode {
 		hostDev, err := getLink(cfg.Device, cfg.HWAddr, cfg.KernelPath, cfg.PCIAddr, cfg.auxDevice)
 		if err != nil {
 			return fmt.Errorf("failed to find host device: %v", err)
 		}
 
+		networkState = &HostNetworkState{
```
karampok: `HostNetworkState` is partially initialized here from `hostDev` (name, link-up flag), then `hostDev` is passed again to `captureHostNetworkState` which extracts more fields from the same object. Consider a single constructor (e.g. `newHostNetworkState(hostDev, interfaceNetworkEnabled)`) that does all initialization in one place.

plugins/main/host-device/host-device.go
```diff
@@ -153,6 +176,15 @@ func cmdAdd(args *skel.CmdArgs) error {
 		result.Interfaces[0].Name = contDev.Attrs().Name
 		// Set the MAC address of the interface
 		result.Interfaces[0].Mac = contDev.Attrs().HardwareAddr.String()
+
+		if interfaceNetworkEnabled {
+			if err := networkState.applyToPod(containerNs, contDev); err != nil {
+				return err
+			}
+			if cfg.IPAM.Type == "" {
+				return printLinkWithNetworkState(contDev, cfg.CNIVersion, containerNs, networkState)
```
karampok: This early return for the no-IPAM + `useInterfaceNetwork` case is buried inside the `interfaceNetworkEnabled` block. Consider consolidating all no-IPAM exit paths together:
```go
if cfg.IPAM.Type == "" {
    if cfg.DPDKMode {
        return types.PrintResult(result, cfg.CNIVersion)
    }
    if interfaceNetworkEnabled {
        return printLinkWithNetworkState(contDev, cfg.CNIVersion, containerNs, networkState)
    }
    return printLink(contDev, cfg.CNIVersion, containerNs)
}
```
The control flow reads top-down: no IPAM? pick the right printer.

plugins/main/host-device/host-device.go
```diff
@@ -496,6 +532,76 @@ func printLink(dev netlink.Link, cniVersion string, containerNs ns.NetNS) error
 	return types.PrintResult(&result, cniVersion)
 }
 
+func routeStateToCNIRoute(route routeState) *types.Route {
```
karampok: `routeStateToCNIRoute` belongs in `host-network-state.go` alongside the `routeState` type definition and the rest of the state handling logic.

plugins/main/host-device/host-network-state.go
```diff
@@ -0,0 +1,231 @@
...
+// (addresses, routes, and rules) that should be applied to the container interface.
+type HostNetworkState struct {
+	HostIfName    string       `json:"hostIfName"`
+	HostLinkWasUp bool         `json:"hostLinkWasUp"`
+	Addresses     []string     `json:"addresses,omitempty"`
+	Routes        []routeState `json:"routes,omitempty"`
+	Rules         []ruleState  `json:"rules,omitempty"`
+}
+
+type routeState struct {
```
karampok: Why is `routeState` needed as an intermediate type instead of using [`types.Route`](https://github.com/containernetworking/cni/blob/v1.3.0/pkg/types/types.go#L180) directly?

The `HostNetworkState` is never serialized to JSON -- it is captured in `cmdAdd`, applied in the same call, and discarded. And `routeStateToCNIRoute` converts back to `types.Route` for printing anyway.

The only field `routeState` has that `types.Route` lacks is `Source`. That could be handled with a thin wrapper or by keeping `netlink.Route` directly for the capture/apply path.

Also, the field naming (`Destination`/`Gateway`/`Metric`) diverges from both CNI (`Dst`/`GW`/`Priority`) and netlink (`Dst`/`Gw`/`Priority`) conventions.

plugins/main/host-device/host-network-state.go
```diff
@@ -0,0 +1,231 @@
...
+		return fmt.Errorf("failed to list routes for host device %s: %w", hostDev.Attrs().Name, err)
+	}
+	for _, route := range routes {
+		if route.Table == localRouteTable {
+			continue
+		}
+		if route.Protocol == syscall.RTPROT_KERNEL {
+			continue
+		}
+		// Skip RA-learned routes: they carry a kernel-managed lifetime and
```
karampok: The comment is incorrect. There is no "RA daemon" on the receiving side -- the kernel processes Router Advertisements natively via the `accept_ra` sysctl.

When the interface moves into the container namespace, the router on the link continues sending RAs. If `accept_ra` is enabled in the container netns (kernel default), the kernel will install fresh RA routes with proper lifetimes automatically.

Skipping RA routes is correct, but for the opposite reason: you skip them because the container **will** get new ones with proper lifetimes, not because it won't.

plugins/main/host-device/host-network-state.go
```diff
@@ -0,0 +1,231 @@
...
+		ruleErr := netlink.RuleAdd(nlRule)
+		if ruleErr != nil && !isAlreadyExistsErr(ruleErr) {
+			return fmt.Errorf("failed to add copied rule (src=%s table=%d): %w", rule.Source, rule.Table, ruleErr)
+		}
+	}
+
+	return nil
+}
+
+func isAlreadyExistsErr(err error) bool {
```
karampok: The string matching fallback in `isAlreadyExistsErr` is unnecessary. `errors.Is(err, syscall.EEXIST)` is sufficient for netlink operations (`vishvananda/netlink` returns `syscall.Errno` values). The `"object already exists"` string does not come from any standard errno path.

Consider the `controller-runtime` `IgnoreAlreadyExists` pattern:
```go
func ignoreExists(err error) error {
    if errors.Is(err, syscall.EEXIST) {
        return nil
    }
    return err
}
```
Then: `return ignoreExists(netlink.AddrAdd(link, &addr))` -- cleaner than the bool check and branch.

plugins/main/host-device/host-network-state.go
karampok: I still do not think we need serialization or string

karampok (commented):
About the DEL and not restoring: this means that once we have pod using this interface and then delete the pod, the secondary interface remains unusable until node reboots. right?

SchSeba:
Hi @karampok thanks for the comments!
about
> About the DEL and not restoring: this means that once we have pod using this interface and then delete the pod, the secondary interface remains unusable until node reboots. right?

no when the interface is back on the host network namespace networkManager for example sees the device and run the configuration again on the device it can be static,dhcp or what every is convigured in the device

plugins/main/host-device/host-network-state.go
```diff
@@ -0,0 +1,202 @@
...
+	})
+	for _, route := range orderedRoutes {
+		route.LinkIndex = linkIndex
+		if err := ignoreExists(netlink.RouteAdd(&route)); err != nil {
+			return fmt.Errorf("failed to add copied route %v on %s: %w", route, link.Attrs().Name, err)
+		}
+	}
+
+	for _, rule := range state.Rules {
+		nlRule := netlink.NewRule()
```
karampok: `state.Rules` is `[]netlink.Rule` and `nlRule` is also `netlink.Rule` -- creating a new rule and copying fields from the same type is redundant leftover from the old `ruleState` string type. Simplify:
```go
for _, rule := range state.Rules {
    if err := ignoreExists(netlink.RuleAdd(&rule)); err != nil {
        return fmt.Errorf("failed to add copied rule (src=%v table=%d): %w", rule.Src, rule.Table, err)
    }
}
```

plugins/main/host-device/host-network-state_test.go
```diff
@@ -0,0 +1,175 @@
...
+	if err != nil {
+		t.Fatalf("loadConf should succeed: %v", err)
+	}
+	cfg.DPDKMode = true
+	if !useInterfaceNetwork(cfg) || !cfg.DPDKMode {
+		t.Fatalf("expected both useInterfaceNetwork and DPDKMode to be true for this test")
+	}
+}
+
+func TestMergeNetworkStateIntoResult(t *testing.T) {
```
karampok, line 65: Nit: `TestMergeNetworkStateIntoResult`, `TestMergeNetworkStateIntoResultRoutesOnly`, and `TestMergeNetworkStateNilState` share the same structure with different inputs -- could be a single table-driven test.

karampok (commented):
Small commments but this PR look lgtm from my side

karampok:
> Hi @karampok thanks for the comments! about
> 
> > About the DEL and not restoring: this means that once we have pod using this interface and then delete the pod, the secondary interface remains unusable until node reboots. right?
> 
> no when the interface is back on the host network namespace networkManager for example sees the device and run the configuration again on the device it can be static,dhcp or what every is convigured in the device

you assume they are all using NM (and auto connect feature) which probably is not true. I would have thought there is a cloudinit script that does once on boot the network config but I have not worked with clouds for many years, I do not have strong opinion here.

SchSeba:
> > Hi @karampok thanks for the comments! about
> > > About the DEL and not restoring: this means that once we have pod using this interface and then delete the pod, the secondary interface remains unusable until node reboots. right?
> > 
> > 
> > no when the interface is back on the host network namespace networkManager for example sees the device and run the configuration again on the device it can be static,dhcp or what every is convigured in the device
> 
> you assume they are all using NM (and auto connect feature) which probably is not true. I would have thought there is a cloudinit script that does once on boot the network config but I have not worked with clouds for many years, I do not have strong opinion here.

I tested this on two cluster provides for k8s deployments (GCP,IBM cloud) and they both has the same beavior.
also as commented if we see a user request in the future where this solution doesn't work we can iterate again and implement the full cache file for the CNI_DEL to restore the configuration.

plugins/main/host-device/host-network-state.go
SchSeba: done

plugins/main/host-device/host-network-state_test.go
SchSeba: done

host-device
squeed, line 1: nit: looks like you accidentally added a binary file.

SchSeba: shame on me :P 
done

SchSeba:
Hi @s1061123 @squeed let me know if we can merge this PR thanks! :)

LionelJouin (approved):
I talked directly with @SchSeba , this looks good to me.

SchSeba:
Hi @squeed @mlguerrero12 @s1061123 can we merge this one?

plugins/main/host-device/host-network-state.go
```diff
@@ -0,0 +1,198 @@
...
+		return fmt.Errorf("failed to list addresses for host device %s: %w", hostDev.Attrs().Name, err)
+	}
+	for _, addr := range addrs {
+		if addr.Scope != int(netlink.SCOPE_UNIVERSE) || addr.IPNet == nil {
+			continue
+		}
+		// Skip dynamic (SLAAC / DHCPv6) addresses: they are managed by the
+		// kernel with a finite lifetime and there is no renewal daemon in
+		// the container, so they would become stale permanent entries.
+		if addr.Flags&ifaFlagPermanent == 0 {
```
mlguerrero12: can't we not use syscall.IFA_F_PERMANENT?

plugins/main/host-device/host-network-state.go
```diff
@@ -0,0 +1,198 @@
...
+	"github.com/vishvananda/netlink"
+
+	"github.com/containernetworking/cni/pkg/types"
+	current "github.com/containernetworking/cni/pkg/types/100"
+	"github.com/containernetworking/plugins/pkg/netlinksafe"
+	"github.com/containernetworking/plugins/pkg/ns"
+)
+
+const (
+	localRouteTable  = 255
```
mlguerrero12: unix.RT_TABLE_LOCAL
unix.IFA_F_PERMANENT

plugins/main/host-device/host-network-state.go
```diff
@@ -0,0 +1,198 @@
...
+	ifaFlagPermanent = 0x80 // IFA_F_PERMANENT from linux/if_addr.h
+)
+
+// HostNetworkState holds the captured host-side L3 configuration
+// (addresses, routes, and rules) that should be applied to the container interface.
+// This struct is never serialized; it lives only for the duration of a single cmdAdd call.
+type HostNetworkState struct {
+	HostIfName    string
+	HostLinkWasUp bool
```
mlguerrero12: Doesn't seem to have a purpose. Remove?

plugins/main/host-device/host-network-state.go
```diff
@@ -0,0 +1,198 @@
...
+func captureHostNetworkState(state *HostNetworkState, hostDev netlink.Link) error {
+	addrs, err := netlinksafe.AddrList(hostDev, netlink.FAMILY_ALL)
+	if err != nil {
+		return fmt.Errorf("failed to list addresses for host device %s: %w", hostDev.Attrs().Name, err)
+	}
+	for _, addr := range addrs {
+		if addr.Scope != int(netlink.SCOPE_UNIVERSE) || addr.IPNet == nil {
+			continue
+		}
+		// Skip dynamic (SLAAC / DHCPv6) addresses: they are managed by the
```
mlguerrero12: This also applies to DHCPv4. Isn't DHCP via NM/networkd the most common cloud case?

So I wonder if it is not better to copy these addresses (excluding SLAAC) as permanent, since nothing renews them inside the container. Or are we ok documenting that this won't work with DHCP?

plugins/main/host-device/host-network-state.go
```diff
@@ -0,0 +1,198 @@
...
+	}
+
+	rules, err := netlinksafe.RuleList(netlink.FAMILY_ALL)
+	if err != nil {
+		return fmt.Errorf("failed to list rules: %w", err)
+	}
+	for _, rule := range rules {
+		if rule.Src == nil {
+			continue
+		}
```
mlguerrero12, line 115: I believe we should filter by the IPs or tables obtained previously. It could be possible to add rules from other devices or ones that refer to empty tables.

plugins/main/host-device/host-network-state.go
```diff
@@ -0,0 +1,198 @@
...
+	}
+
+	orderedRoutes := make([]netlink.Route, len(state.Routes))
+	copy(orderedRoutes, state.Routes)
+	sort.SliceStable(orderedRoutes, func(i, j int) bool {
+		return orderedRoutes[i].Scope > orderedRoutes[j].Scope
+	})
+	for _, route := range orderedRoutes {
+		route.LinkIndex = linkIndex
+		if err := ignoreExists(netlink.RouteAdd(&route)); err != nil {
```
mlguerrero12, line 183: Say a dynamic IPv4 was skipped, then adding the routes will fail with ENETUNREACH for any route with a gateway (inclding the default one), and EINVAL for rotes carrying src.

I don't think skipping those routes would be safe. Letting the pod fail seems better than having it run without connectivity. Is this behavior intentional?

plugins/main/host-device/host-device.go
```diff
@@ -670,9 +730,9 @@ func validateCniContainerInterface(intf current.Interface) error {
 }
 
 func cmdStatus(args *skel.CmdArgs) error {
-	conf := NetConf{}
```
mlguerrero12, line 673: I think this should be in another commit (or even in another PR I would say). Please add some tests.

`loadConf` will fails for a config that only uses the `deviceId` capability. Runtime config is not injected in the Status cmd. I think, when this capability is true, we shouldn't fail in `loadConf` and simply return nil after IPAM status. Note that any other way to identify the device takes precedence over this capability (switch in `getLink`).

plugins/main/host-device/host-device.go
```diff
@@ -681,7 +741,11 @@ func cmdStatus(args *skel.CmdArgs) error {
 		}
 	}
 
-	// TODO: Check if host device exists.
+	if !conf.DPDKMode {
+		if _, err := getLink(conf.Device, conf.HWAddr, conf.KernelPath, conf.PCIAddr, conf.auxDevice); err != nil {
+			return fmt.Errorf("failed to find host device: %v", err)
```
mlguerrero12: Let's return the new ErrPluginNotAvailable. I think you need to bump the cni version

mlguerrero12 (changes requested):
Thanks @SchSeba. Left some comments.

plugins/main/host-device/host-device.go
SchSeba: I removed this and return to the original TODO for now as  this PR really doesn't need to implement that function

plugins/main/host-device/host-network-state.go
SchSeba: agree I make the DHCPv4 to be permanent and do not skip it

plugins/main/host-device/host-network-state.go
SchSeba: nice catch thanks!

plugins/main/host-device/host-network-state.go
SchSeba: done

mlguerrero12 (approved):
LGTM. Thanks @SchSeba! Left a comment in the docs PR

mlguerrero12:
@s1061123's comments have been addressed

PR #1259, opened 2026-05-18
Move a comment to the right place

danwinship:
noticed by coderabbit while rebasing openshift/containernetworking-plugins

(well, sort of: it noticed that the comment was incorrect where it was, but it didn't notice that it apparently belonged in the following test)


danwinship:
irrelevant now that the test has been reverted

PR #1261, opened 2026-05-24
 ipam: fix inconsistent behavior for same-prefix routes between IPv4 and IPv6

yushoyamaguchi:
# Overview
When multiple routes have the same prefix, IPv4 inserted them as separate entries while IPv6 kernel behavior implicitly merged them into ECMP.
This caused the same config to behave differently depending on the address family.
This patch absorbs the kernel behavior difference on the `plugins` side to ensure consistent ECMP behavior regardless of address family.

## Related issue/PR
https://github.com/containernetworking/plugins/pull/615
https://github.com/containernetworking/plugins/issues/1249

# Current behavior
## IPv4
When we write as below in cni's json,
```
      "routes": [
        { "dst": "0.0.0.0/0", "gw": "10.0.0.1" },
        { "dst": "0.0.0.0/0", "gw": "10.0.0.2" }
      ]
```
Pod's FIB is below
```
default via 10.0.0.1 dev ens3 
default via 10.0.0.2 dev ens3 
```
## IPv6
When we write as below in cni's json,
```
  "routes": [
    { "dst": "::/0", "gw": "2001:db8::1" },
    { "dst": "::/0", "gw": "2001:db8::2" }
  ]
```
Pod's FIB is below
```
default metric 1024 pref medium
        nexthop via 2001:db8::1 dev ens3 weight 1 
        nexthop via 2001:db8::2 dev ens3 weight 1 
```

# Implementation
Group routes with the same destination prefix, table, scope, and priority, then apply them as a single ECMP route via `RouteAddEcmp` with MultiPath.

### DIFF
--- a/pkg/ipam/ipam_linux.go
+++ b/pkg/ipam/ipam_linux.go
@@ -22,6 +22,7 @@ import (
 
 	"github.com/vishvananda/netlink"
 
+	"github.com/containernetworking/cni/pkg/types"
 	current "github.com/containernetworking/cni/pkg/types/100"
 	"github.com/containernetworking/plugins/pkg/ip"
 	"github.com/containernetworking/plugins/pkg/netlinksafe"
@@ -35,6 +36,19 @@ const (
 
 	dadSettleTimeout = 5 * time.Second
 )
+
+type routeKey struct {
+	dst      string
+	table    int
+	scope    int
+	priority int
+}
+
+type routeGroup struct {
+	route netlink.Route
+	dst   net.IPNet
+	gws   []net.IP
+}
 
 // ConfigureIface takes the result of IPAM plugin and
 // applies to the ifName interface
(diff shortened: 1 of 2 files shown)


yushoyamaguchi:
cc @s1061123

yushoyamaguchi:
This inconsistency has been causing confusion in our operation of production environment, so I am submitting this as a bug fix rather than a feature change.

s1061123 (changes requested):
/hold

As mentioned in https://github.com/containernetworking/plugins/issues/1249, this code changes IPAM behavior completely and it causes its regression (i.e. this bugfix causes bigger bugs) and current valid configuration is deprecated. In addition, CNI spec does not specify how CNI result object shows ECMP path as well.

It should be held until we conclude its design discussion, in https://github.com/containernetworking/plugins/issues/1249.

yushoyamaguchi:
ECMP feature will be introduced with new item `gws ` in CNI Spec.
https://hackmd.io/jU7dQ49dQ86ugrXBx1De9w#2026-06-07

PR #1265, opened 2026-06-17
bridge: reliably pin the MAC on a bridge that lost its ports

twz123:
The `ensureAddr` function pins the bridge's MAC address so it doesn't change as ports come and go. The pinning happens after the gateway address is set, so if a run set the address but then failed before pinning, subsequent runs never reattempted to pin the MAC, leaving behind a bridge with a floating MAC. The kernel zeroes a bridge's MAC address when its last port is removed, and `ensureAddr` pins the MAC using the address read when the bridge was looked up, which is before the veth is attached. So for an existing bridge that had previously lost all its ports, the address it tried to pin is all zeros, and pinning an all-zero address fails with `EINVAL`.

In fact, this can happen in practice: an ADD that fails has its veth cleaned up, which, if it was the first ADD, leaves the bridge portless and hence its MAC zeroed, so the next ADD fails to pin it.

Always pin the MAC, even if the gateway address is already present, regardless of the outcomes of previous attempts. Detect zero MACs and regenerate a random MAC address the same way the kernel does. This is fine as it only happens if the bridge has no ports, so no peer can hold stale state about the old address.

### DIFF
--- a/plugins/main/bridge/bridge.go
+++ b/plugins/main/bridge/bridge.go
@@ -271,11 +271,13 @@ func ensureAddr(br netlink.Link, family int, ipn *net.IPNet, forceAddress bool) 
 	}
 
 	ipnStr := ipn.String()
+	addrFound := false
 	for _, a := range addrs {
 
 		// string comp is actually easiest for doing IPNet comps
 		if a.IPNet.String() == ipnStr {
-			return nil
+			addrFound = true
+			break
 		}
 
 		// Multiple IPv6 addresses are allowed on the bridge if the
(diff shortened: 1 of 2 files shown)


squeed:
Thanks for the PR. Yet more fun with the Linux bridge.

plugins/main/bridge/bridge.go
```diff
@@ -293,14 +296,35 @@ func ensureAddr(br netlink.Link, family int, ipn *net.IPNet, forceAddress bool)
...
+	// Check whether the kernel has zeroed the bridge's MAC.
+	// This only happens when the bridge's last port was detached.
+	if isZeroMAC(hwAddr) {
+		// There are no neighbors that could hold stale state about the previous
+		// MAC, so it's fine to generate a suitable stand-in the same way the
+		// kernel does.
+		// https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/tree/include/linux/etherdevice.h?h=v7.1#n237
+		hwAddr = make(net.HardwareAddr, 6)
+		if _, err := rand.Read(hwAddr); err != nil {
+			panic(fmt.Sprintf("failed to generate random MAC address: %v", err))
```
mlguerrero12: let's not use panic

mlguerrero12 (commented):
LGTM. Just a minor comment

twz123:
@mlguerrero12 I've addressed the comment. Mind having another look?

PR #1271, opened 2026-07-08
dhcp plugin: send dhcp release from leased address

Pardhakeswar:
DHCPRELEASE should be sent from the leased client address so the IP source matches ciaddr. Use a UDP socket bound to the lease address for release, with the existing raw socket path kept as a fallback.


Pardhakeswar:
@squeed can you PTAL when you get a chance?

trozet:
@LionelJouin do you mind helping us out here please?

Issue #957, opened 2023-10-16
host-device cannot use linux wireless devices


rfenlon:
When configured to use a wireless device (e.g. wlo1) host-device fails to change the network namespace with an `invalid argument ` response (see netlink/link_linux.go:601). Consider conditionally doing the equivalent of `iw phy <my-phy> set netns name <my-namespace>` when it's a wireless device.

rfenlon:
I dug into this a little bit and found that the linux vendor dependency github.com/vishvananda/netlink does not support sending nl80211 messages to netlink. Here is an example of a working netlink request message to change the namespace of a phy using file descriptor for the new network namespace :

--------------------------   BEGIN NETLINK MESSAGE --------------------------- 
  [NETLINK HEADER] 16 octets 
    .nlmsg_len = 36 
    .type = 32 <0x20> 
    .flags = 5 <REQUEST,ACK> 
    .seq = 1697642139 
    .port = -2016371064 
  [GENERIC NETLINK HEADER] 4 octets 
    .cmd = 49 <NL80211_CMD_SET_WIPHY_NETNS> 
    .version = 0 
    .unused = 0 
  [PAYLOAD] 16 octets 
    08 00 01 00 01 00 00 00 08 00 db 00 04 00 00 00 ................ <NL80211_ATTR_WIPHY 0x01> and <NL80211_ATTR_NETNS_FD 0xdb> 
---------------------------  END NETLINK MESSAGE   --------------------------- 

I'm not a kernel developer so I'm having trouble finding the following:
* netlink header type = 32 // unknown what Generic Netlink Family ID is tied to 32
* payload //  I know 2 octets = NL80211_ATTR_WIPHY, 4 octets = device index, 2 octets = NL80211_ATTR_NETNS_FD, 4 octets = file descriptor. I don't know what the remaining 4 octets are doing.

PR #1274, opened 2026-07-20
host-device: support wireless PHY namespace moves

mbaragiola:
## Summary

Add wireless PHY support to the `host-device` CNI plugin.

Linux wireless devices cannot be moved with the ordinary `RTM_SETLINK`/`LinkSetNsFd` path. This change detects wireless interfaces through sysfs and uses the `nl80211` wiphy move operation instead.

Safety behavior:

- Wireless PHYs with sibling interfaces are rejected because moving a wiphy moves every associated interface.
- Ordinary links retain the existing movement path.
- ADD, DEL, and rollback paths use the matching operation.
- The unavoidable race between sysfs sibling enumeration and the subsequent wiphy move is documented; there is no atomic userspace operation covering both steps.

## Dependency

This PR depends on the Linux-only API proposed in:

- https://github.com/vishvananda/netlink/pull/1203

The current declared `vishvananda/netlink v1.3.1` does not contain `WiphySetNsFd`, so this PR is intentionally opened as a draft until that API is accepted and released (or the maintainers advise on dependency sequencing).

Related issue: https://github.com/containernetworking/plugins/issues/957

## Testing

- Focused wireless tests with a temporary local netlink replacement: passed.
- `go vet` for `plugins/main/host-device` with the temporary local replacement: passed.
- `gofmt` and `git diff --check`: passed.
- `go.mod` and `go.sum` remain unchanged.
- Privileged namespace integration tests cannot run in the current sandbox because namespace setup fails with `operation not permitted`.

### DIFF
--- a/plugins/main/host-device/host-device.go
+++ b/plugins/main/host-device/host-device.go
@@ -23,6 +23,7 @@ import (
 	"os"
 	"path/filepath"
 	"runtime"
+	"strconv"
 	"strings"
 
 	"github.com/vishvananda/netlink"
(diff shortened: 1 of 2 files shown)


mbaragiola:
CI is currently blocked by dependency sequencing, not by the GitHub runner and not by a separate defect found in this CNI diff.

Exact evidence:

- The latest CNI workflow run is `29760443748`, job `88413435298` (`Lint`). golangci-lint reaches Go type checking and fails at `plugins/main/host-device/host-device.go:49:25` with `undefined: netlink.WiphySetNsFd`.
- This branch's `go.mod` and vendor metadata still select `github.com/vishvananda/netlink v1.3.1`. That release does not contain `WiphySetNsFd`; the API is introduced by vishvananda/netlink#1203, whose current head is `8079d6be48527700e1628d3ff69563c94656eb61`.
- I reproduced the identical type-check error locally with the declared dependency.
- In an isolated temporary tree, I supplied the exact API implementation from netlink PR #1203 without changing this branch's `go.mod` or adding a replace directive. The 9 new `wireless device moves` specs passed, and the host-device test package compiled for every architecture used by CNI CI: `amd64 386 arm arm64 s390x mips64le ppc64le riscv64`.
- The other 44 host-device specs could not execute in the local sandbox because test namespace setup failed with `mount --make-rshared /run/user/1000/netns: operation not permitted`; this occurred after successful compilation and is local privilege/environment evidence only.
- I reviewed the full 206-line CNI diff and did not find an independent actionable defect that would explain the CI result.

Note that netlink PR #1203 itself currently has an unrelated Linux check failure: run `29760919064`, job `88415026849`, fails while compiling pre-existing `handle_retry_linux_test.go:60` because `(*Handle).RetryInterrupted` is undefined. Its macOS check passes. That needs to be resolved in the dependency PR, then the netlink change must be merged and released (or otherwise become a publishable module version), after which this CNI PR can update `go.mod` and vendor normally.

No CNI code or module files were changed.

PR #1279, opened 2026-07-21
dhcp: send DHCPRELEASE with the client IP addr as ciaddr

### LINKED ISSUE #1278: dhcp IPAM plugin: `DHCPRELEASE` sent with 0.0.0.0 source IP, ignored by servers
## Summary

The dhcp IPAM plugin sends `DHCPRELEASE` packets with 0.0.0.0 as the IP source address instead of the client's assigned IP.
This violates [RFC 2131 section 4.4.4](https://datatracker.ietf.org/doc/html/rfc2131#section-4.4.4) and causes DHCP servers (e.g. dnsmasq) to silently ignore the release.
Leases are never freed - they only expire via timeout.

### RFC 2131, Section 4.4.4

▎ The client unicasts DHCPRELEASE messages to the server. [...] The client MUST fill in its own IP address in the 'ciaddr' field.

While ciaddr is set correctly in the DHCP payload, the IP header source is 0.0.0.0 because the plugin uses a raw packet socket (the default nclient4.New() behavior) rather than a regular UDP socket bound to the assigned address.

## Root Cause

`release()` in plugins/ipam/dhcp/lease.go calls `newDHCPClient(l.link, l.timeout)` without any client options.
This creates a raw packet socket that constructs IP headers with 0.0.0.0 as the source - appropriate for `DHCPDISCOVER` (no address assigned yet) but wrong for `DHCPRELEASE`.

The underlying library (github.com/insomniacslk/dhcp/dhcpv4/nclient4) already provides `nclient4.WithUnicast()` for exactly this case. Its own documentation on the Release method notes:

▎ Note: some DHCP server requires using assigned IP address as source IP, use nclient4.WithUnicast to create client for such case.

## Packet Capture Evidence

`DHCPRELEASE` as captured on the server interface:

```
12:13:10.523945 IP 0.0.0.0.68 > 192.168.11.1.67: BOOTP/DHCP, Request from aa:bb:cc:00:00:01, length 314
      Client-IP 192.168.11.119
      DHCP-Message (53), length 1: Release
      Server-ID (54), length 4: 192.168.11.1
```

Source IP is 0.0.0.0. dnsmasq receives the packet but does not process it - the lease remains active until it expires.

## Environment

- CNI plugins version: v1.6.2 (also reproducible at HEAD)
- DHCP server: dnsmasq 2.90
- Client library: github.com/insomniacslk/dhcp v0.0.0-20240829085014-a3a4c1f04475

maiqueb:
The dhcp IPAM plugin sends DHCPRELEASE packets with 0.0.0.0 as the IP source address instead of the client's assigned IP.

This violates RFC 2131 section [4.4.4](https://datatracker.ietf.org/doc/html/rfc2131#section-4.4.4) (quote below) and causes DHCP servers (e.g. dnsmasq) to silently ignore the release. Leases are never freed - they only expire via timeout.

> Use of broadcast and unicast
> The DHCP client broadcasts DHCPDISCOVER, DHCPREQUEST and DHCPINFORM
> messages, unless the client knows the address of a DHCP server.  The
> client unicasts DHCPRELEASE messages to the server.

By using the `nclient4.WithUnicast()` option to create the `DHCPClient`, the client will send unicast msgs.

Fixes: #1278

### DIFF
--- a/plugins/ipam/dhcp/lease.go
+++ b/plugins/ipam/dhcp/lease.go
@@ -428,7 +428,14 @@ func (l *DHCPLease) renew() error {
 func (l *DHCPLease) release() error {
 	log.Printf("%v: releasing lease", l.clientID)
 
-	c, err := newDHCPClient(l.link, l.timeout)
+	c, err := newDHCPClient(
+		l.link,
+		l.timeout,
+		nclient4.WithUnicast(&net.UDPAddr{
+			IP:   l.latestLease.ACK.YourIPAddr,
+			Port: nclient4.ClientPort,
+		}),
+	)
 	if err != nil {
 		return err
 	}


maiqueb:
Tested this fix manually and it is working as expected - CNI DEL with DHCP IPAM sends the DHCPRELEASE msg w/ the proper source IP address, and in turn, dnsmasq actually processes the packet and releases the DHCP lease.

zeeke (approved):
+1, he fix aligns with the RFC

SchSeba (approved):
LGTM

MikeZappa87:
LGTM

MikeZappa87 (approved):

MikeZappa87:
@squeed

